PR #79, opened 2016-02-17
Allow overriding settings from Chromium's common.gypi

nevil:
Build settings are coming from the Chromium build settings in packager/build/common.gypi.
We need to override those settings and set CLANG_CXX_LIBRARY = libc++ for MacOSX builds.
packager/common.gypi does some override but only for packages that explicitly include it.
We want a global override.
Introduce packager/common_overrides.gypi which is injected as an argument to gyp in gyp_packager.py

### DIFF
--- a/gyp_packager.py
+++ b/gyp_packager.py
@@ -60,6 +60,9 @@ if __name__ == '__main__':
   # Always include common.gypi.
   args.extend(['-I' + os.path.join(src_dir, 'build', 'common.gypi')])
 
+  # Allow overriding settings from Chromium's common.gypi
+  args.extend(['-I' + os.path.join(src_dir, 'common_overrides.gypi')])
+
   # Set these default GYP_DEFINES if user does not set the value explicitly.
   _DEFAULT_DEFINES = {'test_isolation_mode': 'noop',
                       'use_glib': 0,
--- a/packager/common_overrides.gypi
+++ b/packager/common_overrides.gypi
@@ -1,0 +1,19 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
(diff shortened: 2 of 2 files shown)


kqyang:
@nevil Thanks for working on this problem!

As mentioned in #80, we will fix the compilation issue in the unittest. 

It looks like Chromium already switches to libc++, we will update the DEPS later to use newer version of common.gypi instead :)

nevil:
@kqyang 
I noticed that Chromium already switched, but updating the "build" module introduces some other warnings and errors so I opted for this approach :-)

Just a heads up.
After updating build to latest version you will need to override flags to avoid warnings and errors like:

```
[10/1269] CXX obj/version/version.version.o
Unknown clang plugin argument: follow-macro-expansion
...

../../packager/third_party/protobuf/src/google/protobuf/stubs/stringprintf.cc:164:59: error: unused typedef 'arg_count_mismatch' [-Werror,-Wunused-local-typedef]
  GOOGLE_COMPILE_ASSERT(kStringPrintfVectorMaxArgs == 32, arg_count_mismatch);
                                                          ^
../../packager/third_party/protobuf/src/google/protobuf/stubs/common.h:397:11: note: expanded from macro 'GOOGLE_COMPILE_ASSERT'
          msg[bool(expr) ? 1 : -1]
          ^
...
../../packager/third_party/protobuf/src/google/protobuf/stubs/strutil.h:319:25: error: unused typedef 'sizeof_int64_is_not_sizeof_long_long' [-Werror,-Wunused-local-typedef]
                        sizeof_int64_is_not_sizeof_long_long);
```
And so on.

Thanks

kqyang:
I see. Thanks for the heads up :). We should update protobuf deps at the same time.

PR #106, opened 2016-06-01
Close files in Finalize for WebM segmenters.

balinisteanu:
Now Finalize() will close all the files used by the segmenters.  This
is important to allow WebMMuxer::FireOnMediaEndEvent to get the
correct size of the media file.

This also changes the WebM muxer to use 64-bit numbers for the init
and index ranges to correctly support >4GB files.

b/28194272

Change-Id: Ia84e4a4b0756f89644efea99a1a51968b22a1338


kqyang:
@balinisteanu This change is already in master. I'll close it. Feel free to re-open the bug if I am missing anything here.

PR #115, opened 2016-07-07
Specify Ubuntu 14.04 in Dockerfile to ensure successful build

gkop:
With more recent Ubuntu/Debian, I get this build error when running `ninja -C out/Debug`:

```
../../packager/media/formats/webm/webm_cluster_parser_unittest.cc:603:21: error: default initialization of an object of const type 'const BufferQueue' (aka 'const deque<scoped_refptr<shaka::media::MediaSample> >') without a user-provided default constructor
  const BufferQueue no_text_buffers;
                    ^
                                   {}
1 error generated.
[1238/1302] CXX obj/hls/base/hls_unittest.simple_hls_notifier_unittest.o
ninja: build stopped: subcommand failed.
```

Since CI (Travis) is running Ubuntu 12.04, it seems like a safe bet for the development/test Dockerfile.

I'm working on the corporate CLA now...

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,4 +16,5 @@
 Anders Hasselqvist <anders.hasselqvist@gmail.com>
 Google Inc. <*@google.com>
 Leandro Moreira <leandro.ribeiro.moreira@gmail.com>
+Philo Inc. <*@philo.com>
 The Chromium Authors <*@chromium.org>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -24,6 +24,7 @@
 
 Anders Hasselqvist <anders.hasselqvist@gmail.com>
 Bei Li <beil@google.com>
+Gabe Kopley <gabe@philo.com>
 Jacob Trimble <modmaker@google.com>
 Joey Parrish <joeyparrish@google.com>
 Kongqun Yang <kqyang@google.com>
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:14.04
 
 ENV DEBIAN_FRONTEND noninteractive
 


gkop:
My employer Philo has a corporate CLA signed. We've added my email gabe@philo.com to our CLA google group at https://groups.google.com/forum/#!forum/philo-shaka . gabe@philo.com is also the email address on the commit.

kqyang:
@gkop Thanks for working on improving shaka-packager! May I know what version of Ubuntu are you using?  I tried version 12.04, 14.04 and 15.04 but I am not seeing problems.

gkop:
Whoops, I should have specified - I _only_ get the error with `BUILD_TYPE=Debug`. With `BUILD_TYPE=Release` I get a successful build. I figured for test/development I ought to have `BUILD_TYPE=Debug` working.

I found this first when getting the latest Ubuntu in Docker, which is 16.04 today. I was also able to replicate on my host machine which is running Debian Stretch.

Btw I think it's also a simple good practice to specify a tag for the FROM image in a Dockerfile, especially as we're not testing the Dockerfile in CI. cf. https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375 and http://blog.benhall.me.uk/2015/01/dockerfile-latest-tag-anti-pattern/ .

kqyang:
I see. 

The docker file is actually tested in docker build: https://hub.docker.com/r/google/shaka-packager/builds/. It is only verifying Release build though.

I am fine with using a tagged version instead of "latest". I am a bit uncomfortable to use version 12.04, which is a little bit old. Can you change to version 14.04 or 15.04?

And thanks for reporting and fixing the issue! I have filed a separate bug #118 to track the build break.

gkop:
That's great that Docker Hub builds the image! I suppose the reason this wasn't caught by the Docker Hub build is just that the Dockerfile runs Release rather than Debug.

I changed to 14.04; it works well and I think that because it is an LTS release it is a more helpful choice than 15.04 (even though 15.04 works also).

kqyang:
The change has been merged. Thanks for improving shaka-packager project!

Issue #113, opened 2016-06-28
"Off-by-one frame" bug when packaging from mp2 to mp4?


gkop:
Thanks for your work helping us package our DASH video! :)

Here's my test case:

```bash
#!/usr/bin/env bash

packager | grep "edash-packager version" # version v1.4.1-9f2da1e-release

wget http://static.coshx.com/example.ts

packager input=example.ts,stream_selector=0,output=video_init.mp4,segment_template=\$Time\$.mp4v,output_format=MP4,bandwidth=200000 -single_segment=false

ls -l *.mp4v
mp4dump 356038323.mp4v
```

The resulting media fails to play in the demo Shaka Player app ([using this manifest](https://gist.github.com/gkop/1bed9f8d888f2e4ceb966cf75a80b9d1)) due to an EME error. When I edit the TFDT box on 356038323.mp4v and change the base media decode time from 356035320 to 356038323, then the output _does_ play successfully in the demo Shaka Player app. I'm somewhat new to video, but it makes sense to me that the base media decode time should match the manifest timestamp and the filename timestamp. Instead it appears to be off by one frame. Is this a bug or am I misunderstanding how to use this project?

kqyang:
It is actually a bug in Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=398130#c13, which unfortunately is not fixed yet.

Usually for the first frame, the presentation timestamp and decoding timestamp are the same, but it is not the case for your test content. There is a difference of 3003 ticks. Chrome uses decoding timestamp for the seeking right now, while it should use presentation timestamp instead.

A typical workaround to the problem is to adjust the manifest to use the decoding time instead, i.e.
replacing **<S t="356038323"** with **<S t="356035320"** in the manifest.

Let us know if you have more questions.

gkop:
Hmm, still struggling here - having trouble corroborating with [a recent ffprobe](https://gist.github.com/gkop/175f118a49a9d3d18afcae912c35b186) that the pts and dts are in fact different on the frames of this example fragment:

```
ffprobe -show_frames -show_entries frame=pkt_pts,pkt_dts -of json example.ts
```

The first 229 frames have matching dts and pts, the 230th frame is missing dts. Could that be what is causing the problem here, the missing dts on the 230th frame?

[...]

PR #116, opened 2016-07-07
Remove dead code from box definitions unit test

gkop:
I came across this dead code while poking around to make a patch to use decode time in filename templates (cf. #113 ).

### DIFF
--- a/packager/media/formats/mp4/box_definitions_unittest.cc
+++ b/packager/media/formats/mp4/box_definitions_unittest.cc
@@ -761,16 +761,6 @@ class BoxDefinitionsTestGeneral : public testing::Test {
 
   void Modify(Movie* moov) { moov->tracks.resize(1); }
 
-  void Fill(TrackFragmentDecodeTime* tfdt) {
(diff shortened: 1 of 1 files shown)


packager/media/formats/mp4/box_definitions_unittest.cc
```diff
@@ -761,16 +761,6 @@ class BoxDefinitionsTestGeneral : public testing::Test {
...
-  void Fill(TrackFragmentDecodeTime* tfdt) {
-    tfdt->decode_time = 234029673820ULL;
-    tfdt->version = 1;
-  }
-
-  void Modify(TrackFragmentDecodeTime* tfdt) {
-    tfdt->decode_time = 4567;
-    tfdt->version = 0;
-  }
-
```
kqyang: This is not a dead code. It is used by the TYPED_TEST_P below to make sure tfdt can be written and read back correctly.

gkop: Ah I see, makes complete sense. Thanks!

PR #120, opened 2016-07-12
Support $DecodeTime$ as filename template variable, work-in-progress

gkop:
This not yet merge-able, but is a minimum hack to support my use case of packaging DASH for Chrome when the fragment start PTS and start DTS do not match ( #113 ). I'm working on improving the correctness and design of this patch, but figured I'd share my direction in case you all have early feedback. Thanks!

### DIFF
--- a/packager/media/base/muxer_util.cc
+++ b/packager/media/base/muxer_util.cc
@@ -46,6 +46,10 @@ bool ValidateSegmentTemplate(const std::string& segment_template) {
   // ISO/IEC 23009-1:2012 5.3.9.4.4 Template-based Segment URL construction.
   // Allowed identifiers: $$, $RepresentationID$, $Number$, $Bandwidth$, $Time$.
   // "$" always appears in pairs, so there should be odd number of splits.
+  // We also allow $DecodeTime$ in order to help work around this longstanding
+  // bug in Chrome:
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=398130#c13
+  // cf. https://github.com/google/shaka-packager/issues/113
   if (splits.size() % 2 == 0) {
     LOG(ERROR) << "SegmentTemplate: '$' should appear in pairs.";
     return false;
@@ -72,7 +76,7 @@ bool ValidateSegmentTemplate(const std::string& segment_template) {
       return false;
     } else if (identifier == "Number") {
       has_number = true;
-    } else if (identifier == "Time") {
+    } else if (identifier == "Time" || identifier == "DecodeTime") {
       has_time = true;
     } else if (identifier == "") {
       if (format_pos != std::string::npos) {
@@ -86,11 +90,11 @@ bool ValidateSegmentTemplate(const std::string& segment_template) {
     }
   }
   if (has_number && has_time) {
-    LOG(ERROR) << "SegmentTemplate: $Number$ and $Time$ should not co-exist.";
+    LOG(ERROR) << "SegmentTemplate: $Number$ and $Time$/$DecodeTime$ should not co-exist.";
     return false;
   }
   if (!has_number && !has_time) {
-    LOG(ERROR) << "SegmentTemplate: $Number$ or $Time$ should exist.";
+    LOG(ERROR) << "SegmentTemplate: $Number$ or $Time$/$DecodeTime$ should exist.";
     return false;
   }
   // Note: The below check is skipped.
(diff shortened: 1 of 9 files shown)


packager/media/base/muxer_util.cc
```diff
@@ -86,11 +90,11 @@ bool ValidateSegmentTemplate(const std::string& segment_template) {
     }
   }
   if (has_number && has_time) {
-    LOG(ERROR) << "SegmentTemplate: $Number$ and $Time$ should not co-exist.";
+    LOG(ERROR) << "SegmentTemplate: $Number$ and $Time$/$DecodeTime$ should not co-exist.";
     return false;
   }
   if (!has_number && !has_time) {
-    LOG(ERROR) << "SegmentTemplate: $Number$ or $Time$ should exist.";
+    LOG(ERROR) << "SegmentTemplate: $Number$ or $Time$/$DecodeTime$ should exist.";
```
gkop: We need to figure out exactly how we want to validate these, because we're deviating from ISO/IEC 23009-1.

packager/media/base/muxer_util.h
```diff
@@ -36,6 +36,7 @@ bool ValidateSegmentTemplate(const std::string& segment_template);
 /// @return The segment name with identifier substituted.
 std::string GetSegmentName(const std::string& segment_template,
                            uint64_t segment_start_time,
+                           uint64_t segment_start_decode_time,
                            uint32_t segment_index,
                            uint32_t bandwidth);
```
gkop: This signature is getting a bit unwieldy :/

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -116,7 +116,7 @@ Status TsSegmenter::OpenNewSegmentIfClosed(uint32_t next_pts) {
   if (ts_writer_file_opened_)
     return Status::OK;
   const std::string segment_name =
-      GetSegmentName(muxer_options_.segment_template, next_pts,
+      GetSegmentName(muxer_options_.segment_template, next_pts, next_pts,
```
gkop: TODO ^^

packager/media/formats/mp4/multi_segment_segmenter.cc
```diff
@@ -150,6 +150,7 @@ Status MultiSegmentSegmenter::WriteSegment() {
   } else {
     file_name = GetSegmentName(options().segment_template,
                                sidx()->earliest_presentation_time,
+                               moof()->tracks[0].decode_time.decode_time,
```
gkop: This is the crux of it. I'm pretty sure hardcoding the `[0]` index is a logical bug, but am still grokking the codebase. Also I want to see if I can avoid needing a `moof()` method at all.

packager/media/formats/mp4/segmenter.h
```diff
@@ -111,6 +111,7 @@ class Segmenter {
   Movie* moov() { return moov_.get(); }
   BufferWriter* fragment_buffer() { return fragment_buffer_.get(); }
   SegmentIndex* sidx() { return sidx_.get(); }
+  MovieFragment* moof() { return moof_.get() ; } // NOTE delete me!
```
gkop: Not intending to merge this, need to find a more cohesive solution.

packager/media/formats/webm/multi_segment_segmenter.cc
```diff
@@ -77,8 +77,8 @@ Status MultiSegmentSegmenter::NewSegment(uint64_t start_timescale) {
 
   // Create a new file for the new segment.
   std::string segment_name =
-      GetSegmentName(options().segment_template, start_timescale, num_segment_,
-                     options().bandwidth);
+      GetSegmentName(options().segment_template, start_timescale,
+                     start_timescale, num_segment_, options().bandwidth);
```
gkop: TODO ^^

kqyang:
@gkop Thanks for working on this issue. I was thinking of visiting it some time after July 4th holiday.

I noticed that in your patchset, you handled the segment name generation for multi-segment case (typically used for live). We'll need to handle single-segment case (typically for on-demand) and the mpd segment timeline generation as well. It looks like that there is a simpler solution for this: (1) we could introduce a runtime flag, --use_decoding_timestamp_in_mp4_timeline; (2) when the flag is set, set earliest_presentation_time_ to dts instead of pts in https://github.com/google/shaka-packager/blob/4f9c8ea7817b397c4f2dd83311f528200650bd3d/packager/media/formats/mp4/fragmenter.cc#L71. 

Here is the patch using the above described approach:

[0001-Add-a-runtime-flag-to-use-dts-in-timeline-for-mp4.txt](https://github.com/google/shaka-packager/files/360259/0001-Add-a-runtime-flag-to-use-dts-in-timeline-for-mp4.txt)

gkop:
Fantastic, thanks!

PR #121, opened 2016-07-12
Add a runtime flag to use dts in timeline for mp4 (by @kqyang )

gkop:
(making this PR because I'm having trouble building locally - curious to see build output from Travis)

This flag is needed to workaround the Chromium bug
https://crbug.com/398130, which uses decoding timestamp
in buffered range.

Change-Id: Ib8f18be7165dd968bdc36c18ce29f694235c0c26

### DIFF
--- a/packager/media/file/file.cc
+++ b/packager/media/file/file.cc
@@ -6,9 +6,9 @@
 
 #include "packager/media/file/file.h"
 
+#include <gflags/gflags.h>
 #include <algorithm>
 
-#include <gflags/gflags.h>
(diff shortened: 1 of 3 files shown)


kqyang:
Did you run gclient runhooks? It is needed since gyp file is updated.

gkop:
What I did was re-config gclient against our fork and synced against my revision. Thanks for the tip on runhooks, that'll save time next time.

It works great. Thanks! <3

kqyang:
I am closing this pull request. We'll push the code with this flag soon.

PR #154, opened 2016-09-20
Do not require exclusive bind of udp sockets

sammirata:
Check return value of setsockopt function call (SO_REUSEADDR)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,5 +17,5 @@ Anders Hasselqvist <anders.hasselqvist@gmail.com>
 Google Inc. <*@google.com>
 Leandro Moreira <leandro.ribeiro.moreira@gmail.com>
 Philo Inc. <*@philo.com>
-The Chromium Authors <*@chromium.org>
 Sergio Ammirata <sergio@ammirata.net>
+The Chromium Authors <*@chromium.org>
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,3 +18,4 @@ Google Inc. <*@google.com>
 Leandro Moreira <leandro.ribeiro.moreira@gmail.com>
 Philo Inc. <*@philo.com>
 The Chromium Authors <*@chromium.org>
+Sergio Ammirata <sergio@ammirata.net>
--- a/packager/media/file/udp_file_posix.cc
+++ b/packager/media/file/udp_file_posix.cc
@@ -187,6 +187,9 @@ bool UdpFile::Open() {
   local_sock_addr.sin_family = AF_INET;
   local_sock_addr.sin_port = htons(dest_port);
   local_sock_addr.sin_addr.s_addr  = htonl(dest_addr);
+  // We do not need to require exclusive bind of udp sockets
+  int optval = 1;
+  setsockopt(new_socket.get(), SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
   if (bind(new_socket.get(),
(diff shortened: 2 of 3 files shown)


sammirata:
I signed it!

packager/media/file/udp_file_posix.cc
```diff
@@ -187,6 +187,9 @@ bool UdpFile::Open() {
   local_sock_addr.sin_family = AF_INET;
   local_sock_addr.sin_port = htons(dest_port);
   local_sock_addr.sin_addr.s_addr  = htonl(dest_addr);
+  // We do not need to require exclusive bind of udp sockets
+  int optval = 1;
```
kqyang: Replace "int" with "const int" to make it clear that it is a const input parameter.

packager/media/file/udp_file_posix.cc
```diff
@@ -187,6 +187,9 @@ bool UdpFile::Open() {
...
+  setsockopt(new_socket.get(), SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
```
kqyang: Please check the return value of setsockopt. Log an error and return false if it fails.

I am not familiar with SO_REUSEADDR, just for my own knowledge, does it matter if we call setsockopt before or after bind?

packager/media/file/udp_file_posix.cc
sammirata: Done

packager/media/file/udp_file_posix.cc
sammirata: Done ... You need to set SO_REUSEADDR before calling bind or it will fail

sammirata:
I sent a new pull request with the additional requested changes

AUTHORS
```diff
@@ -18,3 +18,4 @@ Google Inc. <*@google.com>
 Leandro Moreira <leandro.ribeiro.moreira@gmail.com>
 Philo Inc. <*@philo.com>
 The Chromium Authors <*@chromium.org>
+Sergio Ammirata <sergio@ammirata.net>
```
kqyang: Please order in alphabetical order. Same for contributors.

packager/media/file/udp_file_posix.cc
```diff
@@ -188,8 +188,12 @@ bool UdpFile::Open() {
   local_sock_addr.sin_port = htons(dest_port);
   local_sock_addr.sin_addr.s_addr  = htonl(dest_addr);
   // We do not need to require exclusive bind of udp sockets
-  int optval = 1;
-  setsockopt(new_socket.get(), SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
+  const int optval = 1;
+  if (setsockopt(new_socket.get(), SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval)) < 0)
```
kqyang: We have a 80-characters limit per line; also, "{" should not be in its own line. Can you change it to:

```
  if (setsockopt(new_socket.get(), SOL_SOCKET, SO_REUSEADDR, &optval,
                 sizeof(optval)) < 0) {
    LOG(ERROR)
        << "Could not apply the SO_REUSEADDR property to the UDP socket.";
    return false;
  }
```

AUTHORS
sammirata: Done

packager/media/file/udp_file_posix.cc
sammirata: Done

PR #185, opened 2017-01-04
Add support for udp_file in Windows.

capitalm-chen:
The updated file "udp_file_win.cc" is basically a clone of "udp_file_posix.cc" with a few modifications.

- Socket descriptors use the type "SOCKET".
- Replace POSIX socket functions with their Winsock2 counterparts.
- Initialize structs with zero PODs instead of calling bzero(). 
- Bind local socket to INADDR_ANY in multicast mode, since Windows does not allow binding to multicast addresses.
- Add explicit reinterpret_casts.
- Call WSAStartup() and WSACleanup() at the beginning and the end of packager execution.


packager/app/packager_main.cc
```diff
@@ -415,6 +416,14 @@ Status RunRemuxJobs(const std::vector<std::unique_ptr<RemuxJob>>& remux_jobs) {
 }
 
 bool RunPackager(const StreamDescriptorList& stream_descriptors) {
+#if defined(OS_WIN)
+  WSADATA wsa_data;
+  if (WSAStartup(MAKEWORD(2,2), &wsa_data)) {
+    LOG(ERROR) << "Winsock start up failed.";
+    return false;
+  }
+#endif  // defined(OS_WIN)
```
kqyang: Prefer calling WSAStartup only if it is needed. We can use static initializer in upd_file implementation to do that.

Here is one possible implementation:

```cpp
class LibWinsockInitializer {
 public:
  LibWinsockInitializer() {
    WSADATA wsa_data;
    error_ = WSAStartup(MAKEWORD(2, 2), &wsa_data);
  }

 ~LibWinsockInitializer() {
  if (error_ == 0) WSACleanup();
 }

 int error() const { return error_; }

 private:
  int error_;

  DISALLOW_COPY_AND_REASSIGN(LibWinsockInitializer);
};

bool UdpFile::Open() {
  static LibWinsockInitializer lib_winsock_initializer;
  if (lib_winsock_initializer.error() != 0) {
    LOG(ERROR) << "Winsock start up failed with error " << lib_winsock_initializer.error();
    return false;
  }
  ...
}

```

packager/media/file/udp_file.h
```diff
@@ -14,6 +14,12 @@
 #include "packager/base/compiler_specific.h"
 #include "packager/media/file/file.h"
 
+#if defined(OS_WIN)
+#include <winsock2.h>
+#elif ! defined(SOCKET)
```
kqyang: This condition check should be removed. SOCKET should not have been defined. If it is defined and it is not the same as "int", then we have a serious problem here; it is better to let the compiler capture the problem early in that case.

packager/media/file/udp_file_win.cc
```diff
@@ -6,25 +6,61 @@
 
 #include "packager/media/file/udp_file.h"
 
+#include <windows.h>
+#include <ws2tcpip.h>
+#include <memory>
+#pragma comment (lib, "Ws2_32.lib")
+#pragma comment (lib, "Mswsock.lib")
+#pragma comment (lib, "AdvApi32.lib")
```
capitalm-chen: These explicit linking do not seems to be necessary. I'll try removing them.

packager/media/file/file.gyp
```diff
@@ -29,17 +29,7 @@
...
-        ['OS == "win"', {
-          'sources': [
-            'udp_file_win.cc',
-          ],
-        }, {
-          'sources': [
-            'udp_file_posix.cc',
-          ],
-        }],
+        'udp_file.cc',
```
kqyang: Move before udp_file.h in line 29

packager/media/file/udp_file.cc
```diff
@@ -6,25 +6,31 @@
 
 #include "packager/media/file/udp_file.h"
 
+#if defined(OS_WIN)
+#include <windows.h>
+#include <ws2tcpip.h>
+#include <memory>
```
kqyang: Why is \<memory\> needed in windows config but not in other configs?

packager/media/file/udp_file.cc
```diff
@@ -128,20 +163,27 @@ bool UdpFile::Open() {
...
     LOG(ERROR) << "Malformed IPv4 address " << options->address();
     return false;
   }
 
+  struct sockaddr_in local_sock_addr = {0};
+  // TODO(kqyang): Support IPv6.
+  local_sock_addr.sin_family = AF_INET;
+  local_sock_addr.sin_port = htons(options->port());
+  if (IsIpv4MulticastAddress(local_in_addr)) {
+    local_sock_addr.sin_addr.s_addr = htonl(INADDR_ANY);
```
kqyang: Will this work on linux?

packager/media/file/udp_file.cc
```diff
@@ -6,25 +6,31 @@
...
 
-// TODO(tinskip): Adapt to work with winsock.
-
 namespace shaka {
 namespace media {
 
 namespace {
 
-const int kInvalidSocket(-1);
+const SOCKET kInvalidSocket(INVALID_SOCKET);
```
kqyang: You may use INVALID_SOCKET directly in the code below and remove the definition here.

packager/media/file/udp_file.cc
capitalm-chen: Yes, I have tested this on both linux and Windows.

packager/media/file/udp_file.cc
capitalm-chen: Hi, I just tested this, and it is not required in either.

packager/media/file/udp_file.cc
```diff
@@ -90,58 +94,93 @@ bool UdpFile::Tell(uint64_t* position) {
...
+                &local_in_addr) != 1) {
     LOG(ERROR) << "Malformed IPv4 address " << options->address();
     return false;
   }
 
+  struct sockaddr_in local_sock_addr = {0};
+  // TODO(kqyang): Support IPv6.
+  local_sock_addr.sin_family = AF_INET;
+  local_sock_addr.sin_port = htons(options->port());
+  if (IsIpv4MulticastAddress(local_in_addr)) {
```
kqyang: Can you assign this to a local variable to avoid duplicate computation below in line 198, e.g.

const bool is_multicast = IsIpv4MulticastAddress(local_in_addr);
...

capitalm-chen: Got it. Thanks a lot.

kqyang:
@capitalm-chen Change is merged. Thanks for making shaka-packager better!

Issue #216, opened 2017-03-23
Sliding window logic active even though -generate_static_mpd is specified


sylt:
## System info
**Operating System**: Ubuntu 14.04
**Shaka Packager Version**: c223bc9-release

## Issue and steps to reproduce the problem

In this issue, I have used the same input stream as in #214 (i.e., a stream of 4096 seconds).

When segmenting my stream and playing it back in the shaka-player, I saw that I always started at position 4096 - 1800 = 2296 seconds = 38:16 minutes into the stream. The reason for this was that my \<SegmentTemplate\> tag contained a startNumber attribute of 230, which will effectively take us 2296 seconds into the stream.

The command I was using is this:

    packager \
        'input=input_video.webm,
         stream=video,
         init_segment=output_video_init.webm,
         segment_template=output_video_$Number$.webm' \
        -generate_static_mpd \
        -mpd_output video.mpd

As can be seen, the "-generate_static_mpd" option was used. The outcome was surprising to me, since I expected a start number of 1, so I took a dive into the code. I found out that this happened due to Representation::SlideWindow() logic being active even though -generate_static_mpd was specified.

As can be seen at the [very first code line in Representation::SlideWindow()](https://github.com/google/shaka-packager/blob/master/packager/mpd/base/mpd_builder.cc#L1311), we only check the length of the time_shift_buffer_depth. However, since that flag will default to 1800 seconds by default, this logic will still be active.

In my local copy, I have solved the issue by adding a check for the MPD type in Representation::SlideWindow(). I.e., if the MPD is static, we skip the SlideWindow logic. It could also be solved in [shaka::media::GetMpdOptions()](https://github.com/google/shaka-packager/blob/master/packager/app/packager_util.cc#L175) by setting the time_shift_buffer_depth to 0.0 when we're generating a static MPD file.

Do you have any preference of how you would like it solved? I could submit a pull request for this as well. Obviously, this can be solved by just adding "-time_shift_buffer_depth 0" in the command and call it a day, but to me it felt like it shouldn't be needed.

Thank you!

kqyang:
Good finding! Thanks for improving shaka-packager. I prefer adding a check for the MPD type in Representation::SlideWindow as the MPD should really be static for "static MPD" :). @rkuroiwa What do you think?

kqyang:
Thanks for fixing the bug! Your effort is really appreciated!

PR #218, opened 2017-03-24
Don't do timeshift SlideWindow() logic for static MPD

sylt:
This is a fix for issue #216 where the Representation::SlideWindow() logic was
still active even though a static MPD was being asked to be generated (using
the option -generate_static_mpd). This could cause static streams longer than
1800 seconds to start playing at (end-position-for-stream - 1800 seconds).

This change is in compliance with the current usage documentation for the
shaka-packager, which suggests that -time_shift_buffer_depth only is relevant
for dynamic media presentations.


rkuroiwa:
lgtm

PR #219, opened 2017-03-24
Add support for skip_encryption stream descriptor

sylt:
This option makes it possible to control encryption for specific streams, i.e. for audio and video separately.

In my set-up, I want to have the ability to control which tracks gets to be encrypted or not (in my case, I wanted the sound to be in clear). I though that others perhaps wanted this capability as well, hence the pull request. 

It's possible though that the maintainers think otherwise :) Let's have a try.

### DIFF
--- a/packager/app/packager_main.cc
+++ b/packager/app/packager_main.cc
@@ -107,7 +107,9 @@ const char kUsage[] =
     "    The group ID for the output stream. For HLS this is used as the\n"
     "    GROUP-ID attribute for EXT-X-MEDIA.\n"
     "  - playlist_name: Required for HLS output.\n"
-    "    Name of the playlist for the stream. Usually ends with '.m3u8'.\n";
+    "    Name of the playlist for the stream. Usually ends with '.m3u8'.\n"
+    "  - skip_encryption=0|1: Optional. Defaults to 0 if not specified. If\n"
+    "    it is set to 1, no encryption of the stream will be made.\n";
 
 const char kMediaInfoSuffix[] = ".media_info";
 
@@ -378,7 +380,7 @@ bool CreateRemuxJobs(const StreamDescriptorList& stream_descriptors,
     handlers.push_back(chunking_handler);
 
     Status status;
-    if (encryption_key_source) {
+    if (encryption_key_source && !stream_iter->skip_encryption) {
       auto new_encryption_options = encryption_options;
       // Use Sample AES in MPEG2TS.
       // TODO(kqyang): Consider adding a new flag to enable Sample AES as we
--- a/packager/app/test/packager_test.py
+++ b/packager/app/test/packager_test.py
@@ -240,6 +240,15 @@ class PackagerAppTest(unittest.TestCase):
     self._VerifyDecryption(self.output[0], 'bear-640x360-a-golden.mp4')
     self._VerifyDecryption(self.output[1], 'bear-640x360-v-golden.mp4')
 
+  def testPackageWithEncryptionOfOnlyVideoStream(self):
+    self.packager.Package(
+        self._GetStreams(['audio,skip_encryption=1', 'video']),
+        self._GetFlags(encryption=True))
+    self._DiffGold(self.output[0], 'bear-640x360-a-golden.mp4')
+    self._DiffGold(self.output[1], 'bear-640x360-v-cenc-golden.mp4')
+    self._DiffGold(self.mpd_output, 'bear-640x360-a-clear-v-cenc-golden.mpd')
+    self._VerifyDecryption(self.output[1], 'bear-640x360-v-golden.mp4')
+
   def testPackageWithEncryptionAndTrickPlay(self):
     self.packager.Package(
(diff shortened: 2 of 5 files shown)


packager/app/packager_main.cc
```diff
@@ -106,7 +106,10 @@ const char kUsage[] =
     "    The group ID for the output stream. For HLS this is used as the\n"
     "    GROUP-ID attribute for EXT-X-MEDIA.\n"
     "  - playlist_name: Required for HLS output.\n"
-    "    Name of the playlist for the stream. Usually ends with '.m3u8'.\n";
+    "    Name of the playlist for the stream. Usually ends with '.m3u8'.\n"
+    "  - skip_encryption: Optional value which specifies if encryption of\n"
+    "    the stream shall be skipped. By default, if encryption is enabled,\n"
```
kqyang: Make it clear that it expects a numeric value instead of "true|false", e.g.

skip_encryption: Optional, if set to nonzero, encryption of the stream shall be skipped. ...

sylt: Yes, good point! I'll have a go at fixing it.

I'm actually considering if one should allow setting skip_encryption=true, since it perhaps reads a bit better. Let me know if you have any particular preference here!

I also have to mention: Thank you for the friendly and good feedback I've gotten from you during the last few days! :) I personally find it very encouraging!

kqyang: I am fine with both. I am slightly leaning towards 0|1 for consistency since we also use 0|1 in udp options.

> I also have to mention: Thank you for the friendly and good feedback I've gotten from you during the last few days! :) I personally find it very encouraging!

No problem. Thanks for your work on improving shaka-packager :-)

kqyang:
@sylt Are you still working on this? Are you still interested in completing this pull request? Thanks.

sylt:
Thanks for the reminder! I must have done something wrong or misunderstood something, but I did update the pull-request with your suggestions in March (as in, the diff looked up-to-date when I checked this PR). I thought that github would send some sort of notification of this being done, but maybe that's not the case. Sigh, I need to read up on this :)

Anyway, I have rebased my change-set, so the pull-request should be up-to-date now (with your suggestions incorporated).

packager/app/test/packager_test.py
```diff
@@ -593,6 +593,16 @@ def testPackageWithLiveProfileAndKeyRotationAndNonDashIfIop(self):
         self.mpd_output,
         'bear-640x360-av-live-cenc-rotation-non-iop-golden.mpd')
 
+  def testPackageWithEncryptionOfOnlyVideoStream(self):
+    self.packager.Package(
+        self._GetStreams(['audio', 'video'],
```
kqyang: Prefer something like 

self.packager.Package(self._GetStreams(['audio,skip_encryption=1','video']), ....)

It is more flexible and probably clearer as well.

The changes in _GetStreams is not needed.

packager/app/test/packager_test.py
```diff
@@ -593,6 +593,16 @@ def testPackageWithLiveProfileAndKeyRotationAndNonDashIfIop(self):
...
+                         skip_encryption_of_first_stream=True),
+        self._GetFlags(encryption=True))
+    self._DiffGold(self.output[0], 'bear-640x360-a-golden.mp4')
+    self._DiffGold(self.output[1], 'bear-640x360-v-cenc-golden.mp4')
+    self._DiffGold(self.mpd_output, 'bear-640x360-a-clear-v-venc-golden.mpd')
```
kqyang: s/venc/cenc/

packager/app/test/packager_test.py
```diff
@@ -593,6 +593,16 @@ def testPackageWithLiveProfileAndKeyRotationAndNonDashIfIop(self):
         self.mpd_output,
         'bear-640x360-av-live-cenc-rotation-non-iop-golden.mpd')
 
+  def testPackageWithEncryptionOfOnlyVideoStream(self):
```
kqyang: Can you move this function after line 242, i.e. after testPackageWithEncryption?

packager/app/packager_main.cc
```diff
@@ -107,7 +107,10 @@ const char kUsage[] =
     "    The group ID for the output stream. For HLS this is used as the\n"
     "    GROUP-ID attribute for EXT-X-MEDIA.\n"
     "  - playlist_name: Required for HLS output.\n"
-    "    Name of the playlist for the stream. Usually ends with '.m3u8'.\n";
+    "    Name of the playlist for the stream. Usually ends with '.m3u8'.\n"
+    "  - skip_encryption: Optional. If this is set to a non-zero value,\n"
```
kqyang: This description does not match what is implemented, rephrase it as:

- skip_encryption=0|1: Optional. Default to 0 if not specified. If it is set to 1, ...

sylt:
All comments accepted/handled! Much better :) Pull request has been updated. Let's see!

kqyang:
@sylt LGTM. Thanks for coming up with the idea and working on this! Your contribution is appreciated!

kqyang:
Will merge once continuous integration tests pass.

PR #226, opened 2017-04-11
[fix] kAdtsNumChannelsTable

zevlg:
Standard specifies 3 channels (from,left,rigth) for channel configuration value equal to 3


zevlg:
Just signed CLA

kqyang:
@zevlg Thanks for fixing this issue!

Issue #258, opened 2017-07-28
Fairplay support patch


Palethorn:
**Shaka Packager Version**: b451d3a7ca6f454b4905043d56dd9620247d8c71

**Packager Command**:
```./packager \ 'input=udp://127.0.0.1:11000,stream=3,segment_template=fairplay_video-$Number$.ts,playlist_name=video.m3u8' \
 'input=udp://127.0.0.1:11000,stream=0,segment_template=fairplay_audio1-$Number$.ts,playlist_name=audio1.m3u8,hls_group_id=audio1,hls_name=EN' \
  --enable_fixed_key_encryption \
  --clear_lead 0 \
  --key ... \
  --key_id ... \
  --iv ... \
  --hls_playlist_type LIVE  \
  --hls_master_playlist_output="playlist.m3u8"
```
I would like to properly patch shaka-packager to produce SAMPLE-AES crypted content. The end result is having this kind of header in HLS chunklist:
```#EXT-X-KEY:METHOD=SAMPLE-AES,URI="skd://entry",KEYFORMAT="com.apple.streamingkeydelivery",KEYFORMATVERSIONS="1"```
instead of this one which requires transfering key in clear format:
```#EXT-X-KEY:METHOD=SAMPLE-AES,URI="data:text/plain;base64,...",IV=0x...,KEYFORMAT="identity"```
I already have it working by hardcoding the header into media_playlist. I would like to submit a proper patch for this. Any pointers on how should I approach this task?

Palethorn:
Would I be able to add another FOURCC to packager/media/base/fourccs.h, for example FOURCC_cbcf for fairplay, and then perform necessary checks?

kqyang:
@Palethorn Thanks for helping on improving Shaka Packager.

I suggest adding a new key system instead of using fourcc to signal fairplay since fairplay is really a key system protocol while fourcc is used to indicate the encryption format.

(1) Define a system id for fairplay:

https://github.com/google/shaka-packager/blob/master/packager/media/base/fixed_key_source.h#L22

Fairplay does not an official system id defined right now. We can use an arbitrary unique 16-byte string here.

(2) Add fairplay key system to encryption_key->key_system_info when appropriate flag is defined:

https://github.com/google/shaka-packager/blob/master/packager/media/base/fixed_key_source.cc#L115

(3) Add appropriate logic below to notify encryption to the media playlists:

[...]

PR #260, opened 2017-08-01
#258 Fairplay key system support

Palethorn:
I moved base64 encode of the URI to the each individual protection system case because the old base64 format was messing with the fairplay key system on safari. I think it always expects it to be in URI format.


Palethorn:
I signed it!

kqyang:
Optional, but feel free to add your name / your company's name to https://github.com/google/shaka-packager/blob/master/CONTRIBUTORS and https://github.com/google/shaka-packager/blob/master/AUTHORS file if you want.

packager/media/base/fixed_key_source.h
```diff
@@ -23,6 +23,10 @@ const uint8_t kCommonSystemId[] = {0x10, 0x77, 0xef, 0xec, 0xc0, 0xb2,
                                    0x4d, 0x02, 0xac, 0xe3, 0x3c, 0x1e,
                                    0x52, 0xe2, 0xfb, 0x4b};
 
+const uint8_t kFairplaySystemId[] = {0x29, 0x70, 0x1F, 0xE4, 0x3C, 0xC7,
+                                     0x4A, 0x34, 0x8C, 0x5B, 0xAE, 0x90,
+                                     0xC7, 0x43, 0x9A, 0x47};
```
Palethorn: Yes, no problem.

packager/hls/base/simple_hls_notifier.cc
```diff
@@ -357,12 +368,28 @@ bool SimpleHlsNotifier::NotifyEncryptionUpdate(
...
+  if (IsFairplaySystemId(system_id)) {
+    // Use key_id as the key_uri. The player needs to have custom logic to
+    // convert it to the actual key url.
+    std::string key_uri_data;
+    std::string key_uri_data_base64;
+    key_uri_data.assign(key_id.begin(), key_id.end());
+    base::Base64Encode(key_uri_data, &key_uri_data_base64);
+    NotifyEncryptionToMediaPlaylist(encryption_method,
+                                    kUriFairplayBase64Prefix + key_uri_data_base64, std::vector<uint8_t>(), std::vector<uint8_t>(),
+                                    "com.apple.streamingkeydelivery", "1", media_playlist.get());
```
Palethorn: My editor is set to 120 so I haven't notices. Will do.

packager/hls/base/simple_hls_notifier.cc
```diff
@@ -41,6 +42,22 @@ bool IsCommonSystemId(const std::vector<uint8_t>& system_id) {
          std::equal(system_id.begin(), system_id.end(), media::kCommonSystemId);
 }
 
+std::string Base64EncodeUri(const std::string& prefix,
```
rkuroiwa: There is base::Base64UrlEncode() which does base64-url-encoding
https://en.wikipedia.org/wiki/Base64#URL_applications

So I prefer renaming this to Base64EncodeData().

packager/hls/base/simple_hls_notifier.cc
```diff
@@ -41,6 +42,22 @@ bool IsCommonSystemId(const std::vector<uint8_t>& system_id) {
...
+                            const std::string& data) {
+    std::string data_base64;
+    base::Base64Encode(data, &data_base64);
```
rkuroiwa: As mentioned above there is also base 64 encoding for URLs.
I don't quite remember why I didn't do it for the identity key impl, but I think most people prefer base 64 URL encoding. WDYT?

Palethorn: Not really sure. I went with the assumption that there won't be any weird characters in key_id. I think it could either be 16byte UUID or a numeric id, and it's sent as a POST parameter in player example so there is no need to be urlencoded. Or is there?

kqyang: @rkuroiwa That is a good point. @Palethorn key_id can include arbitrary bytes, so it is possible that it may include illegal url chars after base64 conversion. (UUID is actually a form of hex representation of the data) Can you help us change the code here to use [Base64UrlEncode](https://cs.chromium.org/chromium/src/base/base64url.h?rcl=2491c29c85d9a50c2b6f51aece17343aa71061c2&l=30)? Thanks.

Palethorn: OK, I'll replace the Base64Encode with Base64UrlEncode if everyone agrees.

kqyang: No, it should apply to Widevine and identity key systems as well. It is probably better to change it in a separate patch, so I think it is fine to leave it as is in this patch. Feel free to submit another pull request on this if you want; if not, we'll address it :)

Palethorn: OK, I'll leave it as is for now.

Palethorn:
I'll try and add unit test for simple_hls_notifier.

Palethorn:
I've added the necessary test.

packager/hls/base/simple_hls_notifier_unittest.cc
```diff
@@ -645,6 +650,43 @@ TEST_F(SimpleHlsNotifierTest, EncryptionScheme) {
...
+TEST_F(SimpleHlsNotifierTest, NotifyEncryptionUpdateFairplay) {
+  // Pointer released by SimpleHlsNotifier.
+  MockMediaPlaylist* mock_media_playlist =
+      new MockMediaPlaylist(kLivePlaylist, "playlist.m3u8", "", "");
+  SimpleHlsNotifier notifier(kLivePlaylist, kTestTimeShiftBufferDepth,
+                             kTestPrefix, kAnyOutputDir, kMasterPlaylistName);
+  const uint32_t stream_id =
+      SetupStream(kSampleAesProtectionScheme, mock_media_playlist, &notifier);
+  const std::vector<uint8_t> key_id(16, 0x00);
+  const std::vector<uint8_t> iv(16, 0x45);
```
kqyang: Remove. Not used.

packager/hls/base/simple_hls_notifier_unittest.cc
```diff
@@ -645,6 +650,43 @@ TEST_F(SimpleHlsNotifierTest, EncryptionScheme) {
       stream_id, key_id, common_system_id_, iv, dummy_pssh_data));
 }
 
+// Verify that the Fairplay systemID is correctly handled when constructing
+// encryption info
```
kqyang: nit: Add a period in the end.

packager/hls/base/simple_hls_notifier_unittest.cc
```diff
@@ -645,6 +650,43 @@ TEST_F(SimpleHlsNotifierTest, EncryptionScheme) {
...
+TEST_F(SimpleHlsNotifierTest, NotifyEncryptionUpdateFairplay) {
+  // Pointer released by SimpleHlsNotifier.
+  MockMediaPlaylist* mock_media_playlist =
+      new MockMediaPlaylist(kLivePlaylist, "playlist.m3u8", "", "");
+  SimpleHlsNotifier notifier(kLivePlaylist, kTestTimeShiftBufferDepth,
+                             kTestPrefix, kAnyOutputDir, kMasterPlaylistName);
+  const uint32_t stream_id =
+      SetupStream(kSampleAesProtectionScheme, mock_media_playlist, &notifier);
+  const std::vector<uint8_t> key_id(16, 0x00);
```
kqyang: Suggest using non zero id instead, e.g. 

const std::vector<uint8_t> key_id(16, 0x12);

packager/hls/base/simple_hls_notifier_unittest.cc
```diff
@@ -645,6 +650,43 @@ TEST_F(SimpleHlsNotifierTest, EncryptionScheme) {
...
+  const std::vector<uint8_t> iv(16, 0x45);
+  std::vector<uint8_t> pssh_box = { 0x00, 0x00, 0x00, 0x34, 0x70, 0x73, 0x73,
+                                    0x68, 0x01, 0x00, 0x00, 0x00, 0x29, 0x70,
+                                    0x1F, 0xE4, 0x3C, 0xC7, 0x4A, 0x34, 0x8C,
+                                    0x5B, 0xAE, 0x90, 0xC7, 0x43, 0x9A, 0x47,
+                                    0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00,
+                                    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+                                    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+                                    0x00, 0x00, 0x00 };
```
kqyang: I don't think it matters. You may replace it with a dummy value, e.g.

const std::vector<uint8_t> dummy_pssh_data(10, 'p');

kqyang:
This is merged. @Palethorn Thanks for working on this.

Palethorn:
No problem. This was a good experience.

PR #280, opened 2017-10-28
Minor fixes (semantic proposal)

kant:


kqyang:
No longer applicable.

PR #281, opened 2017-10-28
Minor fix (proposal)

kant:


vaage:
@kqyang, do you see any issues with this change?

kqyang:
Not really, though I'd prefer merging this PR with #282.

Issue #305, opened 2017-12-15
Docs: Issues with HLS documentation


ithinkihaveacat:
(1) One of the [HLS examples in the docs](https://google.github.io/shaka-packager/html/tutorials/hls.html#examples) appears to be incorrect; in the "TS output" section, the `output` options should instead be `segment_template`.

Making this change fixes the error: "Please specify segment_template. Single file TS output is not supported."

<strike>
(2) The other issue with this example is that it says it uses "the H264 streams created in Media Encoding" however the [media encoding sample code](https://google.github.io/shaka-packager/html/tutorials/encoding.html) does not create the necessary `*.m3u8` files referenced by `playlist_name`. These need to be adjusted so that they read:

```sh
$ ffmpeg -i original.mp4 -c:a copy \
  -vf "scale=-2:360" \
  -c:v libx264 -profile:v baseline -level:v 3.0 \
  -x264opts scenecut=0:open_gop=0:min-keyint=72:keyint=72 \
  -minrate 600k -maxrate 600k -bufsize 600k -b:v 600k \
  -y h264_baseline_360p_600.mp4
  h264_360p.m3u8 # ADD THIS LINE
```
</strike>

(2) Document that you probably don't want `playlist_name` to include a path.

If a path is inadvertently specified, a confusing and misleading error is generated; this command

```sh
$ packager in=public/foo/360p_600.mp4,stream=video,segment_template=public/foo/360p_\$Number\$.ts,playlist_name=public/foo/360p.m3u8 --hls_master_playlist_output public/foo/master.m3u8
```

unexpectedly fails with the error: "Failed to open file public/foo/public/foo/360p.m3u8".

I think what happened is that it tried to create this file without creating the intermediate paths, which fails.

ithinkihaveacat:
Actually, I'm not sure this works the way I think it works. Does the stream descriptor `playlist_name` instruct shaka packager to *create* a playlist file? Or is it supposed to already exist? If it already exists, what is the input file (`in`) used for?

kqyang:
Thanks for bringing up the issues.

For (1), we will address it ASAP.

For (2), playlist_name was initially designed to be just a name instead of a path - it will generated in the same directory as master playlist; we discovered later that user might want to put playlists for different streams in different directories thus we need to support playlist_name being a path. It is still relative to master playlist path to be backward compatible.

I agree that it is a bit confusing. Maybe we should make playlist_name a path relative to current directory instead of master playlist.

[...]

PR #307, opened 2017-12-18
Update playlist_name docs

ithinkihaveacat:
Make it clear that shaka packager will create the playlist file. (See #305.)


kqyang:
@ithinkihaveacat Thanks for the fix. Can you update https://github.com/google/shaka-packager/blob/master/packager/app/packager_main.cc#L99 as well?

ithinkihaveacat:
@kqyang Thanks, I updated packager_main.cc.

kqyang:
Great, thanks. I have merged the CL.

PR #310, opened 2017-12-19
Clean up #ifndef in packager/media.

hchen1506:
Run followings:
find packager -type f -exec sed -i -e 's/#ifndef\ MEDIA_/#ifndef\ PACKAGER_MEDIA_/g' {} \;
find packager -type f -exec sed -i -e 's/#define\ MEDIA_/#define\ PACKAGER_MEDIA_/g' {} \;
find packager -type f -exec sed -i -e 's/#endif\ \ \/\/\ MEDIA_/#endif\ \ \/\/\ PACKAGER_MEDIA_/g' {} \;

### DIFF
--- a/packager/media/base/aes_encryptor.h
+++ b/packager/media/base/aes_encryptor.h
@@ -6,8 +6,8 @@
 //
 // AES Encryptor implementation using openssl.
 
-#ifndef MEDIA_BASE_AES_ENCRYPTOR_H_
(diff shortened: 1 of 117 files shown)


hchen1506:
Note that the total changed line is 337 less than 3 * 117  = 351 since some files do not have "// MEDIA..." after #endif.

kqyang:
Thanks. Merged.

Issue #332, opened 2018-02-22
Source Specific Multicast Not Supported (?)


GScharfofMars:
## System info
**Operating System**: Ubuntu 16.04 LTS
**Shaka Packager Version**: v2.0.0-1d6a2de-release

## Issue and steps to reproduce the problem
The following  is an address redacted snippet from a test script. The redacted address streams have been shown to exist and be accessible to the host via other programs such as vlc and multicat. 

**Packager Command**:
STREAM_SRC=XX.XX.XX.XX
INTERFACE=YY.YY.YY.YY
STREAM_ADDR=ZZ.ZZ.ZZ.ZZ
STREAM_PORT=PPPPP

./shaka-packager --v=1 \
in=upd://${STREAM_ADDR}:${STREAM_PORT}?source=${STREAM_SRC}?interface=${INTERFACE}?reuse=1,stream=audio,init_segment=audio_init.mp4,segment_template='audio_$Number$.m4s',playlist_name=audio.m3u8,hls_group_id=audio,hls_name=ENGLISH \
in=upd://${STREAM_ADDR}:${STREAM_PORT}?source=${STREAM_SRC}?interface=${INTERFACE}?reuse=1,stream=video,init_segment=h264_init.mp4,segment_template='h264_$Number$.m4s',playlist_name=h264_.m3u8 \
--hls_master_playlist_output h264_master.m3u8

The result is:
./testPackager.sh 
[0222/181854:INFO:demuxer.cc(89)] Demuxer::Run() on file 'upd://REDACTED?source=REDACTED?interface=REDACTED?reuse=1'.
[0222/181854:INFO:demuxer.cc(161)] Initialize Demuxer for file 'upd://REDACTED?source=REDACTED?interface=REDACTED?reuse=1'.
[0222/181854:VERBOSE1:status.cc(72)] 5 (FILE_FAILURE): Cannot open file for reading upd://REDACTED?source=REDACTED?interface=REDACTED?reuse=1
[0222/181854:ERROR:packager_main.cc(470)] Packaging Error: 5 (FILE_FAILURE): Cannot open file for reading upd://REDACTED?source=REDACTED?interface=REDACTED?reuse=1

<Please provide any additional information below.>
https://google.github.io/shaka-packager/html/tutorials/live.html indicates that "source" is an optional clause for the upd "file" option. 

**Additional Info**:
[...]

PR #339, opened 2018-03-03
Source Specific Multicast support

fatpelt:
Issue Number #332

### DIFF
--- a/docs/source/options/udp_file_options.rst
+++ b/docs/source/options/udp_file_options.rst
@@ -8,10 +8,15 @@ options:
 
     Allow or disallow reusing UDP sockets.
 
-:interface=<addr>, source=<addr>:
+:interface=<addr>:
 
-    Multicast group interface address. Only the packets sent to this address is
+    Multicast group interface address. Only the packets sent to this address are
     received. Default to "0.0.0.0" if not specified.
+
+:source=<addr>:
+
+    Multicast source ip address.  Only the packets sent from this source address
+    are received.  Defaults to "0.0.0.0" if not specified.
 
 :timeout=<microseconds>:
 
--- a/docs/source/options/udp_file_options.rst
+++ b/docs/source/options/udp_file_options.rst
@@ -15,8 +15,8 @@ options:
 
 :source=<addr>:
 
-    Multicast source ip address.  Only the packets sent from this source address
-    are received.  Defaults to "0.0.0.0" if not specified.
+    Multicast source ip address. Only the packets sent from this source address
+    are received. Defaults to "0.0.0.0" if not specified.
 
 :timeout=<microseconds>:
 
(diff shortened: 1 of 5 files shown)


fatpelt:
i have validated it with a source yes.  we use SSM multicast pretty extensively at work. SSM is a pretty standard acronym, but i don't mind changing the references per your thoughts.  I thought about using the empty string, and don't mind. it wasn't immediately clear to me if there was a project standard on that.  also the spacing. 

i'll get all these fixed up and an updated pull request issued.  

i also realized that i need to update docs.  i'm assuming that docs/source/ is what's behind https://google.github.io/shaka-packager/html/tutorials/live.html ?

kqyang:
Great. Thanks.

Yes, I prefer to spell out SSM, which makes it easier for people like me to read the code :). As for the empty string, I am fine with both options.

We do have a strict requirement on the coding styles, including alignment, spaces etc. We follow Chromium coding style: https://chromium.googlesource.com/chromium/src/+/master/styleguide/c++/c++.md. This helps code readability.

> i also realized that i need to update docs. i'm assuming that docs/source/ is what's behind https://google.github.io/shaka-packager/html/tutorials/live.html ?

Good catch. Yes, these are the two files you may want to touch: docs/source/options/udp_file_options.rst and docs/source/tutorials/live.rst.

fatpelt:
ok.  i *think* i got them all.  the code i modified shows clean in the format check, but i'm betting i missed something.  also, sorry for the multiple commits rather than a rollback.  i'm still figuring this git thing out (i'm a cvs/svn old hat)

docs/source/options/udp_file_options.rst
```diff
@@ -8,11 +8,16 @@ options:
...
-:interface=<addr>, source=<addr>:
+:interface=<addr>:
 
-    Multicast group interface address. Only the packets sent to this address is
+    Multicast group interface address. Only the packets sent to this address are
     received. Default to "0.0.0.0" if not specified.
 
+:source=<addr>:
+
+    Multicast source ip address.  Only the packets sent from this source address
```
kqyang: Remove an extra space between sentences. We usually have only one space between sentences. similar below.

packager/file/udp_file.cc
```diff
@@ -204,22 +204,54 @@ bool UdpFile::Open() {
   }
 
   if (is_multicast) {
-    struct ip_mreq multicast_group;
-    multicast_group.imr_multiaddr = local_in_addr;
+    if (options->is_source_specific_multicast()) {
+      // user has specified they want a specific source sender.  let's use that
+      // here
```
kqyang: remove the comment. not useful here.

kqyang:
Don't worry about the multiple commits. Github provides a handy function to squash and merge the changes into one.

PR #350, opened 2018-03-11
Specify Initial Sequence Number

sammirata:
With this enhancement, you can now specify the initial sequence number
to be used on the generated segments when calling the packager.
With the old implementation, it was always starting with "1".

### DIFF
--- a/packager/app/muxer_flags.cc
+++ b/packager/app/muxer_flags.cc
@@ -50,3 +50,7 @@ DEFINE_bool(mp4_use_decoding_timestamp_in_timeline,
             "be used when generating media timeline, e.g. timestamps in sidx "
             "and mpd. This is to workaround a Chromium bug that decoding "
             "timestamp is used in buffered range, https://crbug.com/398130.");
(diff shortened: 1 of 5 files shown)


packager/app/muxer_flags.cc
```diff
@@ -50,3 +50,7 @@ DEFINE_bool(mp4_use_decoding_timestamp_in_timeline,
             "be used when generating media timeline, e.g. timestamps in sidx "
             "and mpd. This is to workaround a Chromium bug that decoding "
             "timestamp is used in buffered range, https://crbug.com/398130.");
+DEFINE_int32(mp4_sequence_number_initial,
+             1,
```
vaage: Could you rephrase the description to say something like "Sets the initialize sequence number in the mp4 moof->mfhd. Otherwise defaults to 1.".

vaage: I find this name a little awkward, could you rename this to "mp4_initial_sequence_number"?

kqyang:
Moved to in #519

Issue #419, opened 2018-06-26
Windows build failed with Ninja


Jason239:
I don't know if it's right to ask a question here.
I have a question about building on Windows
I was succeeded to build binary immediately on Ubuntu.
but on Windows, failed.
After entering the command 'ninja -C out \ Release_x64', an error occurs during the build with 'The warn is treated as an error and no generated 'object' file exists.' message.
especially, most of this errors were files in 'boringssl'.
I tried to edit .gyp scripts, but it didn't work.

what should I do?

kqyang:
@LeeSangGoo Thanks for filing the issue. Can you provide more information? For example, (1) Windows version. (2) Packager source code revision. (3) The error messages.

LeeSangGoo:
@kqyang  Thanks for your quick reply. 
This is the execution environment.

(1) Windows version : Windows 10 Pro, x64
(2) Packager source code revision : Latest version(6b4a75b6bacdac6b0b9f598c365c8f84f0a7fd70)
(3) Visual Studio version : 2015 Professional v14.0.25431.01 Update 3
(4) The error messages : 
[failed_log.txt](https://github.com/google/shaka-packager/files/2140434/failed_log.txt)

Please tell me if you need any more information.

kqyang:
@LeeSangGoo Thanks for the info! 

Here is what I saw from the log:

```bash
d:\works\shaka_packager\src\packager\third_party\boringssl\src\crypto\internal.h: error C2220: °æ°í°¡ ¿À·ù·Î Ã³¸®µÇ¾î »ý
¼ºµÈ 'object' ÆÄÀÏÀÌ ¾ø½À´Ï´Ù.
d:\works\shaka_packager\src\packager\third_party\boringssl\src\crypto\internal.h: warning C4819: ÇöÀç ÄÚµå ÆäÀÌÁö(949)¿¡
¼­ Ç¥½ÃÇÒ ¼ö ¾ø´Â ¹®ÀÚ°¡ ÆÄÀÏ¿¡ µé¾î ÀÖ½À´Ï´Ù. µ¥ÀÌÅÍ°¡ ¼Õ½ÇµÇÁö ¾Ê°Ô ÇÏ·Á¸é ÇØ´ç ÆÄÀÏÀ» À¯´ÏÄÚµå Çü½ÄÀ¸·Î ÀúÀåÇÏ½Ê½Ã¿À.
...
[...]

PR #427, opened 2018-07-09
Added 4267, 4819 in each .gyp files for Windows build

Jason239:
Modifying only the common.gyp file does not fix the problem.

So I fixed each modules .gyp file which error occurred.

### DIFF
--- a/packager/file/file.gyp
+++ b/packager/file/file.gyp
@@ -12,6 +12,7 @@
     {
       'target_name': 'file',
       'type': '<(component)',
+      'msvs_disabled_warnings': [ 4267, 4819, ],
       'sources': [
         'callback_file.cc',
         'callback_file.h',
@@ -42,6 +43,7 @@
     {
       'target_name': 'file_unittest',
       'type': '<(gtest_target_type)',
+      'msvs_disabled_warnings': [ 4267, 4819, ],
       'sources': [
         'callback_file_unittest.cc',
         'file_unittest.cc',
--- a/packager/hls/hls.gyp
+++ b/packager/hls/hls.gyp
@@ -12,6 +12,7 @@
     {
       'target_name': 'hls_builder',
       'type': '<(component)',
+      'msvs_disabled_warnings': [ 4267, 4819, ],
       'sources': [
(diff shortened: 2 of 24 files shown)


LeeSangGoo:
I signed it

kqyang:
@LeeSangGoo Thanks for the pull request!

Can you help try if it works with just the below change:

```diff
diff --git c/packager/common.gypi w/packager/common.gypi
index 6f61d74ec5..de51955881 100644
--- c/packager/common.gypi
+++ w/packager/common.gypi
@@ -42,16 +42,7 @@
             '-Wno-reserved-user-defined-literal',
           ],
         },
-        'conditions': [
-          ['OS == "win"', {
-            'msvs_settings': {
-              'VCCLCompilerTool': {
-                'WarnAsError': 'true',
-                'DisableSpecificWarnings': ['4125']
-              },
-            },
-          }],
-        ],
+        'msvs_disabled_warnings': [ 4125, 4819, ],
       }, {
         # We do not have control over non-shaka code. Disable some warnings to
         # make build pass.
@@ -61,6 +52,7 @@
             '-Wno-unguarded-availability',
           ],
         },
+        'msvs_disabled_warnings': [ 4819, ],
         'conditions': [
           ['clang==0', {
             'cflags': [
```

Thanks for your help!

LeeSangGoo:
Oh I'm sorry, I thought I had already tried it, but I guess I did something wrong.
It succeeds even if I modify only that part.
I will withdraw this pull request.

Thank you very much!

LeeSangGoo:
I would appreciate if you apply that part.

kqyang:
@Jason239 There is nothing wrong in your side :) The patch I gave above is a little bit different to the patch I gave earlier in #419 (I thought the problem was only with third_party builds, but I learned from your patch here that some internal build targets were affected too - we'll dig into it in the future as it should not happen). And yes, we will submit the fix soon. Thanks for your help in verifying the fix!

Jason239:
Ah, I see. That's good :)
So now this issue will be closed.

Thanks !

PR #475, opened 2018-09-13
Clear up condition in AesPatternCryptor

axden:
The condition was the same in meaning,
but a bit confusing in syntax.


axden:
I signed it!

PR #510, opened 2018-11-22
Specify Initial Sequence Number and Use the ISO639 Language Descriptor when available on the input

sammirata:
Here is my original pull request from March with the changes that were suggested. I also have an additional commit for using the ISO639 Language Descriptor when available on the input (the code for this latter fix was found on the issues thread).


sammirata:
Any word on this request? Can you merge the first commit and tell me how to fix the second one?

kqyang:
@sammirata The change looks good to me. Do you mind push the first commit to the original pull request: https://github.com/google/shaka-packager/pull/350 and just keep the second commit in this pull request?

sammirata:
I do not know how to ... The branch was so old that I created a fresh copy on both my machine and github and that is why the two new commits ended up in a new pull request. #350 shows "sammirata wants to merge 1 commit into google:master from unknown repository"
If you guide me, I will be happy to move the commit.

PR #517, opened 2018-11-28
WebVtt parser: Do not fail when we do not recognize a block

sammirata:
WebVtt parser: Do not fail when we do not recognize a block, just skip it instead

### DIFF
--- a/packager/media/formats/webvtt/webvtt_parser.cc
+++ b/packager/media/formats/webvtt/webvtt_parser.cc
@@ -175,7 +175,8 @@ bool WebVttParser::Parse() {
 
     LOG(ERROR) << "Failed to determine block classification:\n"
                << BlockToString(block.data(), block.size());
-    return false;
+    // Ignore the unknown block
+    continue;
   }
 
   return keep_reading_;


packager/media/formats/webvtt/webvtt_parser.cc
```diff
@@ -175,7 +175,8 @@ bool WebVttParser::Parse() {
 
     LOG(ERROR) << "Failed to determine block classification:\n"
```
kqyang: Can you change it to:

```c++
LOG(WARNING) << "Ignore unrecognized block:\n" 
             << BlockToString(block.data(), block.size());
continue;
```
You'll also need to update the test, e.g. the WebVttParserTest.FailToParseCueWithArrowId. You can change it to something like this:

```
TEST_F(WebVttParserTest, IgnoreUnrecognizedBlock) {
  const char* text =
      "WEBVTT\n"
      "\n"
      "-->\n"
      "00:01:00.000 --> 01:00:00.000\n"
      "subtitle\n";

  ASSERT_NO_FATAL_FAILURE(SetUpAndInitializeGraph(text));

  ASSERT_OK(parser_->Run());
}
```

joeyparrish:
Closing due to lack of response.

PR #611, opened 2019-07-11
no else return

Korilakkuma:
### This PR will ...

This PR will unify "no else return".

- Unify code format
- Low nest level

### DIFF
--- a/packager/hls/base/simple_hls_notifier.cc
+++ b/packager/hls/base/simple_hls_notifier.cc
@@ -184,14 +184,15 @@ base::Optional<MediaPlaylist::EncryptionMethod> StringToEncryptionMethod(
     const std::string& method) {
   if (method == "cenc") {
     return MediaPlaylist::EncryptionMethod::kSampleAesCenc;
-  } else if (method == "cbcs") {
+  }
+  if (method == "cbcs") {
     return MediaPlaylist::EncryptionMethod::kSampleAes;
-  } else if (method == "cbca") {
+  }
+  if (method == "cbca") {
     // cbca is a place holder for sample aes.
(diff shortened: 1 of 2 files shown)


kqyang:
@Korilakkuma Thanks for the pull request, but can you please explain the background of this change? What problem is this CL trying to address?

Korilakkuma:
@kqyang Thanks for comment.

> the background of this change

Because it seems using a early return, I fixed code that does not using a early return.

> What problem is this CL trying to address?

This PR will unify coding style and improve readability by low nest level..

packager/hls/base/simple_hls_notifier.cc
```diff
@@ -380,9 +381,8 @@ bool SimpleHlsNotifier::NotifyNewSegment(uint32_t stream_id,
         if (!WriteMediaPlaylist(master_playlist_dir_, playlist))
           return false;
       }
-    } else {
-      if (!WriteMediaPlaylist(master_playlist_dir_, media_playlist.get()))
-        return false;
+    } else if (!WriteMediaPlaylist(master_playlist_dir_, media_playlist.get())) {
+      return false;
```
kqyang: I'd prefer keeping the original style for this one, which is more readable.

Korilakkuma:
@kqyang 

> I don't see a semantics change anywhere. Can you clarify what you mean by "early return"?

Sorry, I mistaken. Correctly, not "early return" but "no else return".

``` C++
if (x) {
  return y;
} else {
  return z;
}
```

->

```C++
if (x) {
  return y;
}

return z;
```

kqyang:
Thanks @Korilakkuma. The change has been merged.

PR #624, opened 2019-07-18
Cleanup condition on `bind` error

Korilakkuma:
### This PR will ...

This PR will cleanup condition for `bind` (`bind` returns `-1` on error)

### DIFF
--- a/packager/file/udp_file.cc
+++ b/packager/file/udp_file.cc
@@ -199,7 +199,7 @@ bool UdpFile::Open() {
 
   if (bind(new_socket.get(),
            reinterpret_cast<struct sockaddr*>(&local_sock_addr),
-           sizeof(local_sock_addr))) {
+           sizeof(local_sock_addr)) < 0) {
     LOG(ERROR) << "Could not bind UDP socket";
     return false;
   }


Korilakkuma:
@kqyang 

Thanks for comment.

- This PR will fix condition for bind
  - This is obvious
- This PR will add shutdown to destructor
  - The connection is not closed by `close`. `shutdown` destroys the all of connection

kqyang:
> The connection is not closed by close

That is not true. Every documentation I found suggested that close() does shutdown internally: 
[1] https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-closesocket,
[2] https://www.ibm.com/support/knowledgecenter/en/SSB23S_1.1.0.15/gtpc2/cpp_close.html

There are also mentions that calling shutdown explicitly on UDP socket is not recommended: [3] https://lists.apple.com/archives/macnetworkprog/2008/Jan/msg00002.html.

Korilakkuma:
Sorry ... I mistaken. The connection-oriented protocol like TCP needs `shutdown`. But, the connectionless-oriented protocol like UDP does not need `shutdown`. Therefore, this PR does not need `shutdown`.

packager/file/udp_file.cc
```diff
@@ -199,7 +199,7 @@ bool UdpFile::Open() {
 
   if (bind(new_socket.get(),
            reinterpret_cast<struct sockaddr*>(&local_sock_addr),
-           sizeof(local_sock_addr))) {
+           sizeof(local_sock_addr)) == -1) {
```
kqyang: Can you use "< 0" for consistency?

Also, just want to clarify my understanding, this is actually a clean up instead of a fix as it does not change the actual semantics. Is my understanding correct?

Korilakkuma: > Can you use "< 0" for consistency?

Yes. I can.

> this is actually a clean up instead of a fix as it does not change the actual semantics.

Yes. Because `bind` returns `0` on success, current condition returns `false`.

kqyang:
@Korilakkuma The code is now merged. Thanks for improving Shaka Packager.

PR #630, opened 2019-08-08
Reduce copying ADTS header

Korilakkuma:
### This PR will ...

This PR will use copied ADTS header.

This PR is needed to reduce unnecessary copying ADTS header.

### DIFF
--- a/packager/media/codecs/aac_audio_specific_config.cc
+++ b/packager/media/codecs/aac_audio_specific_config.cc
@@ -162,6 +162,33 @@ bool AACAudioSpecificConfig::ConvertToADTS(std::vector<uint8_t>* buffer) const {
   return true;
 }
 
+std::vector<uint8_t> AACAudioSpecificConfig::GetAACwithADTSHeader(const uint8_t* data, size_t length) const {
+  DCHECK(audio_object_type_ >= 1 && audio_object_type_ <= 4 &&
+         frequency_index_ != 0xf && channel_config_ <= 7);
+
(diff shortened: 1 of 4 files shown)


packager/media/codecs/aac_audio_specific_config.cc
```diff
@@ -162,6 +162,31 @@ bool AACAudioSpecificConfig::ConvertToADTS(std::vector<uint8_t>* buffer) const {
   return true;
 }
 
+std::vector<uint8_t> AACAudioSpecificConfig::GetADTSHeader(std::vector<uint8_t>* buffer, std::vector<uint8_t>::size_type length) const {
+  size_t size = static_cast<size_t>(length) + kADTSHeaderSize;
+
+  DCHECK(audio_object_type_ >= 1 && audio_object_type_ <= 4 &&
+         frequency_index_ != 0xf && channel_config_ <= 7);
+
+  std::vector<uint8_t> adts(*buffer);
```
kqyang: This is worse than the original implementation. 

There is an implicit copying in line 171 and another implicit copying in line 179. Plus another copying in line 120 of pes_packet_generator.cc, there are a total of 3 copying. While in the original code, there are two copying.

You can reduce the copying from 2 to 1 by

(1) passing sample.data() and sample.size() to GetADTSHeader function.
(2) Construct the vector with the correct size: std::vector<uint8_t> adts(kADTSHeaderSize + length); 
(3) Append the data after line 188: adts.insert(adts.end(), data, data + length)

PR #633, opened 2019-08-14
Reduce copying ADTS header

Korilakkuma:
### This PR will ...

This PR will use copied ADTS header.

This PR is needed to reduce unnecessary copying ADTS header.

### DIFF
--- a/packager/media/codecs/aac_audio_specific_config.cc
+++ b/packager/media/codecs/aac_audio_specific_config.cc
@@ -137,27 +137,28 @@ bool AACAudioSpecificConfig::Parse(const std::vector<uint8_t>& data) {
          channel_config_ <= 7;
 }
 
-bool AACAudioSpecificConfig::ConvertToADTS(std::vector<uint8_t>* buffer) const {
-  size_t size = buffer->size() + kADTSHeaderSize;
-
+bool AACAudioSpecificConfig::ConvertToADTS(std::vector<uint8_t>& audio_frame, const uint8_t* data, size_t data_size) const {
   DCHECK(audio_object_type_ >= 1 && audio_object_type_ <= 4 &&
          frequency_index_ != 0xf && channel_config_ <= 7);
+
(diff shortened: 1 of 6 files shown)


packager/media/codecs/aac_audio_specific_config.cc
```diff
@@ -162,6 +162,33 @@ bool AACAudioSpecificConfig::ConvertToADTS(std::vector<uint8_t>* buffer) const {
   return true;
 }
 
+std::vector<uint8_t> AACAudioSpecificConfig::GetAACwithADTSHeader(const uint8_t* data, size_t length) const {
+  DCHECK(audio_object_type_ >= 1 && audio_object_type_ <= 4 &&
+         frequency_index_ != 0xf && channel_config_ <= 7);
+
+  std::vector<uint8_t> aac_with_adts_header(data, data + length);
```
kqyang: There are still two copies (line 169 and line 179), which is identical to the original code. 

You can remove the extra copying by appending the data later, e.g.

std::vector<uint8_t> aac_with_adts_header;
aac_with_adts_header.reserve(length +  kADTSHeaderSize);
...
aac_with_adts_header.insert(aac_with_adts_header.end(), kADTSHeaderSize, 0);
...
aac_with_adts_header.insert(aac_with_adts_header.end(), data, data+length);

packager/media/codecs/aac_audio_specific_config.cc
```diff
@@ -162,6 +162,33 @@ bool AACAudioSpecificConfig::ConvertToADTS(std::vector<uint8_t>* buffer) const {
   return true;
 }
 
+std::vector<uint8_t> AACAudioSpecificConfig::GetAACwithADTSHeader(const uint8_t* data, size_t length) const {
```
kqyang: Can you update ConvertToADTS function directly?

You may change the function signature to:

bool ConvertToADTS(const uint8_t* data, size_t data_size, std::vector<uint8_t>* output_buffer)

PR #638, opened 2019-08-19
Update general_encryption_options with a better clear_lead description

herlon214:
Added a better description to `clear_lead` docs with the same content as defined in the change log (https://github.com/google/shaka-packager/commit/9e9833ea63cf75c2bcd3d1ee8202c92450e5b461). I'm learning about the DRM and it took me hours to figure out why my segments were not being encrypted (similar to https://github.com/google/shaka-player/issues/1883).

### DIFF
--- a/docs/source/options/general_encryption_options.rst
+++ b/docs/source/options/general_encryption_options.rst
@@ -13,6 +13,12 @@ General encryption options
 --clear_lead <seconds>
 
     Clear lead in seconds if encryption is enabled.
+    Shaka Packager does not support partial encrypted segments, all the 
+    segments including the partial segment overlapping with the initial 
+    'clear_lead' seconds are not encrypted, with all the following segments 
+    encrypted. If segment_duration is greater than 'clear_lead', then only the
+    first segment is not encrypted.
+    Default: 5
 
 --protection_systems
 


herlon214:
@googlebot I signed it!

docs/source/options/general_encryption_options.rst
```diff
@@ -13,6 +13,9 @@ General encryption options
 --clear_lead <seconds>
 
     Clear lead in seconds if encryption is enabled.
+    Shaka Packager does not support partial encrypted segments, so if segment_duration
```
kqyang: Can you break the new line after so if?

docs/source/options/general_encryption_options.rst
```diff
@@ -13,6 +13,9 @@ General encryption options
...
+    is 6 seconds, then only the first segment is in clear (without encryption), with all the
```
kqyang: This is only true if the value of clear_lead is smaller than segment_duration. Can you change it to something like:

so if segment_duration is greater than clear_lead, then only the first segment is in clear (without encryption) with all the following segments encrypted.

herlon214:
Could you check if it's better now @kqyang ?

docs/source/options/general_encryption_options.rst
```diff
@@ -13,6 +13,12 @@ General encryption options
 --clear_lead <seconds>
 
     Clear lead in seconds if encryption is enabled.
+    Shaka Packager does not support partial encrypted segments, so if 
+    segment_duration is greater than clear_lead, then only the 
+    first segment is in clear (without encryption), with all 
+    the following segments encrypted. In the opposite way, if your 
+    segment is lower than clear_lead it will not be encrypted.
```
kqyang: The last sentence is a bit difficult to understand. How about changing the whole paragraph to something like this:

Shaka Packager does not support partial encrypted segments, all the segments
including the partial segment overlapping with the initial `clear_lead` seconds are
not encrypted, with all the following segments encrypted. For example, if
`segment_duration` is greater than `clear_lead`, then the first segment is not
encrypted.
then only the first segment is

docs/source/options/general_encryption_options.rst
```diff
@@ -12,7 +12,12 @@ General encryption options
 
 --clear_lead <seconds>
 
-    Clear lead in seconds if encryption is enabled.
+    Shaka Packager does not support partial encrypted segments, all the 
+    segments including the partial segment overlapping with the initial 
+    'clear_lead' seconds are not encrypted, with all the following segments 
+    encrypted. For example, if segment_duration is greater than 'clear_lead', 
+    the first segment is not encrypted, then only the first segment is
```
kqyang: Sorry for typo in my previous comments (was distracted).

I meant to keep the sentence "Clear lead in seconds if encryption is enabled." It should be something like:

    Clear lead in seconds if encryption is enabled.
    Shaka Packager does not support partial encrypted segments, all the 
    segments including the partial segment overlapping with the initial 
    'clear_lead' seconds are not encrypted, with all the following segments 
    encrypted. If segment_duration is greater than 'clear_lead', then only the
    first segment is not encrypted.
    Default: 5

herlon214: No problem, I pushed your text now.

kqyang:
@herlon214 Thanks!

Issue #598, opened 2019-06-18
Support 5.1 (side) AAC encoded by FFmpeg (actual channel layout is stored in PCE)


joeyparrish:
## System info
**Operating System**: gLinux
**Shaka Packager Version**: v2.3.0

## Issue and steps to reproduce the problem

**Packager Command**:

```sh
# Extract 30 seconds of surround sound from Sintel into MPEG2-TS (attached below)
ffmpeg -i Sintel.2010.4k.mkv -f mpegts -map 0:a -c:a aac -b:a 192k -t 30 surround.ts
# Package it to output.mp4
packager input=surround.ts,stream=audio,output=output.mp4 --mpd_output output.mpd
```
[surround.zip](https://github.com/google/shaka-packager/files/3303586/surround.zip)

What is the expected result?
I would expect to see the output file generated.

What happens instead?
Instead, I get lots of this:

```
[0618/145709:ERROR:adts_header.cc(76)] Not implemented reached in
  virtual bool shaka::media::mp2t::AdtsHeader::Parse(const uint8_t *, size_t)
  ADTS frames with more than one data block not supported.
```

Followed by this:
```
[0618/145709:ERROR:packager_main.cc(523)] Packaging Error: 6 (END_OF_STREAM): 
```

And no output.mp4 is generated.

<Please provide any additional information below.>

If Instead, I convert the 5.1 surround into stereo, everything works:

```sh
# Use -ac 2 this time to convert down to stereo (attached below)
ffmpeg -i Sintel.2010.4k.mkv -f mpegts -map 0:a -c:a aac -b:a 192k -t 30 -ac 2 stereo.ts
packager input=stereo.ts,stream=audio,output=output.mp4 --mpd_output output.mpd
```
[stereo.zip](https://github.com/google/shaka-packager/files/3303592/stereo.zip)

kqyang:
@joeyparrish Thanks for filing the issue. We do not support more than one data block in an ADTS frames right now. 

It is a nice feature to support. Move to backlog for now until we have resources to work on it.

PhilBax:
Thought I'd take a quick look at this issue... 

@joeyparrish:

If I download your .ts file, I can reproduce the problem. Interestingly, if I open your file in VLC, I hear some odd bit of music that is very stuttery (about 0.25 seconds of silence, 0.25 seconds of music, repeated through all 30 seconds of the clip). VLC claims the audio is ADTS stereo... so I'm wondering if I'm missing a codec or something.

[...]

PR #656, opened 2019-09-26
Working toward fixing #598

PhilBax:


PhilBax:
@googlebot I signed it!

packager/media/formats/mp2t/adts_header.cc
```diff
@@ -63,7 +63,7 @@ bool AdtsHeader::Parse(const uint8_t* adts_frame, size_t adts_frame_size) {
   // Skip private stream bit.
   RCHECK(frame.SkipBits(1));
   RCHECK(frame.ReadBits(3, &channel_configuration_));
-  RCHECK((channel_configuration_ > 0) &&
+  RCHECK((channel_configuration_ >= 0) &&
```
kqyang: This condition can be removed as |channel_configuration_| is unsigned.

packager/media/formats/mp2t/adts_header.cc
```diff
@@ -109,7 +117,7 @@ uint32_t AdtsHeader::GetSamplingFrequency() const {
 }
 
 uint8_t AdtsHeader::GetNumChannels() const {
-  DCHECK_GT(channel_configuration_, 0);
+  DCHECK_GE(channel_configuration_, 0);
```
kqyang: The line can be removed as |channel_configuration_| is unsigned.

packager/media/formats/mp2t/audio_header.h
```diff
@@ -50,6 +50,8 @@ class AudioHeader {
   /// @return the size of frame (header + payload).
   virtual size_t GetFrameSize() const = 0;
 
+  virtual size_t GetFrameSizeWithoutParsing(const uint8_t* data, size_t num_bytes) const {return 0;}
```
kqyang: Can you add a comment like the functions above? Also define it as pure virtual.

You'll also need to implement the similar function in ac3_header.h.

packager/media/formats/mp2t/es_parser_audio.cc
```diff
@@ -56,20 +56,39 @@ static bool LookForSyncWord(const uint8_t* raw_es,
...
+
+     if (!audio_header->IsSyncWord(cur_buf))
       continue;
 
     // Check whether there is another frame |size| apart from the current one.
     const size_t remaining_size = static_cast<size_t>(raw_es_size - offset);
     const int kSyncWordSize = 2;
-    if (remaining_size >= audio_header->GetFrameSize() + kSyncWordSize &&
-        !audio_header->IsSyncWord(&cur_buf[audio_header->GetFrameSize()])) {
+    const size_t frame_size = audio_header->GetFrameSizeWithoutParsing(cur_buf, audio_header->GetMinFrameSize()); /* Get Frame size without parsing the frame */
```
kqyang: audio_header->GetMinFrameSize() should be replaced with remaining_size.

packager/media/formats/mp2t/es_parser_audio.cc
```diff
@@ -56,20 +56,39 @@ static bool LookForSyncWord(const uint8_t* raw_es,
...
+    //   continue;
 
-    if (!audio_header->Parse(cur_buf, raw_es_size - offset))
+    // // Check whether there is another frame |size| apart from the current one.
+    // const size_t remaining_size = static_cast<size_t>(raw_es_size - offset);
+    // const int kSyncWordSize = 2;
+    // if (remaining_size >= audio_header->GetFrameSize() + kSyncWordSize &&
+    //     !audio_header->IsSyncWord(&cur_buf[audio_header->GetFrameSize()])) {
+    //   continue;
+    // }
```
kqyang: Remove these commented codes.

packager/media/formats/mp2t/adts_header.cc
PhilBax: Oh! So it is!

packager/media/formats/mp2t/audio_header.h
PhilBax: Yeah, this was just a quick-and-dirty implementation to get it to build. I haven't looked at the ac3 side of things. I might ping you if it's not immediately obvious to me how to implement this function for AC3 audio.

PhilBax:
Should I squash the pull request down to a single commit? Do you do squash-merges? Do you prefer full-history?

packager/media/formats/mp2t/audio_header.h
PhilBax: Ah! Okay, I get was the adts version was doing. Should be able to handle the ac3 version, I think.

PhilBax: Since GetFrameSize() actually has to do some logic, would it be bad form to have GetFrameSizeWithoutParsing() actually set fscod_ and frmsizecod_ and then call GetFrameSize() (though, I guess that would make the function non-const)?

Would it be best to have both GetFrameSizes call a common function to do the actual frame size logic?

Or is the AC3 standard fixed enough that duplicating the small amount of logic involved is fine by y'all?

PhilBax: Like, does this work for you?

```
/// Calculate the size of the frame from the sample rate code and the 
/// frame size code.
/// @return the size of the frame (header + payload).
size_t CalcFrameSize(uint8_t fscod, uint8_t frmsizecod) {
  const size_t kNumFscode = arraysize(kAc3SampleRateTable);
  DCHECK_LT(fscod, kNumFscode);
  DCHECK_LT(frmsizecod, arraysize(kFrameSizeCodeTable));
  // The order of frequencies are reversed in |kFrameSizeCodeTable| compared to
  // |kAc3SampleRateTable|.
  const int index = kNumFscode - 1 - fscod;
  return kFrameSizeCodeTable[frmsizecod][index] * 2;
}

size_t Ac3Header::GetFrameSize() const {
  return CalcFrameSize(fscod_, frmsizecod_);
}

size_t Ac3Header::GetFrameSizeWithoutParsing(const uint8_t* data, size_t num_bytes) const {
  uint8_t fscod = data[5] & 0xc0;
  uint8_t frmsizecod = data[5] & 0x3f;
  return CalcFrameSize(fscod, frmsizecod);
}
```

PhilBax:
> Can you run "git clang-format " to format the code properly?

TIL! :smile:

packager/media/formats/mp2t/audio_header.h
kqyang: Yes, that looks very good to me!

> uint8_t fscod = data[5] & 0xc0;

Should be data[4] instead of data[5].

Also, there should be check of 'num_bytes' to make sure that it is >4.

kqyang:
> Should I squash the pull request down to a single commit? Do you do squash-merges? Do you prefer full-history?

Yes, I will do "squash and merge", so you don't need to squash the pull request into a single commit.

Btw, if you don't have 'clang-format' installed already, you can follow this instruction to install it: https://github.com/google/shaka-packager/blob/master/packager/tools/git/check_formatting.py#L16.

Thanks again for your help!

packager/media/formats/mp2t/audio_header.h
PhilBax: Good catch! I hadn't actually tested it yet. 'preciate it!

PhilBax:
I did exactly that! I'd seen clang-format before, but didn't know about the git integration. Really neat tool!

PhilBax:
For the unit tests... do you want *new* unit tests? Or should I just add an EXPECT_EQ to the existing parse tests, but before the parse actually happens? Technically the new function is "parsing" a small amount of info out of the header, and it would be nice to just go ahead and use the setup for the existing three ac3 tests. But I fully understand making it separate as well.

If new tests, should I add one to match each of the ac3 "parse" tests? Or just a single one for ac3 and one for adts?

kqyang:
> should I just add an EXPECT_EQ to the existing parse tests, but before the parse actually happens?

SGTM!

PhilBax:
K, should be good-to-go I think. Fixed a logic issue with the AC3 parsing in the code I pasted in the above comment, and added some unit tests which are now passing!

packager/media/formats/mp2t/ac3_header.cc
```diff
@@ -103,14 +103,28 @@ size_t Ac3Header::GetHeaderSize() const {
...
-  DCHECK_LT(frmsizecod_, arraysize(kFrameSizeCodeTable));
+  DCHECK_LT(fscod, kNumFscode);
+  DCHECK_LT(frmsizecod, arraysize(kFrameSizeCodeTable));
   // The order of frequencies are reversed in |kFrameSizeCodeTable| compared to
   // |kAc3SampleRateTable|.
-  const int index = kNumFscode - 1 - fscod_;
-  return kFrameSizeCodeTable[frmsizecod_][index] * 2;
+  const int index = kNumFscode - 1 - fscod;
+  return kFrameSizeCodeTable[frmsizecod][index] * 2;
+}
```
kqyang: Can you move it to the anonymous namespace above, i.e. after line 42?

packager/media/formats/mp2t/ac3_header.cc
```diff
@@ -103,14 +103,28 @@ size_t Ac3Header::GetHeaderSize() const {
   return 0;
 }
 
-size_t Ac3Header::GetFrameSize() const {
+/// Calculate the size of the frame from the sample rate code and the 
+/// frame size code.
+/// @return the size of the frame (header + payload).
```
kqyang: nit: Use normal comment style for internal functions, e.g.

// Calculate the size of the frame from the sample rate code and the 
// frame size code.

packager/media/formats/mp2t/ac3_header.cc
```diff
@@ -103,14 +103,28 @@ size_t Ac3Header::GetHeaderSize() const {
...
-  return kFrameSizeCodeTable[frmsizecod_][index] * 2;
+  const int index = kNumFscode - 1 - fscod;
+  return kFrameSizeCodeTable[frmsizecod][index] * 2;
+}
+
+size_t Ac3Header::GetFrameSize() const {
+  return CalcFrameSize(fscod_, frmsizecod_);
+}
+
+size_t Ac3Header::GetFrameSizeWithoutParsing(const uint8_t* data, size_t num_bytes) const {
```
kqyang: This is longer than 80 characters. Have you run "git clang-format"?

packager/media/formats/mp2t/es_parser_audio.cc
```diff
@@ -56,20 +56,26 @@ static bool LookForSyncWord(const uint8_t* raw_es,
...
     const uint8_t* cur_buf = &raw_es[offset];
 
-    if (!audio_header->IsSyncWord(cur_buf))
-      continue;
-
-    if (!audio_header->Parse(cur_buf, raw_es_size - offset))
+     if (!audio_header->IsSyncWord(cur_buf))
       continue;
 
     // Check whether there is another frame |size| apart from the current one.
```
kqyang: Can you move this comment before line 71?

packager/media/formats/mp2t/es_parser_audio.cc
```diff
@@ -56,20 +56,26 @@ static bool LookForSyncWord(const uint8_t* raw_es,
...
     const size_t remaining_size = static_cast<size_t>(raw_es_size - offset);
```
kqyang: Thinking about it again, I think there should be another condition here:

```c
if (remaining_size < audio_header->GetMinFrameSize()) {
  // Not a full frame: will resume when we have more data.
  return false;
}
```

packager/media/formats/mp2t/es_parser_audio.cc
```diff
@@ -56,20 +56,26 @@ static bool LookForSyncWord(const uint8_t* raw_es,
...
     const int kSyncWordSize = 2;
-    if (remaining_size >= audio_header->GetFrameSize() + kSyncWordSize &&
-        !audio_header->IsSyncWord(&cur_buf[audio_header->GetFrameSize()])) {
+    const size_t frame_size =
+        audio_header->GetFrameSizeWithoutParsing(cur_buf, remaining_size);
+    if (frame_size < audio_header->GetMinFrameSize())
+      continue;
```
kqyang: Can you add a comment here:

```c
if (frame_size < audio_header->GetMinFrameSize()) {
  // Too short to be a valid frame.
  continue;
}
```

packager/media/formats/mp2t/es_parser_audio.cc
```diff
@@ -56,20 +56,26 @@ static bool LookForSyncWord(const uint8_t* raw_es,
...
+    if (remaining_size < frame_size)
+      return false;
```
kqyang: Similarly,

```c
    if (remaining_size < frame_size) {
      // Not a full frame: will resume when we have more data.
      return false;
    }
```

packager/media/formats/mp2t/ac3_header.cc
PhilBax: Sure can!

packager/media/formats/mp2t/ac3_header.cc
PhilBax: Ah! I wondered. Should've checked the style guide.

packager/media/formats/mp2t/ac3_header.cc
PhilBax: I did, but I committed this before running it. Apparently git clang-format only runs on the lines you're actively submitting? I presume there must be a flag to run on the whole file. Will look into it!

packager/media/formats/mp2t/es_parser_audio.cc
PhilBax: Sure thing!

packager/media/formats/mp2t/ac3_header.cc
kqyang: You can run `git clang-format <base_commit>`

For example, if `HEAD^` is the base commit, you can run `git clang-format HEAD^`

PhilBax: Gotcha! Done!

packager/media/formats/mp2t/es_parser_audio.cc
```diff
@@ -59,17 +59,29 @@ static bool LookForSyncWord(const uint8_t* raw_es,
...
+        audio_header->GetFrameSizeWithoutParsing(cur_buf, remaining_size);
+    if (frame_size < audio_header->GetMinFrameSize())
+      // Too short to be a valid frame.
+      continue;
+    if (remaining_size < frame_size)
+      // Not a full frame: will resume when we have more data.
+      return false;
+    if (remaining_size < audio_header->GetMinFrameSize()) {
+      // Not a full frame: will resume when we have more data.
+      return false;
```
kqyang: Just realized that this will never occur as 

remaining_size = raw_es_size - offset >= raw_es_size - (raw_es_size - audio_header->GetMinFrameSize()) = audio_header->GetMinFrameSize(). 

Can you remove the three lines from 72 - 74?

Sorry about the previous suggestion earlier.

packager/media/formats/mp2t/adts_header.cc
```diff
@@ -89,6 +88,14 @@ size_t AdtsHeader::GetFrameSize() const {
   return frame_size_;
 }
 
+size_t AdtsHeader::GetFrameSizeWithoutParsing(const uint8_t* data,
+                                              size_t num_bytes) const {
+  if (num_bytes < 6)
+    return 0;
```
kqyang: For consistency with Ac3Header, replace it with DCHECK_GT(num_bytes, 5)?

packager/media/formats/mp2t/ac3_header.cc
```diff
@@ -104,13 +117,15 @@ size_t Ac3Header::GetHeaderSize() const {
...
-  // |kAc3SampleRateTable|.
-  const int index = kNumFscode - 1 - fscod_;
-  return kFrameSizeCodeTable[frmsizecod_][index] * 2;
+  return CalcFrameSize(fscod_, frmsizecod_);
+}
+
+size_t Ac3Header::GetFrameSizeWithoutParsing(const uint8_t* data,
+                                             size_t num_bytes) const {
+  DCHECK_GT(num_bytes, static_cast<size_t>(4));
+  uint8_t fscod = (data[4] & 0xc0) >> 6;
```
kqyang: nit: can be replaced with

uint8_t fscod = data[4] >> 6;

PhilBax: Ah! Fair enough!

packager/media/formats/mp2t/adts_header.cc
```diff
@@ -89,6 +88,13 @@ size_t AdtsHeader::GetFrameSize() const {
   return frame_size_;
 }
 
+size_t AdtsHeader::GetFrameSizeWithoutParsing(const uint8_t* data,
+                                              size_t num_bytes) const {
+  DCHECK_GT(num_bytes, 5)
```
kqyang: Missing ";" in the end. 

It causing building failures: https://ci.appveyor.com/project/shaka/shaka-packager/builds/27774035

PhilBax: Oi! My lack of sleep is showing. :smile:

PhilBax: Also, I have separate folders for the "official" shaka build, and my fork (which lacks the build-related files). I should've taken the time to overlay my fork on top of the folder with the build files.

packager/media/formats/mp2t/adts_header.cc
```diff
@@ -89,6 +88,13 @@ size_t AdtsHeader::GetFrameSize() const {
   return frame_size_;
 }
 
+size_t AdtsHeader::GetFrameSizeWithoutParsing(const uint8_t* data,
+                                              size_t num_bytes) const {
+  DCHECK_GT(num_bytes, 5);
```
kqyang: Ah, there are still failures on Windows: https://travis-ci.org/google/shaka-packager/builds/591706925?utm_source=github_status&utm_medium=notification

You can probably fix it by either

```c++
DCHECK_GT(num_bytes, 5u);
```

or 

```C++
DCHECK_GT(num_bytes, static_cast<size_t>(5));
```

kqyang:
@PhilBax The CL is now merged. Thanks for improving Shaka Packager!

PhilBax:
Thank you for all the help!!

PR #683, opened 2019-11-26
Fix problem with default pssh_data

lasotaartur:
While creating pssh object if the argument pssh_data is passed as the empty string then default value '' is assigned to pssh_data. Later this value is merging with binary values and error occurs.

### DIFF
--- a/packager/tools/pssh/pssh-box.py
+++ b/packager/tools/pssh/pssh-box.py
@@ -78,7 +78,7 @@ class Pssh(object):
     self.version = version
     self.system_id = system_id
     self.key_ids = key_ids or []
-    self.pssh_data = pssh_data or ''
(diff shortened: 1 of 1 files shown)


lasotaartur:
> @googlebot I signed it!

@googlebot I signed it!

kqyang:
@lasotaartur Thanks for working on it! The change looks good to me.

Can you share a little bit more on where the problem was before your fix? How can we re-produce the problem? It will be great if you can file an issue so the fix is better tracked.

FYI, I tried with empty pssh-data in python2: `python2 out/Debug/pssh-box.py --common`, which works without problem. I tried the same command with python3, which fails even with your fix. So if your goal is to fix the tool with python3, it needs some updates.

Thanks.

Issue #701, opened 2020-01-22
[TS] Segment template $Time$ is truncated to 32 bits


Canta:
## System info
**Operating System**: CentOS Linux release 7.7.1908 (Core)
**Shaka Packager Version**: c7312176078940ba44b4dc5c8967eb0f3d691ea4

## Issue and steps to reproduce the problem

**Quick description**:
When packager runs for a long enough time, `$$Time$$` value resets and goes back from zero.

**Packager Command**:
```
shaka-packager 
'in=input, stream_selector=0,segment_template=/path/240_$$Time$$.ts,playlist_name=240.m3u8,drm_label=SD,bandwidth=512000' 
'in=input, stream_selector=1,segment_template=/path/360_$$Time$$.ts,playlist_name=360.m3u8,drm_label=SD,bandwidth=768000' 
'in=input, stream_selector=2,segment_template=/path/480_$$Time$$.ts,playlist_name=480.m3u8,drm_label=SD,bandwidth=1536000' 
'in=input, stream_selector=3,segment_template=/path/720_$$Time$$.ts,playlist_name=720.m3u8,drm_label=HD,bandwidth=3072000' 
'in=input, stream_selector=4,segment_template=/path/audio_$$Time$$.ts,playlist_name=audio.m3u8,drm_label=AUDIO,bandwidth=64000,language=eng' 
--hls_master_playlist_output /path/master.m3u8 
--hls_playlist_type LIVE 
--segment_duration 3.2 
--minimum_update_period 3.2
--time_shift_buffer_depth 7200 
--preserved_segments_outside_live_window 100 
--dump_stream_info 
```

**What is the expected result?**:
To increase linearly the value of `$$Time$$`

**What happens instead?**:
`$$Time$$` resets its value.

**Additional information**:

Please take a look at this HLS medialist extract:
```
[user@server ~]# cat /path/audio.m3u8
(...)
#EXTINF:3.200,
audio_4294641887.ts
#EXTINF:3.200,
audio_4294929887.ts
#EXTINF:3.200,
audio_250591.ts
#EXTINF:3.200,
audio_538591.ts
#EXTINF:3.200,
audio_826591.ts
(...)
```

There, you can see a sudden time jump, from `4294929887` to `250591`.
I believe what's happening is that this `$$Time$$` value increased over the `uint32` limit of `4294967295` (`0xffffffff`).

[...]

PR #706, opened 2020-01-31
Change $Time$ template value for TS segments from uint32 to int64

Canta:
See #701 for details. This PR should fix it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,12 @@ before_deploy:
   - rm -rf deploy
   - mkdir deploy
   - |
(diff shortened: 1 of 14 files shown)


Canta:
I don't understand why this is including previously merged files. I believe it's some github thing. So I created it as a "draft" hoping to have some options to deal with that situation. I don't see any, and also can't delete the PR. 

So, I guess if I rebase my fork to a newer commit, things will get fixed here. Will try that.

Canta:
Closing this, as will try again with the rebased fork.

PR #713, opened 2020-02-17
Support signalling of last segment number. 

sr1990:
Hi, 
I have added a feature as per 4.4.3.6. Signalling the last segment number in Period: 
To indicate last segment number, add a Supplemental Descriptor with @schemeIdUri set to http://dashif.org/guidelines/last-segment-number with the @value set to the last segment number.

Example command :
```
packager \
  'in=h264_baseline_360p_600.mp4,stream=audio,init_segment=audio/init.mp4,segment_template=audio/$Number$.m4s' \
  'in=input_text.vtt,stream=text,init_segment=text/init.mp4,segment_template=text/$Number$.m4s' \
  'in=h264_baseline_360p_600.mp4,stream=video,init_segment=h264_360p/init.mp4,segment_template=h264_360p/$Number$.m4s' \
  'in=h264_main_480p_1000.mp4,stream=video,init_segment=h264_480p/init.mp4,segment_template=h264_480p/$Number$.m4s' \
  'in=h264_main_720p_3000.mp4,stream=video,init_segment=h264_720p/init.mp4,segment_template=h264_720p/$Number$.m4s' \
  'in=h264_high_1080p_6000.mp4,stream=video,init_segment=h264_1080p/init.mp4,segment_template=h264_1080p/$Number$.m4s' \
  --generate_static_live_mpd --mpd_output h264.mpd --last_segment_number "video:400,audio:100" 
```
Example mpd: http://dash.akamaized.net/dash264/TestCasesIOP41/LastSegmentNumber/1/manifest_last_segment_num.mpd

Implementation: I have added the following argument to indicate the last segment number: --last_segment_number "video:400,audio:100"
Syntax is comma separated  `<adaptationset content type>:<last segment number>`.

Source: https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf

### DIFF
--- a/packager/mpd/base/xml/xml_node.cc
+++ b/packager/mpd/base/xml/xml_node.cc
@@ -436,10 +436,9 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
       segment_template.SetIntegerAttribute("duration",
                                            segment_infos.front().duration);
       if (FLAGS_dash_add_last_segment_number_when_needed) {
-        uint32_t last_segment_number = 0;
-        for (const auto& segment_info_element: segment_infos) {
-	  last_segment_number = last_segment_number + 
-		                segment_info_element.repeat + 1;
+        uint32_t last_segment_number = start_number - 1;
+        for (const auto& segment_info_element : segment_infos) {
+          last_segment_number += segment_info_element.repeat + 1;
 	}
         AddSupplementalProperty(
           "http://dashif.org/guidelines/last-segment-number",
--- a/packager/mpd/base/xml/xml_node.cc
+++ b/packager/mpd/base/xml/xml_node.cc
@@ -24,6 +24,12 @@ DEFINE_bool(segment_template_constant_duration,
             "Generates SegmentTemplate@duration if all segments except the "
             "last one has the same duration if this flag is set to true.");
 
+DEFINE_bool(dash_add_last_segment_number_when_needed,
+            false,
+            "Adds a Supplemental Descriptor with @schemeIdUri "
+            "set to http://dashif.org/guidelines/last-segment-number with "
+            "the @value set to the last segment number.");
+
 namespace shaka {
 
 using xml::XmlNode;
@@ -429,6 +435,16 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
     if (IsTimelineConstantDuration(segment_infos, start_number)) {
       segment_template.SetIntegerAttribute("duration",
                                            segment_infos.front().duration);
+      if (FLAGS_dash_add_last_segment_number_when_needed) {
+        uint32_t last_segment_number = 0;
+        for (const auto& segment_info_element: segment_infos) {
+	  last_segment_number = last_segment_number + 
+		                segment_info_element.repeat + 1;
+	}
+        AddSupplementalProperty(
+          "http://dashif.org/guidelines/last-segment-number",
+          std::to_string(last_segment_number));	
+      }
     } else {
       XmlNode segment_timeline("SegmentTimeline");
       if (!PopulateSegmentTimeline(segment_infos, &segment_timeline) ||
--- a/packager/mpd/base/xml/xml_node.cc
+++ b/packager/mpd/base/xml/xml_node.cc
@@ -437,9 +437,9 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
                                            segment_infos.front().duration);
       if (FLAGS_dash_add_last_segment_number_when_needed) {
         uint32_t last_segment_number = start_number - 1;
-        for (const auto& segment_info_element : segment_infos) {
+        for (const auto& segment_info_element : segment_infos) 
           last_segment_number += segment_info_element.repeat + 1;
-	}
+	
         AddSupplementalProperty(
           "http://dashif.org/guidelines/last-segment-number",
           std::to_string(last_segment_number));	
--- a/packager/app/mpd_flags.cc
+++ b/packager/app/mpd_flags.cc
@@ -62,10 +62,3 @@ DEFINE_bool(
     "completely."
     "Ignored if $Time$ is used in segment template, since $Time$ requires "
     "accurate Segment Timeline.");
-DEFINE_string(last_segment_number,
-             "",
-             "Indicate the last segment number in adaptation-set with "
-             "comma separated <AdaptationSet content type>:<last segment number> "
-	     "values. This will add a Supplemental Descriptor with @schemeIdUri"
-	     "set to http://dashif.org/guidelines/last-segment-number with"
-	     "the @value set to the last segment number.");
--- a/packager/app/mpd_flags.cc
+++ b/packager/app/mpd_flags.cc
@@ -62,3 +62,10 @@ DEFINE_bool(
     "completely."
     "Ignored if $Time$ is used in segment template, since $Time$ requires "
     "accurate Segment Timeline.");
+DEFINE_string(last_segment_number,
+             "",
+             "Indicate the last segment number in adaptation-set with "
+             "comma separated <AdaptationSet content type>:<last segment number> "
+	     "values. This will add a Supplemental Descriptor with @schemeIdUri"
+	     "set to http://dashif.org/guidelines/last-segment-number with"
+	     "the @value set to the last segment number.");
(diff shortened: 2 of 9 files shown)


sr1990:
Hi @kqyang , 
Thanks for taking a look. Following is what I think:
1. Per 4.4.3.6 Signalling the last segment number in Period, there are three ways to signal the last segment number:

a. Use the lmsg signalling as defined in clause 4.4.3.5.
b. Use the Segment Timeline with @r value greater or equal to 0.
c. Add a Supplemental Descriptor with @schemeIdUri set to http://dashif.org/guidelines/last-segment-number with the @value set to the last segment number.

This PR works on (c), which is only needed if (b) is not used, i.e. it is only needed in this branch: https://github.com/google/shaka-packager/blob/master/packager/mpd/base/xml/xml_node.cc#L430.

-> Yes (c) can be used to indicate the last segment number if (b) is not present. If both are not present, it can be calculated from period duration. For example, on client side, Exoplayer will use option (c) to calculate the last segment number if option (b) is not present. 

If you look at getSegmentCount() method in the following pull request  https://github.com/google/ExoPlayer/pull/5908/files  
or the following file: https://github.com/google/ExoPlayer/blob/release-v2/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/SegmentBase.java ,
Segment timeline takes a priority over the last segment number that is provided by the supplemental property, i.e., option (b) has higher priority over (c)
and while playing the following mpd example:  http://dash.akamaized.net/dash264/TestCasesIOP41/LastSegmentNumber/1/manifest_last_segment_num.mpd 
Exoplayer calculates the last segment using Period duration for video and audio adaptation set which is 371 if (b) and (c) are not present and in this case player stops playing the video at segment 368 as it is indicated via Supplemental property. 

Also this supplemental property is mentioned at Adaptation set level. Do you think there is a need to add changes at representation level at https://github.com/google/shaka-packager/blob/master/packager/mpd/base/xml/xml_node.cc#L430 ? (https://github.com/Dash-Industry-Forum/DASH-IF-IOP/issues/340)

2. Is there any reason to provide last segment number externally via a command line flag? Shaka Packager can calculate the last segment number from the start number and the number of entries in segment_infos: https://github.com/google/shaka-packager/blob/master/packager/mpd/base/xml/xml_node.cc#L400. So we only need a boolean flag to indicate whether to generate the new Supplemental Property. It can be named as something like: --dash_add_last_segment_number_when_needed.

This option is something that can be internally calculated as you have mentioned and added as a supplemental property but as seen above, Exoplayer can calculate the the last segment number using period if option (b) or (c) are not present. Having this flag as an option gives the content provides a way to set the last segment number if they do not want to go with option (b) and rely on player's calculation of last segment number using the period duration. I think we should change the flag to --dash_last_segment_number instead of the one that I have mentioned. 

kqyang:
> Having this flag as an option gives the content provides a way to set the last segment number if they do not want to go with option (b) and rely on player's calculation of last segment number using the period duration.

Can you describe the usage scenario, i.e. when users do not want to go with option (b) or relies on player's calculation of last segment number using the period duration?

sr1990:
> Can you describe the usage scenario, i.e. when users do not want to go with option (b) or relies on player's calculation of last segment number using the period duration?
@kqyang Thanks for the reply. Do you want to open a ticket with DASH-IF to check what use cases they might have and why they added the option and also to check if the above interpretation is correct? 
The use case that I can think of for 
static mpd: If there is a period containing multiple adaptation sets (one adaptation set containing only 4k and second one containing the rest of the resolution ladder) then the content provider can limit the duration of the video stream in 4k and force the player to play the lower resolution for subscription/payment reasons.
live mpd: explicitly stating the last segment number in one of the mpd update.  
Again, I might be wrong with the above use cases, so I think it's better that we clarify.

kqyang:
> Do you want to open a ticket with DASH-IF to check what use cases they might have and why they added the option and also to check if the above interpretation is correct?

No. I understand the general use case of segment number. 

What I don't understand is why you want to allow users to specify the segment number, which is not scalable, i.e. it does not make sense to me to expect users to know the exact last segment number and it is also not possible for live packaging.

> static mpd: If there is a period containing multiple adaptation sets (one adaptation set containing only 4k and second one containing the rest of the resolution ladder) then the content provider can limit the duration of the video stream in 4k and force the player to play the lower resolution for subscription/payment reasons.

For this use case, a better option is to just don't include the 4k stream during packaging.

> live mpd: explicitly stating the last segment number in one of the mpd update.

Yes, that is one of the main use case for last segment number. I don't think this PR can support this use case as it is not a constant.

On the other hand, it can be supported in packager with just a boolean flag, something like this:

```diff
diff --git i/packager/mpd/base/xml/xml_node.cc w/packager/mpd/base/xml/xml_node.cc
index 51435183..1c7776af 100644
--- i/packager/mpd/base/xml/xml_node.cc
+++ w/packager/mpd/base/xml/xml_node.cc
@@ -400,7 +400,8 @@ bool RepresentationXmlNode::AddVODOnlyInfo(const MediaInfo& media_info) {
 bool RepresentationXmlNode::AddLiveOnlyInfo(
     const MediaInfo& media_info,
     const std::list<SegmentInfo>& segment_infos,
-    uint32_t start_number) {
+    uint32_t start_number,
+    bool add_last_segment_when_needed) {
   XmlNode segment_template("SegmentTemplate");
   if (media_info.has_reference_time_scale()) {
     segment_template.SetIntegerAttribute("timescale",
@@ -429,6 +430,12 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
     if (IsTimelineConstantDuration(segment_infos, start_number)) {
       segment_template.SetIntegerAttribute("duration",
                                            segment_infos.front().duration);
+      if (add_last_segment_when_needed) {
+        const uint32_t last_segment_number = start_number + segment_infos.size() - 1;
+        AddSupplementalProperty(
+            "http://dashif.org/guidelines/last-segment-number",
+            std::to_string(last_segment_number));
+      }
     } else {
       XmlNode segment_timeline("SegmentTimeline");
       if (!PopulateSegmentTimeline(segment_infos, &segment_timeline) ||
```

sr1990:
Ok yes! I agree. My mistake.That is what I was trying to do. So if the flag is enabled SegmentTimeline would be skipped and the supplemental property would be added. Wasn't thinking through packaging point of view. Won't this add the property at Representation level?

kqyang:
Great. We are on the same page then. Would you be willing to update your PR with the new approach?

> Won't this add the supplemental property at Representation level?

Correct. It will be added to Representation. My understanding is that it can be either in AdaptationSet or Representation.

If needed, we can examine the contents of the Representations and move common fields to AdaptationSet level.

sr1990:
Yes I will work on it

packager/mpd/base/xml/xml_node.cc
```diff
@@ -429,6 +435,16 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
     if (IsTimelineConstantDuration(segment_infos, start_number)) {
       segment_template.SetIntegerAttribute("duration",
                                            segment_infos.front().duration);
+      if (FLAGS_dash_add_last_segment_number_when_needed) {
+        uint32_t last_segment_number = 0;
```
kqyang: last_segment_number should start with `start_number`.

packager/mpd/base/xml/xml_node.cc
```diff
@@ -429,6 +435,16 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
...
+        for (const auto& segment_info_element: segment_infos) {
```
kqyang: There should be a space between "segment_info_element" and ":". You may also use clang-format: https://github.com/google/shaka-packager/blob/master/packager/tools/git/check_formatting.py.

packager/mpd/base/xml/xml_node.cc
```diff
@@ -429,6 +435,16 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
...
+	  last_segment_number = last_segment_number +
```
kqyang: Two more spaces indent.

packager/mpd/base/xml/xml_node.cc
```diff
@@ -429,6 +435,16 @@ bool RepresentationXmlNode::AddLiveOnlyInfo(
     if (IsTimelineConstantDuration(segment_infos, start_number)) {
       segment_template.SetIntegerAttribute("duration",
                                            segment_infos.front().duration);
+      if (FLAGS_dash_add_last_segment_number_when_needed) {
+        uint32_t last_segment_number = 0;
+        for (const auto& segment_info_element: segment_infos) {
+	  last_segment_number = last_segment_number + 
+		                segment_info_element.repeat + 1;
+	}
```
kqyang: Two more spaces indent.

kqyang: This is not done yet. You may also remove the brackets for the for loop in line 440.

Issue #542, opened 2019-01-17
[DASH] Allow AdaptationSet switching on FLAC and Opus (or make the codecs customize-able)


ctoshni:
Hi, i'm using latest Shaka version ;

As per DASH-IF-IOP-v4.3 - "A content author may signal such seamless switching property across Adaptation Sets by providing a Supplemental Descriptor along with an Adaptation Set with @schemeIdUri set to  `urn:mpeg:dash:adaptation-set-switching:2016` and the @value is a comma separated list of Adaptation Set IDs that may be seamlessly switched to from this Adaptation Set."

I have 2 adaptation sets with same content but different codecs and I want to have seamless adaptive bitrate switching between them.

Which shaka-packager command can I use to generate a mpd with @schemeIdUri set to  `urn:mpeg:dash:adaptation-set-switching:2016` in adaptation sets?

kqyang:
@ctoshni Shaka Packager generates the tag when seamless switching is possible, e.g. for SD and HD AdaptationSets, who are encrypted with different keys.

> I have 2 adaptation sets with same content but different codecs and I want to have seamless adaptive bitrate switching between them.

Normally for different codecs, seamless switching is not possible. May I know which codecs you'd like to allow seamless switching and which platform are you thinking of?

kqyang:
I am not sure what is the best way to support it. One option is to add an argument to allow users to specify the switchable codecs. What do you think?

kqyang:
@ctoshni We don't have bandwidth to work on it right now. I'll put it in v2.5 milestone for now.

Let us know if you'd like to help. We can give you pointers if you are interested.

ctoshni:
> Let us know if you'd like to help. We can give you pointers if you are interested.

I would be happy to help but will not be able to do due to bandwidth issues. 
As you have made it part of v2.5 milestone, closing the issue.

kqyang:
It needs to stay open to be included in future milestones :)

kqyang:
We have two options:

  1. Add a new flag to allow users to specify the set of switchable codecs, e.g. `--allow_switching="opus,flac"` if users want to allow switching between Opus and FLAC; it can also support multiple sets of switching: `--allow_switching="opus,flac;h264,h265"` to allow switching between Opus and FLAC, and also switching between h264 and h265.

  2. Add a boolean flag to allow switching between different codecs: `--allow_codec_switching`. 

[...]

Issue #177, opened 2016-12-07
(audio) track order in manifest.mpd


hariszukanovic:
It seems that the (audio) track order in output manifest is not controllable. The order in the manifest is used by many players to decide in which order to display the items in the audio selector UI control.

I tried specifying inputs in different order in the packager command line but I always get the same order in manifest.mpd

example:
shaka-packager
 'input=udp://238.100.100.2:6000,stream=video,stream_selector=0,init_segment=init-t11-video0.m4s,segment_template=chunk-t11-video0-$Number$.m4s,output_format=mp4'
 'input=udp://238.100.100.2:6000,stream=video,stream_selector=1,init_segment=init-t11-video1.m4s,segment_template=chunk-t11-video1-$Number$.m4s,output_format=mp4'
 'input=udp://238.100.100.2:6000,stream=video,stream_selector=2,init_segment=init-t11-video2.m4s,segment_template=chunk-t11-video2-$Number$.m4s,output_format=mp4'
 'input=udp://238.100.100.2:6000,stream=audio,language=mkd,stream_selector=4,init_segment=init-t11-audio-mkd.m4s,segment_template=chunk-t11-audio-mkd-$Number$.m4s,bandwidth=64000,output_format=mp4'
 'input=udp://238.100.100.2:6000,stream=audio,language=sqi,stream_selector=5,init_segment=init-t11-audio-sqi.m4s,segment_template=chunk-t11-audio-sqi-$Number$.m4s,bandwidth=64000,output_format=mp4'
 'input=udp://238.100.100.2:6000,stream=audio,language=srp,stream_selector=3,init_segment=init-t11-audio-srp.m4s,segment_template=chunk-t11-audio-srp-$Number$.m4s,bandwidth=64000,output_format=mp4'

outputs this manifest.mpd
http://pastebin.com/CUYZ6NDt

Myself, I would expect that audio stream be ordered like on the command line: mkd, sqi and srp.
If this is not expected behavior, perhaps it could be an acceptable suggestion for an improvement?

kqyang:
Yes, that is something we need to improve. I will take this into consideration when doing Q1 2017 planning :)

cgp1081:
@kqyang We're finding the order of audio tracks has impact on user experience.  

Roku for example, will default to English, but if * key is pressed, then menu closed, it will auto switch to Spanish if its listed first in MPD.  We can only control the experience through track order.

Currently the packager is listing Spanish first regardless or the order of arguments in the command

[...]

PR #726, opened 2020-03-09
Adding allow_codec_switching parameter

sr1990:
This PR is related to https://github.com/google/shaka-packager/issues/542 . 
Adding allow_codec_switching=true, will allow switching between different codecs, if they have the same language, media type (audio, video etc) and container type.
```
$ packager \
  in=h264_baseline_360p_600.mp4,stream=audio,output=audio.mp4 \
  in=input_text.vtt,stream=text,output=output_text.vtt \
  in=h264_baseline_360p_600.mp4,stream=video,output=h264_360p.mp4 \
  in=h264_main_480p_1000.mp4,stream=video,output=h264_480p.mp4 \
  in=h264_main_720p_3000.mp4,stream=video,output=h264_720p.mp4 \
  in=h264_high_1080p_6000.mp4,stream=video,output=h264_1080p.mp4 \
  in=h265_high_1080p_6000.mp4,stream=video,output=h265_1080p.mp4 \ 
  --mpd_output h264.mpd --allow_codec_switching=true
```


packager/app/mpd_flags.cc
```diff
@@ -62,3 +62,8 @@ DEFINE_bool(
     "completely."
     "Ignored if $Time$ is used in segment template, since $Time$ requires "
     "accurate Segment Timeline.");
+DEFINE_bool(allow_codec_switching,
```
kqyang: This block is not formatted correctly. Please follow either the blocks from line 49 - 52 or the blocks from line 53 - 64. Alternatively, you may need to use clang-format: https://github.com/google/shaka-packager/blob/master/packager/tools/git/check_formatting.py

packager/app/mpd_flags.cc
```diff
@@ -62,3 +62,8 @@ DEFINE_bool(
...
+            false,
+	    "if enabled, allow switching between different codecs, if they "
+	    "have the same language, media type (audio, video etc) and "
+	    "container type");
```
kqyang: Can you add a period, i.e. "."  after type?

packager/mpd/base/adaptation_set.cc
```diff
@@ -165,12 +165,14 @@ class RepresentationStateChangeListenerImpl
 
 AdaptationSet::AdaptationSet(const std::string& language,
                              const MpdOptions& mpd_options,
-                             uint32_t* counter)
+                             uint32_t* counter,
+                             const MediaInfo& media_info)
```
kqyang: Move "const MediaInfo& media_info" before "uint32_t* counter".

We follow Google C++ style guide, where input parameters are always placed before output parameters: https://google.github.io/styleguide/cppguide.html#Output_Parameters.

I thought about it again. This change is probably not needed if we don't plan to refactor out ProtectedContentMap. 

Instead, what you need is to change line 85 of period.cc to include the bool flag.

packager/app/mpd_flags.cc
kqyang: s/switching/adaptive switching/

packager/mpd/base/adaptation_set.h
```diff
@@ -135,6 +135,9 @@ class AdaptationSet {
   /// @param id is the new ID to be set.
   void set_id(uint32_t id) { id_ = id; }
 
+  /// Get media info.
+  const MediaInfo& getMediaInfo() const { return media_info_;}
```
kqyang: s/getMediaInfo/media_info/ to be consistent.

packager/mpd/base/mpd_utils.cc
```diff
@@ -142,7 +142,7 @@ std::string GetBaseCodec(const MediaInfo& media_info) {
   return codec;
 }
 
-std::string GetAdaptationSetKey(const MediaInfo& media_info) {
+std::string GetAdaptationSetKey(const MediaInfo& media_info, bool no_codec) {
```
kqyang: s/no_codec/ignore_codec/

packager/mpd/base/period.h
```diff
@@ -130,7 +131,8 @@ class Period {
                const MediaInfo& media_info);
     // Check if the two adaptation sets are switchable.
     bool Switchable(const AdaptationSet& adaptation_set_a,
-                    const AdaptationSet& adaptation_set_b);
+                    const AdaptationSet& adaptation_set_b,
+                    const bool allow_codec_switching);
```
kqyang: remove const here. We don't prefix primitive types with const in this code base.

packager/mpd/base/period.h
```diff
@@ -88,7 +88,8 @@ class Period {
   virtual std::unique_ptr<AdaptationSet> NewAdaptationSet(
       const std::string& lang,
       const MpdOptions& options,
-      uint32_t* representation_counter);
+      uint32_t* representation_counter,
+      const MediaInfo& media_info);
```
kqyang: Same as previous comment. Move media_info before representation_counter.

packager/mpd/base/period.cc
```diff
@@ -112,12 +112,26 @@ AdaptationSet* Period::GetOrCreateAdaptationSet(
 
     for (AdaptationSet* adaptation_set : adaptation_sets) {
       if (protected_adaptation_set_map_.Switchable(*adaptation_set,
-                                                   *new_adaptation_set)) {
+                                                   *new_adaptation_set, false)) {
```
kqyang: The code here shouldn't need to be updated if you don't plan to refactor ProtectedAdaptationSetMap out.

Instead, what you need is to change line 85 to include the bool flag.

packager/mpd/base/mpd_utils.h
```diff
@@ -47,7 +47,8 @@ std::string GetCodecs(const MediaInfo& media_info);
 std::string GetBaseCodec(const MediaInfo& media_info);
 
 // Returns a key made from the characteristics that separate AdaptationSets.
-std::string GetAdaptationSetKey(const MediaInfo& media_info);
+std::string GetAdaptationSetKey(const MediaInfo& media_info,
+           bool no_codec = false);
```
kqyang: We don't normally allow default parameters, which could hide problems. Please do not use default parameter here.

packager/mpd/base/adaptation_set.cc
sr1990: @kqyang Just including the bool flag on line 85 will not create a new adaptation set along with switching supplemental property. I will take a look into refactoring out ProtectedContentMap. Please let me know if you have any pointers.

kqyang: > Just including the bool flag on line 85 will not create a new adaptation set along with switching supplemental property

There are a few additional changes associated with that. See below:

```c++
  // AdaptationSets with the same key could differ in ContentProtection or codec (depends
  // on whether codec switching is allowed).
  const std::string key = GetAdaptationSetKey(media_info, mpd_options_.allow_codec_switching);
  std::list<AdaptationSet*>& adaptation_sets = adaptation_set_list_map_[key];

  for (AdaptationSet* adaptation_set : adaptation_sets) {
    // TODO: Change Match to check if codecs is the same too.
    if (protected_adaptation_set_map_.Match(*adaptation_set, media_info))
      return adaptation_set;
  }

  // None of the adaptation sets match with the new content protection.
  // Need a new one.
  const std::string language = GetLanguage(media_info);
  std::unique_ptr<AdaptationSet> new_adaptation_set =
      NewAdaptationSet(language, mpd_options_, representation_counter_);
  if (!SetNewAdaptationSetAttributes(language, media_info, adaptation_sets,
                                     new_adaptation_set.get())) {
    return nullptr;
  }

  if (content_protection_in_adaptation_set &&
      media_info.has_protected_content()) {
    protected_adaptation_set_map_.Register(*new_adaptation_set, media_info);
    AddContentProtectionElements(media_info, new_adaptation_set.get());
  }

  for (AdaptationSet* adaptation_set : adaptation_sets) {
    if (protected_adaptation_set_map_.Switchable(*adaptation_set,
                                                 *new_adaptation_set)) {
      adaptation_set->AddAdaptationSetSwitching(new_adaptation_set.get());
      new_adaptation_set->AddAdaptationSetSwitching(adaptation_set);
    }
  }

```

> I will take a look into refactoring out ProtectedContentMap. Please let me know if you have any pointers.

That will be great. I think we can start with the change above and we can file a separate issue for the refactoring.

sr1990: For comparing the codec in Match(),  do you think adding codec as AdaptationSet member and setting it via SetNewAdaptationSetAttributes make sense or is there any other way to access codec info?

kqyang: I just realized that you can store the codec in AdaptationSet::UpdateFromMediaInfo function.

packager/mpd/base/adaptation_set.h
sr1990: removed mediainfo from adaptation set

sr1990:
@kqyang CI jobs are failing. Please review and let me know.

packager/mpd/base/period.cc
```diff
@@ -240,11 +239,12 @@ const AdaptationSet* Period::FindOriginalAdaptationSetForTrickPlay(
   MediaInfo media_info_no_trickplay = media_info;
   media_info_no_trickplay.mutable_video_info()->clear_playback_rate();
 
-  std::string key = GetAdaptationSetKey(media_info_no_trickplay);
+  std::string key = GetAdaptationSetKey(media_info_no_trickplay, false);
```
kqyang: s/false/mpd_options_.mpd_params.allow_codec_switching/

It has to be the same as the key we use when setting up |adaptation_set_list_map_|, otherwise |adaptation_sets| below would be empty.

packager/mpd/base/period.cc
```diff
@@ -240,11 +239,12 @@ const AdaptationSet* Period::FindOriginalAdaptationSetForTrickPlay(
...
   const std::list<AdaptationSet*>& adaptation_sets =
       adaptation_set_list_map_[key];
   for (AdaptationSet* adaptation_set : adaptation_sets) {
-    if (protected_adaptation_set_map_.Match(*adaptation_set, media_info)) {
+    if (protected_adaptation_set_map_.Match(*adaptation_set, media_info,
+                                            true)) {
```
kqyang: Looks like this is a bug in the original code. The flag content_protection_in_adaptation_set should have the same value as the one in line 93.

Can you pass content_protection_in_adaptation_set to SetNewAdaptationSetAttributes then to  FindOriginalAdaptationSetForTrickPlay?

packager/mpd/base/adaptation_set.cc
```diff
@@ -412,10 +412,16 @@ void AdaptationSet::UpdateFromMediaInfo(const MediaInfo& media_info) {
```
kqyang: Prefer just set 

codec_ = GetCodecs(media_info);

packager/mpd/base/period.cc
```diff
@@ -260,7 +260,38 @@ void Period::ProtectedAdaptationSetMap::Register(
...
+        return false;
+
+    } else if (media_info.has_text_info() &&
+               media_info.text_info().has_codec()) {
+      if (adaptation_set_codec.compare(base::SplitString(
+              media_info.text_info().codec(), ".", base::TRIM_WHITESPACE,
+              base::SPLIT_WANT_NONEMPTY)[0]) != 0)
+        return false;
+    }
+  }
```
kqyang: Is SplitString needed? 

with the change suggested above in line 412 of adaptation_set.cc, we can just do:

if (adaptation_set.codec() != GetCodecs(media_info))
  return false;

packager/mpd/base/adaptation_set.h
```diff
@@ -269,6 +269,9 @@ class AdaptationSet {
   // Determined by examining the MediaInfo passed to AddRepresentation().
   std::string content_type_;
 
+  // Codec of AdaptationSet
+  std::string codec_;
```
kqyang: Also define an accessor for codec_ after line 178:

/// @return codec.
std::string codec() const { return codec_; }

packager/app/test/packager_test.py
```diff
@@ -1499,6 +1503,17 @@ def testLiveStaticProfileWithTimeInSegmentName(self):
...
+    streams = [
+        self._GetStream('video', test_file='bear-640x360-hevc.mp4'),
+        self._GetStream('video', test_file='bear-640x360.mp4'),
+        self._GetStream('video', test_file='bear-1280x720.mp4'),
+        self._GetStream('audio', test_file='bear-640x360.mp4'),
+    ]
+
+    self.assertPackageSuccess(streams, self._GetFlags(output_dash=True,allow_codec_switching=True))
+    self._CheckTestResults('audio-video-with-codec-switching')
+
```
kqyang: Can you add another variant with encryption and trick play?

```python
  def testAllowCodecSwitchingWithEncryptionAndTrickplay(self):
    streams = [
        self._GetStream('video', test_file='bear-640x360-hevc.mp4'),
        self._GetStream('video', test_file='bear-640x360.mp4'),
        self._GetStream('video', test_file='bear-1280x720.mp4'),
        self._GetStream('video', test_file='bear-1280x720.mp4', trick_play_factor=1),
        self._GetStream('audio', test_file='bear-640x360.mp4'),
    ]

    self.assertPackageSuccess(streams,
                              self._GetFlags(output_dash=True, 
                                             allow_codec_switching=True, 
                                             encryption=True))
    # Mpd cannot be validated right now since we don't generate determinstic
    # mpd with multiple inputs due to thread racing.
    # TODO(b/73349711): Generate determinstic mpd or at least validate mpd
    #                   schema.
    # See also https://github.com/google/shaka-packager/issues/177.
```

packager/app/test/packager_test.py
```diff
@@ -1499,6 +1503,17 @@ def testLiveStaticProfileWithTimeInSegmentName(self):
...
+  def testAllowCodecSwitching(self):
+    streams = [
+        self._GetStream('video', test_file='bear-640x360-hevc.mp4'),
+        self._GetStream('video', test_file='bear-640x360.mp4'),
+        self._GetStream('video', test_file='bear-1280x720.mp4'),
+        self._GetStream('audio', test_file='bear-640x360.mp4'),
+    ]
+
+    self.assertPackageSuccess(streams, self._GetFlags(output_dash=True,allow_codec_switching=True))
+    self._CheckTestResults('audio-video-with-codec-switching')
```
kqyang: Unfortunately this line has to be removed as we don't generate determinstic mpd with multiple inputs right now.

Replace it with:

    # Mpd cannot be validated right now since we don't generate determinstic
    # mpd with multiple inputs due to thread racing.
    # TODO(b/73349711): Generate determinstic mpd or at least validate mpd
    #                   schema.
    # See also https://github.com/google/shaka-packager/issues/177.

packager/mpd/base/period.cc
sr1990: GetCodecs returns the entire codec string (avc1.4d400d / avc1.4d401f) if container type is not webm. Should I change the GetCodecs to return media_info.video_info().codec().substr(0, 4); for all container types or keep the SplitString in Match?

kqyang: Ah, good catch, we shouldn't be using the entire codec string as two different representations in the same adaptation set could have two different codec strings.

Please replace GetCodecs with GetBaseCodec instead.

sr1990: TrickPlayWithMatchingAdaptationSet unit test is failing as the codec in default_adaptation_set_ is no call to UpdateFromMediaInfo.  
So, instead of setting codec from AdaptationSet::UpdateFromMediaInfo do you think it would make sense to set it at Period::SetNewAdaptationSetAttributes. If not, how do you change the unit test? I tried calling the UpdateFromMediaInfo directly but that did not work.

packager/mpd/base/period.cc
sr1990: Yes.

packager/mpd/base/period.cc
kqyang: Sure. You may add a AdaptationSet::set_codec(const std::string& codec) function and call it in Period::SetNewAdaptationSetAttributes. 

(This tells me that we should really refactor the code later to clean up the code and remove ProtectedContentMap)

sr1990:
@googlebot I consent.

packager/app/test/packager_test.py
```diff
@@ -1499,6 +1503,42 @@ def testLiveStaticProfileWithTimeInSegmentName(self):
...
+    #                   schema.
+    # See also https://github.com/google/shaka-packager/issues/177.
+
+  def testAllowCodecSwitchingWithEncryptionAndTrickplay(self):
+    streams = [
+        self._GetStream('video', test_file='bear-640x360-hevc.mp4'),
+        self._GetStream('video', test_file='bear-640x360.mp4'),
+        self._GetStream('video', test_file='bear-1280x720.mp4'),
+        self._GetStream('video', test_file='bear-1280x720.mp4',
+                                           trick_play_factor=1),
```
kqyang: I think trick_play_factor should align with 'video' above. You may use 'pyformat' to align it properly (there could be other formatting issues in this file, so it is fine to leave it as is for now and do it in a separate pull request or leave it to us to address).

packager/mpd/base/adaptation_set.h
```diff
@@ -176,6 +176,13 @@ class AdaptationSet {
   /// @return true if it is a video AdaptationSet.
   bool IsVideo() const;
 
+  /// @return codec.
+  std::string codec() const { return codec_; }
```
kqyang: Use const reference to avoid unnecessary copying:

const std::string& codec() const { return codec_; }

packager/mpd/base/period.cc
```diff
@@ -7,6 +7,7 @@
 #include "packager/mpd/base/period.h"
 
 #include "packager/base/stl_util.h"
+#include "packager/base/strings/string_split.h"
```
kqyang: Is this still needed?

packager/mpd/base/period.cc
```diff
@@ -176,7 +173,8 @@ bool Period::SetNewAdaptationSetAttributes(
     const std::string& language,
     const MediaInfo& media_info,
     const std::list<AdaptationSet*>& adaptation_sets,
-    AdaptationSet* new_adaptation_set) {
+    AdaptationSet* new_adaptation_set,
+    bool content_protection_in_adaptation_set) {
```
kqyang: Swap the order of the last two parameters as Google C++ style guide requires the input argument to be placed before the output argument: https://google.github.io/styleguide/cppguide.html#Output_Parameters

packager/mpd/base/period.cc
sr1990: no. removed it.

packager/app/test/packager_test.py
kqyang: Ooops, I forgot one additional line here in my previous comment:

```python
    self._CheckTestResults(
        'audio-video-with-codec-switching-encryption-trick-play',
        diff_files_policy=DiffFilesPolicy(
            allowed_diff_files=['output.mpd'], exact=False))
```

Similar for the test above:

```
    self._CheckTestResults(
        'audio-video-with-codec-switching',
        diff_files_policy=DiffFilesPolicy(
            allowed_diff_files=['output.mpd'], exact=False))
```

packager/mpd/public/mpd_params.h
```diff
@@ -80,6 +80,9 @@ struct MpdParams {
   /// SegmentTimeline if it is enabled. It will be populated from segment
   /// duration specified in ChunkingParams if not specified.
   double target_segment_duration = 0;
+  /// If enabled, allow switching between different codecs, if they have the
+  /// same language, media type (audio, video etc) and container type.
+  bool allow_codec_switching = false;
```
kqyang: Normally we don't pull changes from master, which may cause confusions to the bots as seen in this PR. I'll see if I can merge it with this later.

sr1990:
@kqyang Do you think it makes sense to check and add audio channel count to adaptation set key in mpd_utils.cc:GetAdaptationSetKey when allow_codec_switching parameter is set to true ?

kqyang:
Audio streams with different number of channels can be put in the same AdaptationSet, regardless of the `allow_codec_switching` flag, so I don't think it should be included in `GetAdaptationSetKey`. Let me know if you have a different opinion.

sr1990:
I think adaptive switch between audio with different channels won't be seamless. On Exoplayer side, there is a flag allowAudioMixedChannelCountAdaptiveness in DefaultTrackSelector which is set to false by default and it filters out the audio track with appropriate channel count at: https://github.com/google/ExoPlayer/blob/e822196336b64208ba7caa557e4c6ee02b03335e/library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java#L1647 and 
as per issue https://github.com/google/ExoPlayer/issues/6497: 
`"Also, by default, we won't allow adaptation between different channel counts (i.e. stereo and 5.1) because there may be an audible discontinuity when we switch formats."`
This flag has been added in 2.10.5 as per https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md. 
So for clients depending on older versions of exoplayer and other players that do not handle this case, we could add a flag (--allow_audio_mixed_channel_count set to true by default) on the packager side to disable audio representations with different channel counts in the same adaptation set. What do you think?

sr1990:
I think adaptive switch between audio with different channels won't be seamless.
On Exoplayer side, there is a flag allowAudioMixedChannelCountAdaptiveness in DefaultTrackSelector which is set to false by default and it filters out the audio track with appropriate channel count at: https://github.com/google/ExoPlayer/blob/e822196336b64208ba7caa557e4c6ee02b03335e/library/core/src/main/java/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.java#L1647 and
as per issue https://github.com/google/ExoPlayer/issues/6497:
`"Also, by default, we won't allow adaptation between different channel counts (i.e. stereo and 5.1) because there may be an audible discontinuity when we switch formats."`
This flag has been added in 2.10.5 as per https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md.

So for players that do not handle this case, we could add a flag (--allow_audio_mixed_channel_count set to true by default) on the packager side to disable audio representations with different channel counts in the same adaptation set or we could let the players handle it. What do you think?

kqyang:
@sr1990 So ExoPlayer already handles the case of a single AdaptationSet with audio streams of different number of channels, the feature for packager is not helpful for ExoPlayer. 

It seems that it may create problems instead, as people would have to package two sets of contents if they want to enable adaptive switching between different audio streams in one platform but disable it in another platform.

sr1990:
`It seems that it may create problems instead, as people would have to package two sets of contents if they want to enable adaptive switching between different audio streams in one platform but disable it in another platform`
I agree. Thanks.

Issue #724, opened 2020-03-06
Initial gclient sync on windows error 


ab5y:
## System info
**Operating System**: WIndows 10 Home
**Shaka Packager Version**: commit 055c67888b5e75b2c0bc88ac58da24e1aec86667

## Issue and steps to reproduce the problem
Followed the steps for building from release on Windows. Got the dependencies installed, ran the gclient config command and then ran the gclient sync command for the first time.

**Packager Command**: gclient sync

Extra steps to reproduce the problem?
None

What is the expected result?
Successful installation of dependencies

What happens instead?
![image](https://user-images.githubusercontent.com/4251453/76112147-a9524480-6007-11ea-8997-c64d7d818ed2.png)

<Please attach the input files or email to shaka-packager-issues@google.com.>
<Please provide any additional information below.>

kqyang:
Do you have `python` installed? What happens if you try to run `python` in the command line?

ab5y:
I have both python2 (2.7.17) and python3 (3.8.0) installed. The `python` command runs python2.

ab5y:
Found [this explanation of the error on S/O ](https://stackoverflow.com/questions/9531683/problems-using-subprocess-call-in-python-2-7-2-on-windows). 
I added the `shell=True` argument to lines 36 and 45 of the `generate_version_string.py` file. It now works, but the version string it generates says "unknown-version". Would this cause an issue?

_Code modification_:
![image](https://user-images.githubusercontent.com/4251453/76162206-4eccfb80-6161-11ea-8010-c51113e913ea.png)

kqyang:
@ab5y What happens if you run `git rev-parse --short HEAD` in command line?

ab5y:
The `git rev-parse --short HEAD` gave the following output:
`055c67888b`
I checked with the main repo and saw that my copy is behind. So I pulled the latest version. This time, with the above changes to the `generate_version_string.py` file, a version string does get printed. But without the changes, the `sync` command still throws the same error. Should I create a pull request?

kqyang:
@ab5y Yes, please.

ab5y:
Could you tell me on which branch to make the request on?

PR #728, opened 2020-03-11
Fix the error thrown by runhooks command on Windows [Issue 724]

ab5y:
As outlined [here](https://github.com/google/shaka-packager/issues/724), the `gclient sync` command was throwing a `python` `subprocess` related error. The call to the subprocess library required an additional `shell=True` argument to be passed in the `generate_version_string.py` file.

### DIFF
--- a/packager/version/generate_version_string.py
+++ b/packager/version/generate_version_string.py
@@ -32,7 +32,7 @@ if 'check_output' not in dir(subprocess):
 if __name__ == '__main__':
   try:
     version_tag = subprocess.check_output(
-        ['git tag --points-at HEAD'],
+        'git tag --points-at HEAD',
         stderr=subprocess.STDOUT, shell=True).rstrip()
   except subprocess.CalledProcessError as e:
     # git tag --points-at is not supported in old versions of git. Just ignore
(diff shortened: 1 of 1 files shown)


ab5y:
@googlebot I signed it!

packager/version/generate_version_string.py
```diff
@@ -33,7 +33,7 @@ def check_output_implementation(*popenargs, **kwargs):
   try:
     version_tag = subprocess.check_output(
         ['git', 'tag', '--points-at', 'HEAD'],
-        stderr=subprocess.STDOUT).rstrip()
+        stderr=subprocess.STDOUT, shell=True).rstrip()
```
kqyang: This is failing on Linux and MAC. I think you need to change ['git', 'tag', '--points-at', 'HEAD'] to ['git tag --points-at HEAD']. The same below.

kqyang: The new code succeeds in Linux / MAC, but fails on Windows, can you try if removing "[]" works on both Linux and Windows?

If not, we'll have to make shell=True conditional on Windows.

kqyang:
@ab5y Thanks for your help. The change has been merged.

Issue #732, opened 2020-03-16
[DASH] Support TrickPlay using separate trick play specific streams


ngbrown:
## System info
**Operating System**: Windows 10 10.0.18362.657
**Shaka Packager Version**: v2.4.1-c731217607-release

## Issue and steps to reproduce the problem

**Packager Command**:

```cmd
packager-win.exe ^
in=audio_48k.m4a,stream=audio,output=output/audio_48k.m4a,playlist_name=output/audio_48k.m3u8,hls_group_id=audio,hls_name=ENGLISH ^
in=audio_128k.m4a,stream=audio,output=output/audio_128k.m4a,playlist_name=output/audio_128k.m3u8,hls_group_id=audio,hls_name=ENGLISH ^
in=360p.m4v,stream=video,output=output/360p.m4v,playlist_name=output/360p.m3u8,iframe_playlist_name=output/360p_iframe.m3u8 ^
in=480p.m4v,stream=video,output=output/480p.m4v,playlist_name=output/480p.m3u8,iframe_playlist_name=output/480p_iframe.m3u8 ^
in=720p.m4v,stream=video,output=output/720p.m4v,playlist_name=output/720p.m3u8,iframe_playlist_name=output/720p_iframe.m3u8 ^
in=1080p.m4v,stream=video,output=output/1080p.m4v,playlist_name=output/1080p.m3u8,iframe_playlist_name=output/1080p_iframe.m3u8 ^
in=trick_480p.m4v,stream=video,output=output/trick_480p.m4v,trick_play_factor=1,playlist_name=output/trick_480p.m3u8 ^
in=trick_1080p.m4v,stream=video,output=output/trick_1080p.m4v,trick_play_factor=1,playlist_name=output/trick_1080p.m3u8 ^
--hls_master_playlist_output output/master.m3u8 ^
--mpd_output output/master.mpd --generate_static_live_mpd
```

Extra steps to reproduce the problem?
1. Generate a separate trick play track instead of sharing with main video.  The main video has I-frames every 6 seconds, while the trick play track should be 1 fps with 1 I-Frame per second.
2. Run repeatedly

What is the expected result?

Sometimes it succeeds:

```
[0316/093610:INFO:demuxer.cc(88)] Demuxer::Run() on file '1080p.m4v'.
[0316/093610:INFO:demuxer.cc(88)] Demuxer::Run() on file '480p.m4v'.
[...]

PR #736, opened 2020-03-20
[DASH] Support TrickPlay using separate trick play specific streams

sr1990:
This PR is related to https://github.com/google/shaka-packager/issues/732

### DIFF
--- a/packager/mpd/base/period.cc
+++ b/packager/mpd/base/period.cc
@@ -216,22 +216,28 @@ bool Period::SetNewAdaptationSetAttributes(
     }
 
     if (media_info.video_info().has_playback_rate()) {
+      std::string trick_play_reference_adaptation_set_key;
       AdaptationSet* trick_play_reference_adaptation_set =
           FindMatchingAdaptationSetForTrickPlay(
-              media_info, content_protection_in_adaptation_set);
+              media_info, content_protection_in_adaptation_set,
+              &trick_play_reference_adaptation_set_key);
       if (trick_play_reference_adaptation_set) {
         new_adaptation_set->AddTrickPlayReference(
             trick_play_reference_adaptation_set);
       } else {
-        std::string key = GetAdaptationSetKeyForTrickPlay(media_info);
-        trickplay_cache_[key].push_back(new_adaptation_set);
+        trickplay_cache_[trick_play_reference_adaptation_set_key].push_back(
+            new_adaptation_set);
       }
     } else {
+      std::string trick_play_adaptation_set_key;
       AdaptationSet* trickplay_adaptation_set =
           FindMatchingAdaptationSetForTrickPlay(
-              media_info, content_protection_in_adaptation_set);
-      if (trickplay_adaptation_set)
+              media_info, content_protection_in_adaptation_set,
+              &trick_play_adaptation_set_key);
+      if (trickplay_adaptation_set) {
         trickplay_adaptation_set->AddTrickPlayReference(new_adaptation_set);
+        trickplay_cache_.erase(trick_play_adaptation_set_key);
+      }
     }
 
   } else if (media_info.has_text_info()) {
@@ -245,28 +251,28 @@ bool Period::SetNewAdaptationSetAttributes(
 
 AdaptationSet* Period::FindMatchingAdaptationSetForTrickPlay(
     const MediaInfo& media_info,
-    bool content_protection_in_adaptation_set) {
+    bool content_protection_in_adaptation_set,
+    std::string* adaptation_set_key) {
   std::list<AdaptationSet*>* adaptation_sets = nullptr;
-  std::string key;
   const bool is_trickplay_adaptation_set =
       media_info.video_info().has_playback_rate();
   if (is_trickplay_adaptation_set) {
-    MediaInfo media_info_no_trickplay = media_info;
-    key = Period::GetAdaptationSetKeyForTrickPlay(media_info);
-    adaptation_sets = &adaptation_set_list_map_[key];
-  } else {
-    key = GetAdaptationSetKey(media_info,
-                              mpd_options_.mpd_params.allow_codec_switching);
-    if (trickplay_cache_.find(key) == trickplay_cache_.end())
+    *adaptation_set_key = GetAdaptationSetKeyForTrickPlay(media_info);
+    if (adaptation_set_list_map_.find(*adaptation_set_key) ==
+        adaptation_set_list_map_.end())
       return nullptr;
-    adaptation_sets = &trickplay_cache_[key];
+    adaptation_sets = &adaptation_set_list_map_[*adaptation_set_key];
+  } else {
+    *adaptation_set_key = GetAdaptationSetKey(
+        media_info, mpd_options_.mpd_params.allow_codec_switching);
+    if (trickplay_cache_.find(*adaptation_set_key) == trickplay_cache_.end())
+      return nullptr;
+    adaptation_sets = &trickplay_cache_[*adaptation_set_key];
   }
   for (AdaptationSet* adaptation_set : *adaptation_sets) {
     if (protected_adaptation_set_map_.Match(
             *adaptation_set, media_info,
             content_protection_in_adaptation_set)) {
-      if (!is_trickplay_adaptation_set && !key.empty())
-        trickplay_cache_.erase(key);
       return adaptation_set;
     }
   }
--- a/packager/mpd/base/period.cc
+++ b/packager/mpd/base/period.cc
@@ -117,6 +117,18 @@ AdaptationSet* Period::GetOrCreateAdaptationSet(
   AdaptationSet* adaptation_set_ptr = new_adaptation_set.get();
   adaptation_sets.push_back(adaptation_set_ptr);
   adaptation_sets_.emplace_back(std::move(new_adaptation_set));
+
+  if (!trickplay_cache.empty()) {
+    if (trickplay_cache.find(key) != trickplay_cache.end()) {
+      std::list<AdaptationSet*> adaptation_set_list = trickplay_cache[key];
+      for (AdaptationSet* adaptation_set : adaptation_set_list) {
+        adaptation_set->AddTrickPlayReference(adaptation_set_ptr);  		
+      }
+      
+      trickplay_cache.erase(key);
+    }
+  }
+
   return adaptation_set_ptr;
 }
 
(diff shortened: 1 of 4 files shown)


packager/mpd/base/period.cc
```diff
@@ -117,6 +117,18 @@ AdaptationSet* Period::GetOrCreateAdaptationSet(
   AdaptationSet* adaptation_set_ptr = new_adaptation_set.get();
   adaptation_sets.push_back(adaptation_set_ptr);
   adaptation_sets_.emplace_back(std::move(new_adaptation_set));
+
+  if (!trickplay_cache.empty()) {
+    if (trickplay_cache.find(key) != trickplay_cache.end()) {
+      std::list<AdaptationSet*> adaptation_set_list = trickplay_cache[key];
+      for (AdaptationSet* adaptation_set : adaptation_set_list) {
+        adaptation_set->AddTrickPlayReference(adaptation_set_ptr);
```
kqyang: Need to check whether content protection matches too as in line 249. 

A few suggestions:

1. I think we can move the code to around line 222 which is better for readability, so something like:

```c++
if (media_info.video_info().has_playback_rate()) {
  // Cache trick play adaptation set if there is any.
  ...
} else {
  // Check if the adaptation set matches with any of the cached trick play adaptation set.
  ...
}
```

2. I think we can reuse FindOriginalAdaptationSetForTrickPlay here, so that we can use the function 
    - To find the original AdaptationSet of the Trick Play AdaptationSet
    - To find the trick play AdaptationSet associated with the current non-trick-play AdaptationSet
    
    The function name needs to be renamed, the result code will be something like this:

```c++
const AdaptationSet* Period::FindMatchingAdaptationSetForTrickPlay(
    const MediaInfo& media_info,
    bool content_protection_in_adaptation_set) {
  std::list<AdaptationSet*>* adaptation_sets = nullptr;
  const bool is_trickplay_adaptation_set = media_info.video_info().has_playback_rate();
  if (is_trickplay_adaptation_set) {
    MediaInfo media_info_no_trickplay = media_info;
    media_info_no_trickplay.mutable_video_info()->clear_playback_rate();
    std::string key = GetAdaptationSetKey(
        media_info_no_trickplay, mpd_options_.mpd_params.allow_codec_switching);
    adaptation_sets = &adaptation_set_list_map_[key];
  } else {
    std::string key = GetAdaptationSetKey(
        media_info, mpd_options_.mpd_params.allow_codec_switching);
    if (trickplay_cache.find(key) == trickplay_cache.end()) 
      return nullptr;
    adaptation_sets = &trickplay_cache[key];
  }
  for (AdaptationSet* adaptation_set : *adaptation_sets) {
    if (protected_adaptation_set_map_.Match(
            *adaptation_set, media_info,
            content_protection_in_adaptation_set)) {
      if (!is_trickplay_adaptation_set)
        trickplay_cache.erase(key);
      return adaptation_set;
    }
  }
  return nullptr;
}
```

packager/mpd/base/period.cc
```diff
@@ -220,11 +232,16 @@ bool Period::SetNewAdaptationSetAttributes(
           FindOriginalAdaptationSetForTrickPlay(
               media_info, content_protection_in_adaptation_set);
       if (!trick_play_reference_adaptation_set) {
-        LOG(ERROR) << "Failed to find original AdaptationSet for trick play.";
-        return false;
+        MediaInfo media_info_no_trickplay = media_info;
+        media_info_no_trickplay.mutable_video_info()->clear_playback_rate();
+        std::string key =
+            GetAdaptationSetKey(media_info_no_trickplay,
+                                mpd_options_.mpd_params.allow_codec_switching);
```
kqyang: Suggest creating a private function in Period to avoid code duplication:

```c++
std::string Period::GetAdaptationSetKeyForTrickPlay(const MediaInfo& media_info) {
  MediaInfo media_info_no_trickplay = media_info;
  media_info_no_trickplay.mutable_video_info()->clear_playback_rate();
  return GetAdaptationSetKey(media_info_no_trickplay,
                             mpd_options_.mpd_params.allow_codec_switching);
}
```

packager/mpd/base/period.cc
```diff
@@ -303,4 +320,11 @@ bool Period::ProtectedAdaptationSetMap::Switchable(
          GetUUIDs(protected_content_it_b->second);
 }
 
+Period::~Period() {
+  if (!trickplay_cache.empty()) {
+    LOG(ERROR) << "Trickplay adaptation set did not get a valid adaptation set "
```
kqyang: I think LOG(WARNING) is more appropriate here.

packager/mpd/base/period.cc
```diff
@@ -303,4 +320,11 @@ bool Period::ProtectedAdaptationSetMap::Switchable(
...
+                  "match. Please check the command line options";
```
kqyang: Add an ending period "." after options.

packager/mpd/base/period.h
```diff
@@ -117,6 +124,10 @@ class Period {
   // if they contain identical ContentProtection elements. This map is only
   // useful when ContentProtection element is placed in AdaptationSet.
   std::map<std::string, std::list<AdaptationSet*>> adaptation_set_list_map_;
+  // Contains Trickplay AdaptationSets grouped by specific adaptation set
+  // grouping key. These AdaptationSets still have not found reference
+  // AdaptationSet.
+  std::map<std::string, std::list<AdaptationSet*>> trickplay_cache;
```
kqyang: Google C++ style guide requires member variables to end with "_", so rename it to trickplay_cache_.

packager/mpd/base/period_unittest.cc
```diff
@@ -316,10 +316,12 @@ TEST_F(PeriodTest, TrickPlayWithNoMatchingAdaptationSet) {
...
                 content_protection_in_adaptation_set_));
-  // A nullptr is returned if it is not able to find matching AdaptationSet.
-  ASSERT_FALSE(testable_period_.GetOrCreateAdaptationSet(
+
+  ASSERT_TRUE(testable_period_.GetOrCreateAdaptationSet(
       ConvertToMediaInfo(kVp9TrickPlayMediaInfo),
       content_protection_in_adaptation_set_));
+
+  ASSERT_TRUE(!testable_period_.GetTrickPlayCache().empty());
 }
```
kqyang: Can you add another unit-test similar to TrickPlayWithMatchingAdaptationSet but with operation at line 272 and 276 reversed to test the new code?

packager/mpd/base/period.h
```diff
@@ -66,6 +66,12 @@ class Period {
     duration_seconds_ = duration_seconds;
   }
 
+  /// @return trickplay_cache.
+  const std::map<std::string, std::list<AdaptationSet*>> GetTrickPlayCache()
```
kqyang: You can name the function as |trickplay_cache| with the variable |trickplay_cache| renamed to |trickplay_cache_|.

packager/mpd/base/period.cc
```diff
@@ -235,26 +243,45 @@ bool Period::SetNewAdaptationSetAttributes(
...
-      adaptation_set_list_map_[key];
-  for (AdaptationSet* adaptation_set : adaptation_sets) {
+  std::list<AdaptationSet*>* adaptation_sets = nullptr;
+  std::string key;
+  const bool is_trickplay_adaptation_set =
+      media_info.video_info().has_playback_rate();
+  if (is_trickplay_adaptation_set) {
+    MediaInfo media_info_no_trickplay = media_info;
+    key = Period::GetAdaptationSetKeyForTrickPlay(media_info);
+    adaptation_sets = &adaptation_set_list_map_[key];
```
kqyang: Should check whether key is in adaptation_set_list_map_ too:

```c++
if (adaptation_set_list_map_.find(key) == adaptation_set_list_map_.end()) 
  return nullptr;
```

packager/mpd/base/period.cc
```diff
@@ -252,26 +243,45 @@ bool Period::SetNewAdaptationSetAttributes(
...
+                              mpd_options_.mpd_params.allow_codec_switching);
+    if (trickplay_cache_.find(key) == trickplay_cache_.end())
+      return nullptr;
+    adaptation_sets = &trickplay_cache_[key];
+  }
+  for (AdaptationSet* adaptation_set : *adaptation_sets) {
     if (protected_adaptation_set_map_.Match(
             *adaptation_set, media_info,
             content_protection_in_adaptation_set)) {
+      if (!is_trickplay_adaptation_set && !key.empty())
```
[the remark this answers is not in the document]
sr1990: Key should not be empty. Will remove the check.

packager/mpd/base/period.cc
```diff
@@ -235,26 +243,45 @@ bool Period::SetNewAdaptationSetAttributes(
...
+        trickplay_cache_.erase(key);
```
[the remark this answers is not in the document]
sr1990: Makes sense. Will move the code.

packager/mpd/base/period.cc
```diff
@@ -216,16 +216,30 @@ bool Period::SetNewAdaptationSetAttributes(
...
-          FindOriginalAdaptationSetForTrickPlay(
-              media_info, content_protection_in_adaptation_set);
-      if (!trick_play_reference_adaptation_set) {
-        LOG(ERROR) << "Failed to find original AdaptationSet for trick play.";
-        return false;
+      std::string trick_play_reference_adaptation_set_key;
+      AdaptationSet* trick_play_reference_adaptation_set =
+          FindMatchingAdaptationSetForTrickPlay(
+              media_info, content_protection_in_adaptation_set,
+              &trick_play_reference_adaptation_set_key);
```
kqyang: The code LGTM with one optional nit: you may move this code outside of the "if else" block as it is used by both cases. 

Btw, have you verified that it addresses #732?

I'll take another look tomorrow and merge the code. Thanks very much for your help in addressing this issue.

sr1990: @kqyang Thanks for reviewing the code. I was able to test the 1 fps stream with trick_play_factor=1 as per issue #732 and the output mpd was correct.
I thought about removing the key declaration and call to FindMatchingAdaptationSetForTrickplay before the "if else" block but kept it for readability. Please let me know what you think.

kqyang: Sure, I feel moving it outside can be a bit more readable (the variables need to be renamed), but I am fine either ways.

Thanks for the hard work on this feature. Merging it now.

Issue #708, opened 2020-02-07
DASH stream descriptor field "roles" not taken into account for WebVTT text streams


francois-cabrol:
## System info
**Operating System**: Windows 10
**Shaka Packager Version**: v2.4.1-c731217607-release

## Issue and steps to reproduce the problem

**Packager Command**: packager in=TestVideo_2160p23.mp4,stream=video,output=TestVideo_2160p23.mp4 in=TestVideo_lcaac_AraSPK.mp4,stream=audio,output=TestVideo_lcaac_AraSPK.mp4 in=TestVideo_lcaac_EngSPK.mp4,stream=audio,output=TestVideo_lcaac_EngSPK.mp4 in=TestVideo_lcaac_FraSPK.mp4,stream=audio,output=TestVideo_lcaac_FraSPK.mp4 in=TestVideo_lcaac_ZhoSPK.mp4,stream=audio,output=TestVideo_lcaac_ZhoSPK.mp4 in=TestVideo_Subtitle_AraCC.vtt,stream=text,lang=ar,dash_roles=caption,out=TestVideo_Subtitle_AraCC.vtt in=TestVideo_Subtitle_EngCC.vtt,stream=text,lang=en,dash_roles=caption,out=TestVideo_Subtitle_EngCC.vtt in=TestVideo_Subtitle_FraSUB.vtt,stream=text,lang=fr,dash_roles=subtitle,out=TestVideo_Subtitle_FraSUB.vtt in=TestVideo_Subtitle_ZhoSUB.vtt,stream=text,lang=zh,dash_roles=subtitle,out=TestVideo_Subtitle_ZhoSUB.vtt --mpd_output master.mpd 

**Extra steps to reproduce the problem?**
N/A

**What is the expected result?**
In the description of the WebVTT streams, I expect a <role> tag with the values specified in the **dash_roles** field (caption or subtitle) nested in the **AdaptationSet** or the **Representation** tags.

**What happens instead?**
No <role> tag is generated.  Here is the extract of the **master.mpd** dealing with the .vtt files:
` <Period id="0">
    <AdaptationSet id="0" contentType="text" lang="ar" subsegmentAlignment="true">
      <Representation id="0" bandwidth="256" mimeType="text/vtt">
        <BaseURL>TestVideo_Subtitle_AraCC.vtt</BaseURL>
      </Representation>
    </AdaptationSet>
    <AdaptationSet id="1" contentType="text" lang="en" subsegmentAlignment="true">
[...]

PR #740, opened 2020-03-26
Consider DASH stream descriptor field "roles" for WebVTT text streams.

sr1990:
This PR is related to #708.

### DIFF
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -402,6 +402,12 @@ bool StreamInfoToTextMediaInfo(const StreamDescriptor& stream_descriptor,
     text_media_info->set_bandwidth(kDefaultTextBandwidth);
   }
 
+  if (!stream_descriptor.dash_roles.empty()) {
+    for (auto dash_role : stream_descriptor.dash_roles) {
+      text_media_info->add_dash_roles(dash_role);
+    }
+  }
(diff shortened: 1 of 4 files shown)


packager/packager.cc
```diff
@@ -402,6 +402,12 @@ bool StreamInfoToTextMediaInfo(const StreamDescriptor& stream_descriptor,
     text_media_info->set_bandwidth(kDefaultTextBandwidth);
   }
 
+  if (!stream_descriptor.dash_roles.empty()) {
+    for (auto dash_role : stream_descriptor.dash_roles) {
```
kqyang: @sr1990 Thanks for addressing this bug! Have you verified that the change works? Can you update the end to end test? You may add a text stream with role to this test: https://github.com/google/shaka-packager/blob/master/packager/app/test/packager_test.py#L671.

Use constant reference to avoid copying:

```c++
for (const auto& dash_role : ...)
```

sr1990: @kqyang Thanks for reviewing. Yes I have verified that the change works. Have added text stream to the mentioned test in packager_test.py.

kqyang: @sr1990 Cool. Thanks. I have merged the change.

Issue #757, opened 2020-04-24
seg fault on shaka::media::webm::Segmenter::Finalize() with empty segment


murillo128:
## System info
**Ubuntu 18.04
**master branch

## Issue and steps to reproduce the problem
I have a media server that receives data over webrtc and produces mpeg dash with shaka packager. Sometimes I am getting a segmentation fault on shaka::media::webm::Segmenter::Finalize()

I think it is because the  medias server didn't receive any more data on the video track before finalizing all the segments and the prev_sample in https://github.com/google/shaka-packager/blob/master/packager/media/formats/webm/segmenter.cc#L154 is null.

Checking the prev_sample for null would just solve the issue.

kqyang:
@murillo128 I think this happens only if it is not receiving any samples, i.e. the stream is empty. Can you confirm if that is the case?

murillo128:
I think this is the case, yes. Anyway adding a check would be harmless and not cause a regression

kqyang:
@murillo128 I am wondering if the check should be added in an earlier stage. In any case, you are welcomed to send a pull request to us!

murillo128:
Done.

If it is in the last segment, I could send an EOS frame, but I don't know how could I fix it from my app if I don't receive any media frame during the segment duration.

PR #759, opened 2020-04-24
Prevent seg fault when finalizing empty segment

murillo128:
as per #757


kqyang:
Thanks. The change has been merged.

PR #761, opened 2020-04-27
Adding check for creating captions with dash_only and hls_only.

sr1990:
- As per @popraduvmp suggestion. 
- Correcting the documentation.
- Adding unit test along with caption.


packager/app/test/packager_test.py
```diff
@@ -719,6 +719,36 @@ def testAudioVideoWithTwoTrickPlayDecreasingRate(self):
...
+      'segment_template=%s-$Number$.m4s' %
+      (test_file, audio_output_prefix, audio_output_prefix),
+      'input=%s,stream=text,init_segment=%s-init.mp4,'
+      'segment_template=%s-$Number$.m4s,dash_only=1' %
+      (text_file, text_output_prefix, text_output_prefix),
+      'input=%s,stream=text,'
+      'segment_template=%s-$Number$.vtt,hls_only=1' %
+      (text_file, text_output_prefix),
+
+    ],
```
kqyang, line 748: Thanks for adding this end to end test. 

Can you replace it with something like below, which is cleaner?

```python
audio_video_streams = self._GetStreams(['audio', 'video'], segmented=True)
dash_text_stream = self._GetStreams(['text'],
                                    test_files=['bear-english.vtt'], 
                                    output_format='mp4',
                                    segmented=True,
                                    dash_only=True)
hls_text_stream = self._GetStreams(['text'],
                                    test_files=['bear-english.vtt'], 
                                    segmented=True,
                                    hls_only=True)
streams = audio_video_streams + dash_text_stream + hls_text_stream
self.assertPackageSuccess(streams, self._GetFlags(output_dash=True, output_hls=True))
```

packager/packager.cc
```diff
@@ -614,7 +614,8 @@ Status CreateTextJobs(
         }
```
kqyang, line 614: Line 605 should be updated too:

```c++
if (hls_listener && !stream.dash_only)
```

popraduvmp: Sorry guys for not responding previously, I was distracted by other things but I started working on this also.
I would suggest adding also at line 610:
```C
if ((stream.segment_template.empty() || !stream.output.empty()) && !stream.dash_only)
```

popraduvmp: Sorry guys for not responding previously, I was distracted by other things but I started working on this also. Thank you @sr1990 for your fix.
I would suggest adding also at line 610:
```C
if ((stream.segment_template.empty() || !stream.output.empty()) && !stream.dash_only)
```

kqyang: @popraduvmp No worries! The code at line 610 is a sub-condition under the `if` at line 605, and `stream.dash_only` is already guaranteed to be false, so the additional check is unnecessary.

popraduvmp: @kqyang You are so right... My bad, still early for me, I just copy pasted my changes quickly and didn't check.

packager/app/test/packager_test.py
```diff
@@ -719,6 +719,21 @@ def testAudioVideoWithTwoTrickPlayDecreasingRate(self):
...
+                                        output_format='mp4',
+                                        segmented=True,
+                                        dash_only=True)
+    hls_text_stream = self._GetStreams(['text'],
+                                       test_files=['bear-english.vtt'],
+                                       segmented=True,
+                                       hls_only=True)
+    streams = audio_video_streams + dash_text_stream + hls_text_stream
+    self.assertPackageSuccess(streams, self._GetFlags(output_dash=True, output_hls=True))
+    self._CheckTestResults('hls-only-dash-only-captions')
```
kqyang, line 735: @sr1990 I am sorry, but we have to replace this one with a different check as packager right now does not generate deterministic mpd:

```python
    # Mpd cannot be validated right now since we don't generate deterministic
    # mpd with multiple inputs due to thread racing.
    # TODO(b/73349711): Generate deterministic mpd or at least validate mpd
    #                   schema.
    self._CheckTestResults(
        'hls-only-dash-only-captions',
        diff_files_policy=DiffFilesPolicy(
            allowed_diff_files=['output.mpd'], exact=False))
```

kqyang:
@sr1990 @popraduvmp Thanks for your help! The change has been merged. Appreciated!

Issue #717, opened 2020-02-26
playback issue of HEVC content with CENC cbcs encryption in AVplayer


jingke2000:
## System info
**Operating System**: macOS High Sierra, V10.13.4
**Shaka Packager Version**: version v2.3.0-5bf8ad5-release
**playback device**: iPad with iOS 13.3.1 

## Issue and steps to reproduce the problem
We have 4K video conent encoded in HEVC packaged in HLS fmp4. The playback from the AVplayer in an iPad with iOS 13.3.1 is fine. However, after CENC cbcs encryption, the playback shows macroblocking issue every 1 or 2seonds. We used Shaka Packager to package or encrypt the 4K content in both cases. Here are the links to the sample conent:
Clear (playback is OK):
https://hls-demo-s3.s3.us-east-2.amazonaws.com/4k_cbcs_testing/3/video.m3u8
Encrypted (playback has macroblocking issue):
https://hls-demo-s3.s3.us-east-2.amazonaws.com/4k_cbcs_testing/4/video.m3u8

**Packager Command**:
"
packager-osx 'in=PEDRO_E_INES_TRAILER_4KUHD_PRORES_04022019_UHD20_L1.mp4 ,stream=video,init_segment=4k_cbcs_shaka_clr_key/video/init.mp4,segment_template=4k_cbcs_shaka_clr_key/video/$Number$.m4s,playlist_name=video.m3u8,iframe_playlist_name=iframe.m3u8' 'in=PEDRO_E_INES_TRAILER_4KUHD_PRORES_04022019_UHD20_L1.mp4,stream=audio,init_segment=4k_cbcs_shaka_clr_key/audio/init.mp4,segment_template=4k_cbcs_shaka_clr_key/audio/$Number$.m4s,playlist_name=audio.m3u8,hls_group_id=audio,hls_name=ENGLISH' --protection_scheme cbcs --enable_raw_key_encryption --keys key_id=55fa9a934d7f3cbebf99b644612598c9:key=03030303030303030303030303030303 --protection_systems FairPlay --iv CC06DA420052BA12925E96EBF98C2F4E --hls_master_playlist_output 4k_cbcs_shaka_clr_key/master.m3u8  --hls_key_uri <key_url>
"

What is the expected result?
Playback of the CENC cbcs encrypted 4K content without visual defects;
What happens instead?
Playback of the CENC cbcs encrypted 4K content shows macroblocking issue every 1~2 seconds;

**Source Content**
Source content can be downloaded here:
[...]

PR #773, opened 2020-05-22
Fixed inaccurate parsing of ref pic list modification (#717)

avrecko:
Should use value from slice header as the default might get overwritten.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@
 #
 # Please keep the list sorted.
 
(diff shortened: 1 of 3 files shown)


avrecko:
@googlebot I signed it!

kqyang:
@avrecko Thanks for the pull request. Would you like to update https://github.com/google/shaka-packager/blob/master/AUTHORS and https://github.com/google/shaka-packager/blob/master/CONTRIBUTORS with your name?

avrecko:
Sure. Done.

Issue #772, opened 2020-05-22
CLion IDE instructions


avrecko:
I'd appreciate if you write how to set up Shaka with CLion IDE in your documentation.

I have no experience with C/C++ build tools. Just a little familiarity with (c)make. I am not a C/C++ developer.

I don't know how to make the CMakeList.txt that CLion wants. It defeats the purpose of having an IDE if code navigation doesn't work.

Your build instructions are very nicely done. I'd really like to see IDE instructions as well. For CMake based projects it is not a problem. But for this I am just scratching my head.

kqyang:
> I'd really like to see IDE instructions as well

That is a good idea, but there are just too many different IDEs people use :). 

I personally don't have experience with CLion. 

As for generating CMakeList.txt, you can try with this command:

```shell
$ GYP_GENERATORS=cmake gclient runhooks
```

It will generate CMakeList.txt in `out/Debug` and `out/Release` directories. Let us know if it works. You are also welcomed to add an IDE instructions page in our documentation: https://github.com/google/shaka-packager/tree/master/docs/source!

avrecko:
The command did generate the CMakeList.txt. But there are some problems:

> 
> CMake Error at CMakeLists.txt:7 (enable_language):
>   No CMAKE_ASM_COMPILER could be found.
> 
>   Tell CMake where to find the compiler by setting either the environment
>   variable "ASM" or the CMake cache entry CMAKE_ASM_COMPILER to the full path
>   to the compiler, or to the compiler name if it is in the PATH.

When I commented out the 

> enable_language(ASM)

I got

> 
> CMake Error at CMakeLists.txt:3811 (add_executable):
>   Cannot find source file:
> 
>     ../../packager/third_party/curl/source/src/tool_mfiles.c
> 
>   Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
>   .hpp .hxx .in .txx

I looked at the source and indeed tool_mfiles.c is missing. There are a bunch of other tool_something.c files but mfiles is not one of them.

avrecko:
What I ended up doing is commenting out curl, protobuf, test lib and putting absolute paths in. Some lines in CMakeLists are referring to macos dir. This is not going to work on linux.

Can you make CMakeLists use absolute paths and don't include macos stuff on linux?

kqyang:
[...]

PR #774, opened 2020-05-24
Ide instructions doc #772

avrecko:
Feel free to change the text as much as you like. I'd just like to see IDE support mentioned.

Alternatively the CMakeLists.txt generation could be mentioned.

### DIFF
--- a/docs/source/build_instructions.md
+++ b/docs/source/build_instructions.md
@@ -338,6 +338,29 @@ resolve the issue:
 $ brew install curl --with-openssl
 ```
 
+### Using an IDE
+
+No specific instructions are available.
+
+You might find Gyp generators helpful. Output is not guaranteed to work.
+Manual editing might be necessary.
+
+To generate CMakeLists.txt in out/Release and out/Debug use:
+
+```shell
+$ GYP_GENERATORS=cmake gclient runhooks
+```
+
+To generate IDE project files in out/Release and out/Debug use:
+
+```shell
+$ GYP_GENERATORS=eclipse gclient runhooks
+$ GYP_GENERATORS=xcode gclient runhooks
+$ GYP_GENERATORS=xcode_test gclient runhooks
+$ GYP_GENERATORS=msvs gclient runhooks
+$ GYP_GENERATORS=msvs_test gclient runhooks
+```
+
 ## Contributing
 
 If you have improvements or fixes, we would love to have your contributions.


docs/source/build_instructions.md
```diff
@@ -338,6 +338,29 @@ resolve the issue:
 $ brew install curl --with-openssl
 ```
 
+### Using an IDE
+
+No specific instructions are available.
+
+You might find Gyp generators helpful. Output is not guaranteed to work.
+Manual editing might be necessary.
```
kqyang, line 346: Do you think if it will be useful to mention what manual editing you have done to make it work in CLion?

avrecko: Sure. I'll think about the docs some more. Maybe better for me to just focus on CLion. I'll revisit this over the weekend.

zdanek: @avrecko any progress in this topic?

avrecko: @zdanek I am preoccupied with regular work. I don't see myself taking the time to revisit this. Maybe you can take a look or @kqyang can either approve or decline this pull request?

I didn't get it to build with the IDE just enough for the code completion to work if I remember correctly. I just ended up removing a bunch of stuff from CMakeList. Hardly good engineering. Not the quality standard I'd target for. Best if a professional C developer makes the instructions.

kqyang: I use VIM with youcompleteme personally, which just works without the need to do extra configuration IIRC. It has auto completion too. 

And sure, I'll merge it as is. I think it is already useful for some people in the current state. 

@zdanek, if you have anything to add, feel free to send a PR to update it.

zdanek: I managed to create  CMakeLists.txt and CLion accepted it. 
I was able to edit code with autocompletition. But it required removing about dozen of cc files from various GYP files as they were absent in source tree. I think it was some kind of mess that arose through the years of coding.
I will try to repeat those steps and check if it does not fail any of the builds / build targets and eventually make an MR.

kqyang: That will be great. Thanks.

zdanek: @kqyang I have fixed gyp files and ```packager/tools/gyp/pylib/gyp/generator/cmake.py``` and now it creates proper CMakeLists.txt that can be used with build tool or IDE based on cmake. CLion is a great example.

Problem is with cmake.py. It's from ```https://chromium.googlesource.com/external/gyp``` I'm not very skilled at cmake. Maybe it can be fixed differently. I will post on GYP tools mailing list regarding this.

There's third concern. ```nocompile.gni``` refers to ```nocompile_driver.py``` script which is not present in source tree.
If not provided, cmake will fail. Eventually it can be created as empty file, because it's not invoked in normal compilation process.

There are some GYP files that should be changed because they point to non existing files in source tree but the GYP files are from other git modules, for example ```https://chromium.googlesource.com/chromium/src/build```
but still the repo moved forward and the GYP file no longer exists in newer source tree.

This is a bit complicated.
It requires to upgrade submodules to versions where GYP files are fixed but not so new as Shaka  will not compile with latest versions of submodules.

What to do?

kqyang: @zdanek, Great to know that you have made it working! Thanks for looking into it.

> Problem is with cmake.py. It's from https://chromium.googlesource.com/external/gyp I'm not very skilled at cmake. Maybe it can be fixed differently. I will post on GYP tools mailing list regarding this.

You'll need to submit a PR to chromium for cmake changes: https://chromium.googlesource.com/chromium/src/+/master/docs/contributing.md.

> nocompile.gni refers to nocompile_driver.py script

We do not use GN right now, so it shouldn't be invoked. Do you know which script invokes nocompile.gni?

> There are some GYP files that should be changed

Do you mean that in order to make cmake working, the GYP files need to be changed too? What are the change OOC? 

And yes, the chromium upstream has deprecated GYP, so we won't be able to change the GYP files. Can it be workarounded?

Issue #753, opened 2020-04-22
Dolby DD+JOC support and using MPEG scheme for audio channel configuration


ShaoWeiguo:
Shaka Packager does not support Dolby DD+JOC, and still Dolby scheme for audio channel configuration descriptor. According to latest DD+ specs ETSI TS 102 366 V1.4.1, should prefer to use MPEG scheme.

I will send the pull request to fix these two issues.

Thanks.

kqyang:
@ShaoWeiguo Great. Thanks for helping out!

PR #775, opened 2020-05-25
dolby/ddpjoc

ShaoWeiguo:
1. Using MPEG scheme to replace Dolby scheme for audio channel configuration descriptor for DD+ in DASH.
2. Support DD+JOC in DASH and HLS.


ShaoWeiguo:
This PR is used to resolve the issue recorded in https://github.com/google/shaka-packager/issues/753.

packager/media/codecs/ec3_audio_util.cc
```diff
@@ -121,8 +189,20 @@ bool ExtractEc3Data(const std::vector<uint8_t>& ec3_data,
...
+  }
+  if (bit_reader.bits_available() < 16) {
+    *ec3_joc_flag = false;
+    *ec3_joc_complexity = 0;
+    return true;
   }
 
+  RCHECK(bit_reader.SkipBits(7));
+  RCHECK(bit_reader.ReadBits(1, ec3_joc_flag));
+  if (*ec3_joc_flag == true) {
```
kqyang, line 203: nit: you can remove "== true"

And a question: do we need two variables? Is it ok to use ec3_joc_complexity==0 to indicate ec3_joc_flag=false?

packager/media/event/muxer_listener_internal.cc
```diff
@@ -122,6 +122,23 @@ void AddAudioInfo(const AudioStreamInfo* audio_stream_info,
     }
     audio_info->mutable_codec_specific_data()->set_ec3_channel_map(
         ec3_channel_map);
+    uint32_t ec3_channel_mpeg_value;
+    if (!CalculateEC3ChannelMPEGValue(codec_config, ec3_channel_mpeg_value)) {
+        LOG(ERROR) << "Failed to calculate EC3 channel configuration \
+          descriptor value with MPEG scheme.";
```
kqyang, line 128: Avoid using line break, which may not be properly rendered, use:
 
       LOG(ERROR) << "Failed to calculate EC3 channel configuration "
                      "descriptor value with MPEG scheme.";

packager/media/codecs/ec3_audio_util.h
```diff
@@ -27,6 +27,19 @@ bool CalculateEC3ChannelMap(const std::vector<uint8_t>& ec3_data,
 ///         success; otherwise 0 is returned.
 size_t GetEc3NumChannels(const std::vector<uint8_t>& ec3_data);
 
+/// Parse data from EC3Specific box, calculate EC3 channel map and then
+/// obtain channel configuration descriptor value with MPEG scheme based on
+/// ETSI TS 102 366 V1.4.1 Digital Audio Compression (AC-3, Enhanced AC-3)
+/// Standard, Table I.1.1.
+bool CalculateEC3ChannelMPEGValue(const std::vector<uint8_t>& ec3_data,
+                                  uint32_t& ec3_channel_mpeg_value);
```
kqyang, line 35: Can you return the output in pointer for consistency, i.e.

bool CalculateEC3ChannelMPEGValue(const std::vector<uint8_t>& ec3_data,
                                  uint32_t* ec3_channel_mpeg_value);

Similar below.

packager/mpd/base/xml/xml_node.cc
```diff
@@ -463,11 +463,36 @@ bool RepresentationXmlNode::AddAudioChannelInfo(const AudioInfo& audio_info) {
     // Convert EC3 channel map into string of hexadecimal digits. Spec: DASH-IF
     // Interoperability Points v3.0 9.2.1.2.
```
kqyang, line 464: Can you include the associated spec for the below change?

packager/media/codecs/ec3_audio_util.cc
ShaoWeiguo: nit: you can remove "== true"

- Yes, I will remove.

And a question: do we need two variables? Is it ok to use ec3_joc_complexity==0 to indicate ec3_joc_flag=false?

- Yes. We need two variables here.  
Comparing with regular DD+ stream, it contains additional 2 (or more) bytes. 
The syntax of the first byte is **reserved bits (6 bits)** + **flag_ec3_extension_type_b (1 bit)** + **flag_ec3_extension_type_a (1 bit)**. 
Only if flag_ec3_extension_type_a == 1, the second byte represent ec-3 joc complexity.
If flag_ec3_extension_type_b == 1, the second byte will have another meaning.
flag_ec3_extension_type_a  and flag_ec3_extension_type_b can't be set to 1 simultaneously.

packager/media/event/muxer_listener_internal.cc
ShaoWeiguo: I will update based on your suggestion.

packager/media/codecs/ec3_audio_util.h
ShaoWeiguo: Sure. I will update CalculateEC3ChannelMPEGValue and GetEc3JocInfo.

packager/media/codecs/ec3_audio_util.cc
ShaoWeiguo: ec3_joc_complexity==0 can be used to indicate indicate ec3_joc_flag=false.
But ec3_joc_complexity !=0 can NOT be used to indicate ec3_joc_flag=ture.

packager/mpd/base/xml/xml_node.cc
ShaoWeiguo: Sure. It comes from ETSI TS 102 366 V1.4.1 Digital Audio Compression (AC-3, Enhanced AC-3), chapter I.1.2.1. I will add the comments.

packager/media/codecs/ec3_audio_util.cc
kqyang: I think there are some misunderstandings. What I mean is that can we keep `ec3_joc_flag` as a local variable and just expose |ec3_joc_complexity| to the caller, i.e.

```c++
  bool ec3_joc_flag;
  RCHECK(bit_reader.ReadBits(1, &ec3_joc_flag));
  if (ec3_joc_flag) {
    ...
  }
```

And in DASH and HLS generation, the code can check whether |ec3_joc_complexity| is 0 first before generating the corresponding DASH/HLS tags.

This works only if 0 is not a valid value for |ec3_joc_complexity|, so we can use it to indicate the presence of the flag.

This allows us to remove |ec3_joc_flag| in the proto. (Actually we can already do that now by using protobuf's has_xxx function to determine the presence of the flag)

packager/media/event/muxer_listener_internal.cc
```diff
@@ -122,6 +122,23 @@ void AddAudioInfo(const AudioStreamInfo* audio_stream_info,
     }
     audio_info->mutable_codec_specific_data()->set_ec3_channel_map(
```
kqyang, line 123: Now that we have more usages of mutable_codec_specific_data, we can create an intermediate variable to make the code more readable, i.e.

```c++
auto* codec_data = audio_info->mutable_codec_specific_data();
codec_data->set_ec3_channel_map(ec3_channel_map);
...
codec_data->set_ec3_channel_mpeg_value(ec3_channel_mpeg_value);
...
```

packager/hls/base/master_playlist_unittest.cc
```diff
@@ -246,12 +251,12 @@ TEST_F(MasterPlaylistTest, WriteMasterPlaylistVideoAndAudio) {
   // First audio, english.m3u8.
   std::unique_ptr<MockMediaPlaylist> english_playlist = CreateAudioPlaylist(
       "eng.m3u8", "english", "audiogroup", "audiocodec", "en", kAudio1Channels,
-      kAudio1MaxBitrate, kAudio1AvgBitrate);
+      kAudio1MaxBitrate, kAudio1AvgBitrate, false, 0);
```
kqyang, line 254: Can we define constants in line 35 to make the code more readable?

```c++
const kEC3JocFlagEnabled = true;
const kEC3JocComplexityZero = 0
const kEC3JocComplexity = 16;
```

Then we can use: 

```c++
kAudio1MaxBitrate, kAudio1AvgBitrate, !kEC3JocFlagEnabled, kEC3JocComplexityZero);
```

packager/hls/base/master_playlist.cc
```diff
@@ -308,10 +308,15 @@ void BuildMediaTag(const MediaPlaylist& playlist,
   const MediaPlaylist::MediaPlaylistStreamType kAudio =
       MediaPlaylist::MediaPlaylistStreamType::kAudio;
   if (playlist.stream_type() == kAudio) {
-    std::string channel_string = std::to_string(playlist.GetNumChannels());
-    tag.AddQuotedString("CHANNELS", channel_string);
+    if (playlist.GetEC3JocFlag()) {
```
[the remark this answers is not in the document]
ShaoWeiguo: Sure. I will.

packager/hls/base/master_playlist_unittest.cc
ShaoWeiguo: Good suggestion. I will update it.

packager/mpd/base/xml/xml_node.cc
```diff
@@ -463,11 +463,42 @@ bool RepresentationXmlNode::AddAudioChannelInfo(const AudioInfo& audio_info) {
...
     const uint16_t ec3_channel_map =
-        base::HostToNet16(audio_info.codec_specific_data().ec3_channel_map());
```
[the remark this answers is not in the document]
ShaoWeiguo: Good suggestion. I will update it.

packager/media/event/muxer_listener_internal.cc
ShaoWeiguo: Good suggestion. I will update it.

packager/media/codecs/ec3_audio_util.cc
ShaoWeiguo: Let me have a discussion with team firstly, and then give you feedback late.

ShaoWeiguo: We can remove the ec3_joc_flag. I have updated the PR following the suggestion. Please have a review.

packager/mpd/base/xml/xml_node.cc
```diff
@@ -460,14 +460,43 @@ bool RepresentationXmlNode::AddAudioChannelInfo(const AudioInfo& audio_info) {
...
+    if (ec3_channel_mpeg_value == NO_MAPPING) {
+      audio_channel_config_value =
         base::HexEncode(&ec3_channel_map, sizeof(ec3_channel_map));
-    audio_channel_config_scheme =
+      audio_channel_config_scheme =
         "tag:dolby.com,2014:dash:audio_channel_configuration:2011";
+    } else {
+      audio_channel_config_value = base::UintToString(ec3_channel_mpeg_value);
+      audio_channel_config_scheme = "urn:mpeg:mpegB:cicp:ChannelConfiguration";
+    }
```
kqyang, line 481: In the latest DASH-IF-IOP: https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf

Section 3.9.4.6. It says:

> If the codec is anyone in Table 22, Table 27, Table 28 or Table 29, then any of the following may be used
> - urn:mpeg: dash:23003:3:audio_channel_configuration:2011 as defined in ISO/IEC 23009-1 [1], 5.8.5.4
> - urn:mpeg:mpegB:cicp:ChannelConfiguration as defined in ISO/IEC 23001-8 [49]
>
> If the codec is ec-3 or ac-4 according to Table 25, then the following shall be used
> - tag:dolby.com,2014: dash:audio_channel_configuration:2011" as defined at http://dashif.org/identifiers/audio-source-data/ (see section 9.2.1.2)

Table 25 is for Dolby EC-3, mlpa and AC-4
Table 22,27,28 is for MPEG-4 AAC; table 29 is for MPEG-H audio.

So per DASH-IF-IOP, we should be using "tag:dolby.com,2014: dash:audio_channel_configuration:2011" instead of "urn:mpeg:mpegB:cicp:ChannelConfiguration" for EC-3. Is there any change in Dolby's internal guidance on this?

packager/hls/base/master_playlist.cc
```diff
@@ -308,10 +308,25 @@ void BuildMediaTag(const MediaPlaylist& playlist,
...
+    // According to HLS spec: https://tools.ietf.org/html/
+    // draft-pantos-hls-rfc8216bis 4.4.6.1. CHANNELS is a quoted-string that
+    // specifies an ordered, slash-separated ("/") list of parameters.
+    // The first parameter is a count of audio channels, and the second
+    // parameter identifies the encoding of object-based audio used by the
+    // Rendition. HLS Authoring Specification for Apple Devices Appendices
+    // documents how to handle Dolby Digital Plus JOC content.
+    // https://developer.apple.com/documentation/http_live_streaming/
+    // hls_authoring_specification_for_apple_devices/
+    // hls_authoring_specification_for_apple_devices_appendices
```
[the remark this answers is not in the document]
ShaoWeiguo: This is the full line, _https://developer.apple.com/documentation/http_live_streaming/hls_authoring_specification_for_apple_devices/hls_authoring_specification_for_apple_devices_appendices_. The total character is 163. Not very clear how to make it to the same line if meeting the line length limitation (<=80). Do you have any suggestion?
BTW, for line 311, I also break the lines for the URL. The full URL shall be https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis. The total character is less than 80, so I can move it to same line.

packager/media/codecs/ec3_audio_util.cc
```diff
@@ -121,8 +188,20 @@ bool ExtractEc3Data(const std::vector<uint8_t>& ec3_data,
...
+    *ec3_joc_complexity = 0;
+    return true;
   }
 
+  RCHECK(bit_reader.SkipBits(7));
+  bool ec3_joc_flag;
+  RCHECK(bit_reader.ReadBits(1, &ec3_joc_flag));
+  if (ec3_joc_flag) {
+    RCHECK(bit_reader.ReadBits(8, ec3_joc_complexity));
+  }
```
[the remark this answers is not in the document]
ShaoWeiguo: You are right, thanks. Fixed it.

packager/media/codecs/ec3_audio_util_unittest.cc
```diff
@@ -45,5 +45,20 @@ TEST(EC3AudioUtilTest, ChannelTest3) {
...
+  const std::vector<uint8_t> ec3_data = {0x14, 0x00, 0x20, 0x0f, 0x00, 0x01,
+                                         0x10};
+
+  uint32_t channel_map;
+  uint32_t ec3_joc_complexity;
+  EXPECT_TRUE(CalculateEC3ChannelMap(ec3_data, &channel_map));
+  EXPECT_EQ(0xF801u, channel_map);
+  EXPECT_EQ(6u, GetEc3NumChannels(ec3_data));
+  EXPECT_TRUE(GetEc3JocComplexity(ec3_data, &ec3_joc_complexity));
+  EXPECT_EQ(16u, ec3_joc_complexity);
```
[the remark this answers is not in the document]
ShaoWeiguo: Sure.

packager/hls/base/master_playlist.cc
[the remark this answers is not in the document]
ShaoWeiguo: I search all the URLs in the projects, find some URLs have more than 80 character in one line. So I modify it, like below.
   // According to HLS spec:
    // https://tools.ietf.org/html/draft-pantos-hls-rfc8216bis 4.4.6.1.
    // CHANNELS is a quoted-string that specifies an ordered,
    // slash-separated ("/") list of parameters. The first parameter is a count
    // of audio channels, and the second parameter identifies the encoding of
    // object-based audio used by the Rendition. HLS Authoring Specification
    // for Apple Devices Appendices documents how to handle Dolby Digital Plus
    // JOC content.
    // https://developer.apple.com/documentation/http_live_streaming/hls_authoring_specification_for_apple_devices/hls_authoring_specification_for_apple_devices_appendices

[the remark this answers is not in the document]
kqyang: Yes, for URLs, it is fine to be more than 80 characters. (Another option is to use shorten url service like bitly.)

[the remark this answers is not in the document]
ShaoWeiguo: I see, thanks.

packager/mpd/base/xml/xml_node.cc
ShaoWeiguo: Yes. Dolby has changed it. Not only for internal guidance, but also try to fix in DASH-IF IOP (still under review). https://github.com/Dash-Industry-Forum/DASH-IF-IOP/issues/268

packager/media/codecs/ec3_audio_util.cc
```diff
@@ -81,10 +81,77 @@ uint8_t ReverseBits8(uint8_t n) {
...
+         kLFEScreen | kLvhRvhPair | kCenterVerticalHeight | kLtsRtsPair |
+         kTopCenterSurround:
+      ret = 17;
+      break;
+    case kCenter | kLeft | kRight | kLsdRsdPair | kLrsRrsPair | kLFEScreen |
+         kLvhRvhPair | kLtsRtsPair:
+      ret = 19;
+      break;
+    default:
+      ret = 0xFFFFFFFF;
```
kqyang, line 145: Should we consider it as an error or will it happen in practice that a channel map does not have a MPEG value?

packager/mpd/base/xml/xml_node.cc
```diff
@@ -460,14 +460,43 @@ bool RepresentationXmlNode::AddAudioChannelInfo(const AudioInfo& audio_info) {
   std::string audio_channel_config_value;
 
   if (audio_info.codec() == kEC3Codec) {
+    const auto& codec_data = audio_info.codec_specific_data();
     // Convert EC3 channel map into string of hexadecimal digits. Spec: DASH-IF
     // Interoperability Points v3.0 9.2.1.2.
```
kqyang, line 465: This is no longer accurate. Can you replace it with the pull request instead as it has not been adopted by DASH IF yet.

packager/media/codecs/ec3_audio_util.cc
ShaoWeiguo: All the “normal” ones can be converted. However, if find one that cannot be converted, I don’t think the spec says it is invalid, just unusual. In such situation, Dolby's channel scheme shall be used.

packager/mpd/base/xml/xml_node.cc
ShaoWeiguo: I am not sure whether I catch your point. The comment is not correct, but converting EC3 channel map into string of hexadecimal digits is necessary.  We need to handle the cases can't be converted.

We shall change the comments into below
    // Convert EC3 channel map into string of hexadecimal digits. Spec: ETSI TS
    // 102 366 V1.4.1 Digital Audio Compression(AC-3, Enhanced AC-3) Standard
    // E.1.3.1.8.

Does it meet your expectation?

kqyang: Sorry for not making it clear. What I mean is updating the comment to reference the latest spec instead of the outdated spec which does not match with the implementation below.

For example, 

```c++
    // Use MPEG scheme if the mpeg value is available and valid, fallback to 
    // EC3 channel mapping otherwise.
    // See https://github.com/Dash-Industry-Forum/DASH-IF-IOP/issues/268.
    const uint32_t ec3_channel_mpeg_value = codec_data.ec3_channel_mpeg_value();
    const uint32_t NO_MAPPING = 0xFFFFFFFF;
    if (ec3_channel_mpeg_value == NO_MAPPING) {
      // Convert EC3 channel map into string of hexadecimal digits. Spec: DASH-IF
      // Interoperability Points v3.0 9.2.1.2.
      const uint16_t ec3_channel_map = 
          base::HostToNet16(codec_data.ec3_channel_map());
      audio_channel_config_value =
          base::HexEncode(&ec3_channel_map, sizeof(ec3_channel_map));
      audio_channel_config_scheme =
          "tag:dolby.com,2014:dash:audio_channel_configuration:2011";
    } else {
      // Calculate EC3 channel configuration descriptor value with MPEG scheme.
      // Spec: ETSI TS 102 366 V1.4.1 Digital Audio Compression
      // (AC-3, Enhanced AC-3) I.1.2.
      audio_channel_config_value = base::UintToString(ec3_channel_mpeg_value);
      audio_channel_config_scheme = "urn:mpeg:mpegB:cicp:ChannelConfiguration";
    }
    ```

kqyang:
@ShaoWeiguo Thanks for the great work! The change has been merged.

Issue #779, opened 2020-06-01
[Feature Request] Support for mpeg1 audio


zevlg:
Currently it is not possible to use inputs with mpeg1 audio streams inside container and to segment it into `.m3u8` playlist containing `.mp3` files

Partial support for mpeg1 audio is done in https://github.com/google/shaka-packager/pull/778

However this pull request implements support for mpeg1 only in mp2t container.  Support for other containers is yet to be done

mpeg1 audio format is described at https://www.datavoyage.com/mpgscript/mpeghdr.htm

PR #780, opened 2020-06-05
Mpeg1 Audio: Support for iso-bmff output, ref #779

zevlg:
- Makes possible to pack mpeg1 audio streams into .mp4 files

### DIFF
--- a/packager/media/codecs/es_descriptor.cc
+++ b/packager/media/codecs/es_descriptor.cc
@@ -168,7 +168,7 @@ void DecoderConfigDescriptor::WriteInternal(BufferWriter* writer) {
   writer->AppendInt(max_bitrate_);
   writer->AppendInt(avg_bitrate_);
 
-  if (!decoder_specific_info_descriptor_is_omitted_)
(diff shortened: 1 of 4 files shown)


packager/media/codecs/es_descriptor.cc
```diff
@@ -167,13 +167,17 @@ void DecoderConfigDescriptor::WriteInternal(BufferWriter* writer) {
   writer->AppendNBytes(buffer_size_db_, 3);
   writer->AppendInt(max_bitrate_);
   writer->AppendInt(avg_bitrate_);
-  decoder_specific_info_descriptor_.Write(writer);
+
+  if (!decoder_specific_info_descriptor_is_ommited_)
```
kqyang: It does not seem that the flag is necessary. What about updating the descriptor to skip if it is empty instead, i.e.

```c++
if (!decoder_specific_info_descriptor_.data().empty())
  decoder_specific_info_descriptor_.Write(writer);

...

  if (decoder_specific_info_descriptor_.data().empty())
    return data_size_without_children;
```

PR #790, opened 2020-06-22
Write to buffer before writing to file for TS segments generation.

sr1990:
PR 1 for Omit Segment Timeline Issue.

### DIFF
--- a/packager/media/formats/mp2t/ts_segmenter.cc
+++ b/packager/media/formats/mp2t/ts_segmenter.cc
@@ -102,7 +102,7 @@ Status TsSegmenter::AddSample(const MediaSample& sample) {
   if (sample.is_encrypted())
     ts_writer_->SignalEncrypted();
 
-  if (!ts_writer_file_opened_ && !sample.is_key_frame())
+  if (!ts_writer_buffer_initialized_ && !sample.is_key_frame())
     LOG(WARNING) << "A segment will start with a non key frame.";
 
   if (!pes_packet_generator_->PushSample(sample)) {
@@ -122,19 +122,16 @@ void TsSegmenter::InjectPesPacketGeneratorForTesting(
 }
 
 void TsSegmenter::SetTsWriterFileOpenedForTesting(bool value) {
-  ts_writer_file_opened_ = value;
+  ts_writer_buffer_initialized_ = value;
 }
 
 Status TsSegmenter::OpenNewSegmentIfClosed(int64_t next_pts) {
-  if (ts_writer_file_opened_)
+  if (ts_writer_buffer_initialized_)
     return Status::OK;
-  const std::string segment_name =
-      GetSegmentName(muxer_options_.segment_template, next_pts,
-                     segment_number_++, muxer_options_.bandwidth);
-  if (!ts_writer_->NewSegment(segment_name))
+  SetSegmentStartTimestamp(next_pts);
+  if (!ts_writer_->NewSegment())
     return Status(error::MUXER_FAILURE, "Failed to initilize TsPacketWriter.");
-  current_segment_path_ = segment_name;
-  ts_writer_file_opened_ = true;
+  ts_writer_buffer_initialized_ = true;
   return Status::OK;
 }
 
@@ -148,13 +145,13 @@ Status TsSegmenter::WritePesPacketsToFile() {
       return status;
 
     if (listener_ && IsVideoCodec(codec_) && pes_packet->is_key_frame()) {
-      base::Optional<uint64_t> start_pos = ts_writer_->GetFilePosition();
+      base::Optional<uint64_t> start_pos = ts_writer_->GetPosition();
 
       const int64_t timestamp = pes_packet->pts();
       if (!ts_writer_->AddPesPacket(std::move(pes_packet)))
         return Status(error::MUXER_FAILURE, "Failed to add PES packet.");
 
-      base::Optional<uint64_t> end_pos = ts_writer_->GetFilePosition();
+      base::Optional<uint64_t> end_pos = ts_writer_->GetPosition();
       if (!start_pos || !end_pos) {
         return Status(error::MUXER_FAILURE,
                       "Failed to get file position in WritePesPacketsToFile.");
@@ -171,30 +168,37 @@ Status TsSegmenter::WritePesPacketsToFile() {
 Status TsSegmenter::FinalizeSegment(uint64_t start_timestamp,
                                     uint64_t duration) {
   if (!pes_packet_generator_->Flush()) {
-    return Status(error::MUXER_FAILURE,
-                  "Failed to flush PesPacketGenerator.");
+    return Status(error::MUXER_FAILURE, "Failed to flush PesPacketGenerator.");
   }
   Status status = WritePesPacketsToFile();
   if (!status.ok())
     return status;
 
-  // This method may be called from Finalize() so ts_writer_file_opened_ could
+  // This method may be called from Finalize() so ts_writer_buffer_initialized_ could
   // be false.
-  if (ts_writer_file_opened_) {
+  if (ts_writer_buffer_initialized_) {
+    std::string segment_path =
+        GetSegmentName(muxer_options_.segment_template, GetSegmentStartTimestamp(),
+                       segment_number_++, muxer_options_.bandwidth);
+
+    if (!ts_writer_->WriteFile(segment_path)) {
+      LOG(ERROR) << "Failed to create new ts segment file.";
+    }
+
     if (!ts_writer_->FinalizeSegment()) {
       return Status(error::MUXER_FAILURE, "Failed to finalize TsWriter.");
     }
+    
     if (listener_) {
       const int64_t file_size =
-          File::GetFileSize(current_segment_path_.c_str());
-      listener_->OnNewSegment(current_segment_path_,
+          File::GetFileSize(segment_path.c_str());
+      listener_->OnNewSegment(segment_path,
                               start_timestamp * timescale_scale_ +
                                   transport_stream_timestamp_offset_,
                               duration * timescale_scale_, file_size);
     }
-    ts_writer_file_opened_ = false;
+    ts_writer_buffer_initialized_ = false;
   }
-  current_segment_path_.clear();
   return Status::OK;
 }
 
--- a/packager/media/formats/mp2t/ts_segmenter.cc
+++ b/packager/media/formats/mp2t/ts_segmenter.cc
@@ -102,7 +102,7 @@ Status TsSegmenter::AddSample(const MediaSample& sample) {
   if (sample.is_encrypted())
     ts_writer_->SignalEncrypted();
 
-  if (!ts_writer_buffer_initialized_ && !sample.is_key_frame())
+  if (!start_of_new_segment_ && !sample.is_key_frame())
     LOG(WARNING) << "A segment will start with a non key frame.";
 
   if (!pes_packet_generator_->PushSample(sample)) {
@@ -121,17 +121,17 @@ void TsSegmenter::InjectPesPacketGeneratorForTesting(
   pes_packet_generator_ = std::move(generator);
 }
 
-void TsSegmenter::SetTsWriterFileOpenedForTesting(bool value) {
-  ts_writer_buffer_initialized_ = value;
+void TsSegmenter::SetSegmentStartedForTesting(bool value) {
+  start_of_new_segment_ = value;
 }
 
-Status TsSegmenter::OpenNewSegmentIfClosed(int64_t next_pts) {
-  if (ts_writer_buffer_initialized_)
+Status TsSegmenter::StartSegmentIfNeeded(int64_t next_pts) {
+  if (start_of_new_segment_)
     return Status::OK;
-  SetSegmentStartTimestamp(next_pts);
-  if (!ts_writer_->NewSegment())
+  segment_start_timestamp_ = next_pts;
+  if (!ts_writer_->NewSegment(&segment_buffer_))
     return Status(error::MUXER_FAILURE, "Failed to initilize TsPacketWriter.");
-  ts_writer_buffer_initialized_ = true;
+  start_of_new_segment_ = true;
   return Status::OK;
 }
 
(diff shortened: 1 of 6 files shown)


packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -102,7 +102,7 @@ Status TsSegmenter::AddSample(const MediaSample& sample) {
   if (sample.is_encrypted())
     ts_writer_->SignalEncrypted();
 
-  if (!ts_writer_file_opened_ && !sample.is_key_frame())
+  if (!ts_writer_buffer_initialized_ && !sample.is_key_frame())
```
kqyang, line 105: Thought about it again. Can we rename it to start_of_new_segment_, which is more appropriate?

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -122,19 +122,16 @@ void TsSegmenter::InjectPesPacketGeneratorForTesting(
 }
 
 void TsSegmenter::SetTsWriterFileOpenedForTesting(bool value) {
-  ts_writer_file_opened_ = value;
+  ts_writer_buffer_initialized_ = value;
 }
 
 Status TsSegmenter::OpenNewSegmentIfClosed(int64_t next_pts) {
```
kqyang, line 128: Can we call it StartSegmentIfNeeded?

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -122,19 +122,16 @@ void TsSegmenter::InjectPesPacketGeneratorForTesting(
 }
 
 void TsSegmenter::SetTsWriterFileOpenedForTesting(bool value) {
```
kqyang, line 124: Can we call it SetSegmentStartedForTesting?

packager/media/formats/mp2t/ts_segmenter.h
```diff
@@ -93,16 +97,13 @@ class TsSegmenter {
...
+  // TsWriter::FinalizeSegment() succeeds.
+  bool ts_writer_buffer_initialized_ = false;
   std::unique_ptr<PesPacketGenerator> pes_packet_generator_;
 
-  // For OnNewSegment().
-  // Path of the current segment so that File::GetFileSize() can be used after
-  // the segment has been finalized.
-  std::string current_segment_path_;
-
+  int segment_start_timestamp_ = -1;
```
kqyang, line 106: Use int64_t

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -148,13 +145,13 @@ Status TsSegmenter::WritePesPacketsToFile() {
       return status;
```
kqyang, line 145: Can we rename the function from WritePesPacketsToFile to WritePesPackets?

packager/media/formats/mp2t/ts_writer.cc
```diff
@@ -163,7 +165,7 @@ TsWriter::TsWriter(std::unique_ptr<ProgramMapTableWriter> pmt_writer)
 
 TsWriter::~TsWriter() {}
 
-bool TsWriter::NewSegment(const std::string& file_name) {
+bool TsWriter::WriteFile(const std::string& file_name) {
```
kqyang, line 168: Prefer calling it WriteToFile

packager/media/formats/mp2t/ts_segmenter.h
```diff
@@ -69,6 +69,10 @@ class TsSegmenter {
 
   /// Only for testing.
   void SetTsWriterFileOpenedForTesting(bool value);
+  
+  /// @param pts value of segment_start_timestamp_.
+  void SetSegmentStartTimestamp(int pts) { segment_start_timestamp_ = pts;};
+  int GetSegmentStartTimestamp() {return segment_start_timestamp_; };
```
kqyang, line 75: Two questions:

(1) Does the two functions need to be public?
(2) If not, do we need the two functions, i.e. we could set and get |segment_start_timestamp_| in the code directly.

packager/media/formats/mp2t/ts_writer.h
```diff
@@ -67,6 +72,7 @@ class TsWriter {
   std::unique_ptr<ProgramMapTableWriter> pmt_writer_;
 
   std::unique_ptr<File, FileCloser> current_file_;
+  std::unique_ptr<BufferWriter> current_buffer_;
```
kqyang, line 75: It does not need to be a unique pointer; also, we can call it |segment_buffer_| i.e. It can be 

```c++
BufferWriter segment_buffer_;
```

I think it will be cleaner to manage the buffer in `TsSegmenter` class. We will need to update the signature of the two functions `NewSegment` and `AddPesPacket` to something like though:

```c++
virtual bool NewSegment(BufferWriter* buffer);
virtual bool AddPesPacket(std::unique_ptr<PesPacket> pes_packet, BufferWriter* buffer);
```

Then we don't need to have the the function `GetPosition` and `WriteFile` in this class.

What do you think?

packager/media/formats/mp2t/ts_writer_unittest.cc
```diff
@@ -144,7 +144,9 @@ TEST_F(TsWriterTest, ClearH264Psi) {
   EXPECT_CALL(*mock_pmt_writer, ClearSegmentPmt(_)).WillOnce(WriteOnePmt());
 
   TsWriter ts_writer(std::move(mock_pmt_writer));
-  EXPECT_TRUE(ts_writer.NewSegment(test_file_name_));
+  EXPECT_TRUE(ts_writer.NewSegment());
+  EXPECT_TRUE(ts_writer.WriteFile(test_file_name_));
```
kqyang, line 148: `test_file_name_` and `SetUp` and `TearDown` can be removed too.

packager/media/formats/mp2t/ts_writer.h
sr1990: Makes sense. I will make this change.

packager/media/formats/mp2t/ts_segmenter.h
sr1990: Yes we can. Will remove the extra functions.

sr1990:
@kqyang have updated the PR as per our discussion above. Please review and let me know.

packager/media/formats/mp2t/ts_segmenter.cc
kqyang: Ping.

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -143,25 +140,28 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
-    Status status = OpenNewSegmentIfClosed(pes_packet->pts());
+    Status status = StartSegmentIfNeeded(pes_packet->pts());
     if (!status.ok())
       return status;
 
     if (listener_ && IsVideoCodec(codec_) && pes_packet->is_key_frame()) {
-      base::Optional<uint64_t> start_pos = ts_writer_->GetFilePosition();
-
+      base::Optional<uint64_t> start_pos = 
+	      (segment_buffer_.Size()!=0)?base::make_optional(segment_buffer_.Size()) : base::nullopt;
```
kqyang, line 149: It does not fail, so we can just use:

```c++
uint64_t start_pos = segment_buffer_.Size();
```

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -143,25 +140,28 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
+      
       const int64_t timestamp = pes_packet->pts();
-      if (!ts_writer_->AddPesPacket(std::move(pes_packet)))
+      if (!ts_writer_->AddPesPacket(std::move(pes_packet), &segment_buffer_))
         return Status(error::MUXER_FAILURE, "Failed to add PES packet.");
 
-      base::Optional<uint64_t> end_pos = ts_writer_->GetFilePosition();
+      base::Optional<uint64_t> end_pos = 
+	      (segment_buffer_.Size()!=0)?base::make_optional(segment_buffer_.Size()) : base::nullopt;
+
```
kqyang, line 157: ditto:

```c++
uint64_t end_pos = segment_buffer_.Size();
```

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -143,25 +140,28 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
       if (!start_pos || !end_pos) {
         return Status(error::MUXER_FAILURE,
                       "Failed to get file position in WritePesPacketsToFile.");
       }
```
kqyang, line 161: The four lines can be removed as the buffer operation does not fail.

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -171,30 +171,46 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
   if (!status.ok())
     return status;
 
-  // This method may be called from Finalize() so ts_writer_file_opened_ could
+  // This method may be called from Finalize() so start_of_new_segment_ could
   // be false.
-  if (ts_writer_file_opened_) {
-    if (!ts_writer_->FinalizeSegment()) {
-      return Status(error::MUXER_FAILURE, "Failed to finalize TsWriter.");
+  if (start_of_new_segment_) {
```
kqyang, line 182: Actually can we call the variable: `segment_started_`, which seems better than `start_of_new_segment_`. Sorry for changing the name multiple times.

Also, I think we can improve the readability here by return early if it is not the case:

```c++
  // This method may be called from Finalize() so start_of_new_segment_ could
  // be false.
  if (!segment_started_)
    return Status::OK;

  std::string segment_path ...
```

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -171,30 +171,46 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
+    std::string segment_path =
+        GetSegmentName(muxer_options_.segment_template, segment_start_timestamp_,
+                       segment_number_++, muxer_options_.bandwidth);
+
+  if ( segment_buffer_.Size() != 0) {
+
```
kqyang, line 188: It is not possible to be zero, so this check is not necessary.

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -171,30 +171,46 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
+    std::unique_ptr<File, FileCloser> segment_file;	  
+    segment_file.reset(File::Open(segment_path.c_str(), "w"));
+    if (!segment_file) {
+      LOG(ERROR) << "Failed to open file " << segment_path;
     }
+  
+  segment_buffer_.WriteToFile(segment_file.get()).ok();
```
kqyang, line 195: RETURN_IF_ERROR(segment_buffer_.WriteToFile(segment_file.get()));

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -171,30 +171,46 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
+    std::string segment_path =
+        GetSegmentName(muxer_options_.segment_template, segment_start_timestamp_,
+                       segment_number_++, muxer_options_.bandwidth);
+
+  if ( segment_buffer_.Size() != 0) {
+
+    std::unique_ptr<File, FileCloser> segment_file;	  
+    segment_file.reset(File::Open(segment_path.c_str(), "w"));
+    if (!segment_file) {
+      LOG(ERROR) << "Failed to open file " << segment_path;
```
kqyang, line 192: return Status(error::FILE_FAILURE,
                    "Cannot open file for write " + segment_path);

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -171,30 +171,46 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
+
+  if (!segment_file.release()->Close()) {
+    return Status(error::MUXER_FAILURE, "Failed to finalize TsWriter.");
+  }
+
     if (listener_) {
       const int64_t file_size =
-          File::GetFileSize(current_segment_path_.c_str());
-      listener_->OnNewSegment(current_segment_path_,
+          File::GetFileSize(segment_path.c_str());
```
kqyang, line 203: We can get the size using `segment_buffer_.Size()`, which avoids an unnecessary file operation. Note that it needs to be executed before `WriteToFile`.

packager/media/formats/mp2t/ts_segmenter.cc
```diff
@@ -171,30 +171,46 @@ Status TsSegmenter::WritePesPacketsToFile() {
...
+    std::unique_ptr<File, FileCloser> segment_file;	  
+    segment_file.reset(File::Open(segment_path.c_str(), "w"));
+    if (!segment_file) {
+      LOG(ERROR) << "Failed to open file " << segment_path;
     }
+  
+  segment_buffer_.WriteToFile(segment_file.get()).ok();
+
+  if (!segment_file.release()->Close()) {
+    return Status(error::MUXER_FAILURE, "Failed to finalize TsWriter.");
```
kqyang, line 198: return Status(
        error::FILE_FAILURE,
        "Cannot close file " + segment_path +
            ", possibly file permission issue or running out of disk space.");

packager/media/formats/mp2t/ts_segmenter.cc
sr1990: Sorry I missed this one. Have renamed the function in the updated PR.

sr1990:
@kqyang thanks for the review. Have updated the PR as per your comments.

packager/media/formats/mp2t/ts_segmenter_unittest.cc
```diff
@@ -319,14 +312,13 @@ TEST_F(TsSegmenterTest, FinalizeSegment) {
...
       .WillOnce(Return(0u));
-  EXPECT_CALL(*mock_ts_writer_, FinalizeSegment()).WillOnce(Return(true));
 
   segmenter.InjectPesPacketGeneratorForTesting(
       std::move(mock_pes_packet_generator_));
   EXPECT_OK(segmenter.Initialize(*stream_info));
   segmenter.InjectTsWriterForTesting(std::move(mock_ts_writer_));
-  segmenter.SetTsWriterFileOpenedForTesting(true);
-  EXPECT_OK(segmenter.FinalizeSegment(0, 100 /* arbitrary duration */));
+  segmenter.SetSegmentStartedForTesting(false);
```
kqyang, line 320: ditto. See the other comments.

Can you add a MuxerListener expection on `OnNewSegment` too to check the segment path is correct?

packager/media/formats/mp2t/ts_segmenter_unittest.cc
```diff
@@ -268,6 +260,7 @@ TEST_F(TsSegmenterTest, PassedSegmentDuration) {
   EXPECT_OK(segmenter.Initialize(*stream_info));
   segmenter.InjectTsWriterForTesting(std::move(mock_ts_writer_));
   EXPECT_OK(segmenter.AddSample(*sample1));
+  segmenter.SetSegmentStartedForTesting(false);
```
kqyang, line 263: What is the reason for `SetSegmentStartedForTesting`? To skip the file operation?

A better way to do it is to use a working file path. I suggest using a memory filesystem. You can do that by changing `segment template` to:

```c++
options.segment_template = "memory://file$Number$.ts";
```

packager/media/formats/mp2t/ts_segmenter_unittest.cc
```diff
@@ -405,7 +396,7 @@ TEST_F(TsSegmenterTest, EncryptedSample) {
   EXPECT_OK(segmenter.Initialize(*stream_info));
   segmenter.InjectTsWriterForTesting(std::move(mock_ts_writer_));
   EXPECT_OK(segmenter.AddSample(*sample1));
-
+  segmenter.SetSegmentStartedForTesting(false);
```
kqyang, line 399: ditto. See the other comments.

packager/media/formats/mp2t/ts_segmenter_unittest.cc
sr1990: As we are using a mock_ts_writer_ (which was previously writing to file using WriteFile), the buffer is empty. So we get a DCHECK failure at line ts_segmenter.cc:191 (`RETURN_IF_ERROR(segment_buffer_.WriteToFile(segment_file.get()));`). 
By calling `segmenter.SetSegmentStartedForTesting(false)`, we are returning from line ts_segmenter.cc:178 (` if (!segment_started_)  return Status::OK;` ) and line 201 (`listener_->OnNewSegment...`) is not executed.

One way is to create a writeToFile function in ts_writer and mock it. Then I could add the expectation  for muxer_listener::OnNewSegment. 
Let me know what you think.

kqyang: Ah, I see. Another way to address it is to update ```AddPesPacket``` override in line 93:

```c++
  // Similar to the hack above but takes a std::unique_ptr.
  MOCK_METHOD1(AddPesPacketMock, bool(PesPacket* pes_packet));
  bool AddPesPacket(std::unique_ptr<PesPacket> pes_packet,
		  BufferWriter* buffer_writer) override {
    // Push dummy data to |buffer_writer| to keep WriteToFile happy.
    buffer_writer->AppendString("dummy PES packet data");
    // No need to keep the pes packet around for the current tests.
    return AddPesPacketMock(pes_packet.get());
  }
```

sr1990: Have made the changes as per your above comment and updated the pr. Please take a look and let me know.

kqyang:
There are some clang formatting issues, but we can address them in a separate commit.

@sr1990 I have merged the change. Thanks for helping refactor TS segment generation to write to buffer first. Appreciated.

PR #791, opened 2020-06-22
Write to buffer before writing to file for Webm segments generation.

sr1990:
PR 2 for Omit Segment Timeline issue.


packager/media/formats/webm/mkv_writer.cc
```diff
@@ -13,8 +13,18 @@ MkvWriter::MkvWriter() : position_(0) {}
```
kqyang, line 13: It will be cleaner to implement another buffer based MkvWriter. We may call it `BufferMkvWriter`. We can then switch multi_segment_segmenter to use the new MkvWriter. Optionally we can switch `two_pass_single_segment_segmenter` to use the new MkvWriter too.

packager/media/formats/webm/mkv_writer.cc
```diff
@@ -13,8 +13,18 @@ MkvWriter::MkvWriter() : position_(0) {}
 
 MkvWriter::~MkvWriter() {}
 
+Status MkvWriter::OpenBuffer() {
+  DCHECK(!file_);
+  current_buffer_  = std::unique_ptr<BufferWriter>(new BufferWriter());
```
kqyang, line 18: As mentioned in TS PR, this does not need to be a unique pointer.

packager/media/formats/webm/mkv_writer.cc
sr1990: Makes sense. Let me try it out.

sr1990:
@kqyang I have created the new class as you suggested. Please review and let me know. Thanks.

sr1990:
What is the correct way to add a new class? I had modified build.ninja file locally to test the changes.

kqyang:
> What is the correct way to add a new class? I had modified build.ninja file locally to test the changes.

Just add the files to webm.gyp: https://github.com/google/shaka-packager/blob/master/packager/media/formats/webm/webm.gyp#L16

packager/media/formats/webm/multi_segment_segmenter.cc
```diff
@@ -28,14 +28,25 @@ Status MultiSegmentSegmenter::FinalizeSegment(uint64_t start_timestamp,
...
+  if (!is_subsegment) {
+    std::string segment_name =
+        GetSegmentName(options().segment_template, start_timestamp,
+                       num_segment_++, options().bandwidth);
+
+    if (!writer_->WriteToFile(segment_name)) {
+      LOG(ERROR) << "Failed creating file for webm segment.";
+    }
+  }
+
```
kqyang, line 40: I think this should be moved after `cluster()->Finalize()`.

packager/media/formats/webm/buffer_mkv_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+  DCHECK(file_);
+
+  const int64_t size = File::CopyFile(source, file_.get(), max_copy);
+  if (size < 0)
+    return size;
+
+  position_ += size;
+  return size;
+}
+
```
kqyang, line 100: They are not needed for buffer based writer. I think we can just remove them.

packager/media/formats/webm/buffer_mkv_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+  if (file_->Seek(position)) {
+    position_ = position;
+    return 0;
+  } else {
+    return -1;
+  }
+}
+
+bool BufferMkvWriter::Seekable() const {
+  return seekable_;
```
kqyang, line 117: I think we can just return false for buffer based writer.

packager/media/formats/webm/buffer_mkv_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+
+mkvmuxer::int32 BufferMkvWriter::Position(mkvmuxer::int64 position) {
+  DCHECK(file_);
+
+  if (file_->Seek(position)) {
+    position_ = position;
+    return 0;
+  } else {
+    return -1;
+  }
```
kqyang, line 113: We don't need to support |file_| read write in buffer based mkv writer. We can just return -1 here.

packager/media/formats/webm/buffer_mkv_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+          file_->Write(data + total_bytes_written, len - total_bytes_written);
+      if (written < 0)
+        return written;
+
+      total_bytes_written += written;
+    }
+
+    DCHECK_EQ(total_bytes_written, len);
+    position_ += len;
+  } else {
```
kqyang, line 76: This branch can be removed as we don't need to support file read/write in buffer based mkv writer.

packager/media/formats/webm/buffer_mkv_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+  const int64_t size = File::CopyFile(source, file_.get(), max_copy);
+  if (size < 0)
+    return size;
+
+  position_ += size;
+  return size;
+}
+
+mkvmuxer::int64 BufferMkvWriter::Position() const {
+  return position_;
```
kqyang, line 102: I think we can just return `segment_buffer_.Size()` here.

packager/media/formats/webm/buffer_mkv_writer.cc
```diff
@@ -0,0 +1,124 @@
...
+Status BufferMkvWriter::CloseFile() {
+  const std::string file_name = file_->file_name();
+  if (!file_.release()->Close()) {
+    return Status(
+        error::FILE_FAILURE,
+        "Cannot close file " + file_name +
+            ", possibly file permission issue or running out of disk space.");
+  }
+  return Status::OK;
+}
```
kqyang, line 59: I think we can combine the three functions to one function: `WriteToFile(const std::string& file_name)`.

packager/media/formats/webm/multi_segment_segmenter.cc
sr1990: @kqyang  Thanks for the review. I kept the file operations and did a writer_->WriteToFile(segment_name) before Cluster->Finalize() because Cluster->Finalize() does the following:
 ```
if (writer_->Seekable()) {

    const int64_t pos = writer_->Position();
 
    if (writer_->Position(size_position_))
      return false;
 
    if (WriteUIntSize(writer_, payload_size(), 8))
      return false;
 
    if (writer_->Position(pos))
      return false;
  }
```
Do you think it makes sense to keep the file operations or modify Write in BufferMkvWriter to insert at a given position?

Comment (kqyang) [reply packager/media/formats/webm/multi_segment_segmenter.cc:40]: No. BufferMkvWriter is not seekable (`bool BufferMkvWriter::Seekable()` should always return false), so the quoted code is not executed.

Comment (sr1990) [reply packager/media/formats/webm/multi_segment_segmenter.cc:40]: Ok. I will update the tests so that the cluster size value which is initialized to "unknown" in Cluster::WriteClusterHeader() is not updated with the above code block when we send false from Seekable().

File: packager/media/formats/webm/webm_parser.cc [hunk 10]
```diff
@@ -948,7 +948,8 @@ bool WebMListParser::OnListEnd() {
     int64_t bytes_parsed = list_state.bytes_parsed_;
     int id = list_state.id_;
 
-    if (bytes_parsed != list_state.size_)
+    if (bytes_parsed != list_state.size_ && 
+        !(list_state.size_ == kWebMUnknownSize && id == kWebMIdCluster))
```
kqyang, line 952: What is this for? What happens without it?

packager/media/formats/webm/buffer_mkv_writer.cc
```diff
@@ -59,62 +34,21 @@ Status BufferMkvWriter::CloseFile() {
...
-
-    DCHECK_EQ(total_bytes_written, len);
-    position_ += len;
-  } else {
-     // Write to buffer if file is not present.
-     const uint8_t* data = reinterpret_cast<const uint8_t*>(buf);
-     segment_buffer_.AppendArray(data, len);
-     position_ = segment_buffer_.Size();
-  }
+  const uint8_t* data = reinterpret_cast<const uint8_t*>(buf);
```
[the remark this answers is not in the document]
sr1990: ` segment_buffer_.AppendArray((uint8_t*)buf, len);` works

packager/media/formats/webm/webm_parser.cc
sr1990: During DoInitilize in MultiSegementSegmenter,
mkvmuxer::3632 writes 
```
 // Write "unknown" (EBML coded -1) as segment size value. We need to write 8      
// bytes because if we are going to overwrite the segment size later we do        
// not know how big our segment will be.                                          
 if (SerializeInt(writer_header_, kEbmlUnknownValue, 8))                          
    return false; 
```

During `Cluster::Finalize:` (mkvmuxer.cc::2561): The above value is not overwritten with the size by seeking to position remembered in DoInitialize as the buffer_mkv_writer is not seekable. For ref, the following code is not executed:
```
if (writer_->Seekable()) {
   const int64_t pos = writer_->Position();
   if (writer_->Position(size_position_))
     return false;
   if (WriteUIntSize(writer_, payload_size(), 8))
     return false;
   if (writer_->Position(pos))
     return false;
 }

```
In `WebMListParser::Parse` (line 770) Parse Error is not returned if Cluster size is kWebMUnknownSize and id is kWebMIdCluster and gets pushed in the list_state stack in `WebMListParser::OnListStart()`. 

In `WebMListParser::OnListEnd()` - The above line 958, the bytes parsed will not be equal to list_state.size as it is unkown and not updated with the cluster size and breaks. Due to this,  `SegmentTestBase::ClusterParser::OnListEnd` (from line webm_parser.cc::966) is never called and in_cluster_ is never modified to false and because of this the parsing of next segment cluster fails and that is why `MultiSegmentSegmenterTest.SplitsFilesOnSegment `fails. 

I have attached a diff of hex dump of the files created with and without this change.

Please let me know if I am missing anything and what you think.

DiffWebM

sr1990:
> Have you tried if the result DASH streams play in Chrome?

I have not tried on Chrome, but exoplayer did not work (playback failed). I have not debugged yet.

packager/media/formats/webm/webm_parser.cc
kqyang: Thanks for info!

It could be what is causing exoplayer to fail, and maybe Chrome too.

Sorry for my oversight on this.

I thought about it again. Another way (probably better) to address this issue is to write to a memory file first, then copy from the memory file to the final file.

The change is also simpler. Sorry again for my oversight.

sr1990: I am able to play video segments individually on both chrome and exoplayer. 
Issue with exoplayer is mpd parsing. MPD at https://github.com/google/shaka-packager/blob/master/packager/app/test/testdata/live-profile-with-webm/output.mpd has `publishTime="some_time" availabilityStartTime="some_time"` which it is unable to parse.

sr1990: Ok was testing the wrong mpd. I packaged the content again and was able to play on exoplayer.

kqyang: > However there is an issue with video start. I have to seek to the beginning of the stream for it to play normally. 

That is weird. 

> I packaged the content again and was able to play on exoplayer. 

Thanks for checking. Can you try if the mpd plays in Chrome? And yes, you need to package it with static_live.

sr1990: @kqyang I am able to play on both Chrome and exoplayer now after packaging with static live.

sr1990: Further looking at exoplayer for reading webm segments,

1.MatroskaExtractor::read() https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/MatroskaExtractor.java#L449
continues to read with DefaultEmblReader::read
https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/DefaultEbmlReader.java#L79

 till it sends back  C.RESULT_END_OF_INPUT. Then the next segment download starts. 

2. DefaultEmblReader::read() also populates the masterElementsStack with the read id and size and pops it out when current position is >= elementEndPosition.
https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/DefaultEbmlReader.java#L111 
https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/DefaultEbmlReader.java#L81

 In this case, there is an entry with elementEndPosition: 72057594037927947  elementId: 524531317.

3. The stack is cleared with DefaultEmblReader::reset
https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/extractor/mkv/DefaultEbmlReader.java#L72
which is called at the beginning of a new segment download. 

4. DefaultEmblReader.java (read()) gets RESULT_END_OF_INPUT from CacheDataSource.java read() when bytesRemaining == 0.
https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/CacheDataSource.java#L302


5. In this case, bytesRemaining is set at CacheDataSource.java
https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/upstream/cache/CacheDataSource.java#L492
which it gets from: 
https://github.com/google/ExoPlayer/blob/5bfad37cd0d2917f8c62440a42e1f65aa535cac7/library/core/src/main/java/com/google/android/exoplayer2/upstream/DataSource.java#L64

6. In this case, it relies on bytesRemaining which is set in (5) and not on elementEndPosition which is stored in masterElementsStack in (2). After (4) the stack is cleared with (3) which removes the entry (elementEndPosition: 72057594037927947  elementId: 524531317). So it does not look like the exoplayer is dependent on segment size value. 

Please let me know what you think and if I am missing anything,

sr1990: @kqyang Any thoughts?

kqyang: Sorry for the late reply. I am still a bit concerned with this change as it does not quite make sense. This is assuming the size of the list is always 1. 

I am a bit concerned with the handling in Chrome and exoplayer. 

Chrome handles the list in the same way as the original code in Shaka Packager (without the workaround in line 951). It works now because the next element is still a cluster so the list does not actually end.

Can you try using `--ad_cues 600;1800;3000` (choose appropriate ad breaks) to create a multi-period MPD to see if it still works in Chrome and ExoPlayer?

Sorry for the extra work.

sr1990: @kqyang I was able to play the content after adding ad cues and having a mult-period mpd in both Chrome and Exoplayer.

I had made the above change as ` TestCase: MultiSegmentSegmenterTest.SplitsFilesOnSegment` was failing.
- SegmentTestBase::ClusterParser::OnListStart(int id) : packager/media/formats/webm/segmenter_test_base.cc is called by  WebMListParser::ParseListElement->WebMListParser::OnListStart to get WebMParserClient webm_parser.cc::931
- For the first segment the parser is initialized correctly as in_cluster in SegmentTestBase::ClusterParser::OnListStart(int id) is false and when id is id: 524531317  size: 72057594037927935 in WebMListParser::OnListEnd() as the size is unknown, SegmentTestBase::ClusterParser::onListEnd() is not called and in_cluster remains true.
- For the second segment, as in_cluster is true, WebMListParser::OnListStart->OnListStart(id) returns null and test case fails.
The above change is to handle this case.

Handling in Chrome:
- `WebMClusterParser::OnListEnd `and `OnListStart `implementations are different from `SegmentTestBase::ClusterParser`. So, not calling `OnListEnd`  and not removing element like id: 524531317  size: 72057594037927935 is handled correctly.

sr1990: Handling in Exoplayer, following are the logs of MasterElementStack. I am able to play multi-period changes with buffer_mkv_writer changes.

With original code:
All the elements in MasterElementStack are popped before reset is called except segmentinfo element.

```
D/WEBMTEST-CacheDataSource: OPEN DATASPEC : /noChangeWithAds/webmstream.mpd
Init segment:

D/WEBMTEST-CacheDataSource: OPEN DATASPEC : /noChangeWithAds/webm_320p/init.webm
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 5  elementEndPosition: 36  elementId: 440786851  element content size: 31
D/WEBMTEST: masterElement StackSize before popping ID: 440786851 SIZE: 1

D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 48  elementEndPosition: 72057594037927983  elementId: 408125543  element content size: 72057594037927935
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 142  elementEndPosition: 235  elementId: 357149030  element content size: 93
D/WEBMTEST: masterElement StackSize before popping ID: 357149030 SIZE: 2
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 240  elementEndPosition: 292  elementId: 374648427  element content size: 52
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 242  elementEndPosition: 292  elementId: 174  element content size: 50
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 276  elementEndPosition: 292  elementId: 224  element content size: 16
D/WEBMTEST: masterElement StackSize before popping ID: 224 SIZE: 4
D/WEBMTEST: masterElement StackSize before popping ID: 174 SIZE: 3
D/WEBMTEST: masterElement StackSize before popping ID: 374648427 SIZE: 2  <-- size is 2 because elementEndPosition: 72057594037927983  elementId: 408125543 is not removed from the stack

D/WEBMTEST-CacheDataSource: ID: 14745 Return end of Input

D/WEBMTEST-ContainerMediaChunk load(): INITIALIZE : /noChangeWithAds/webm_320p/1.webm
D/WEBMTEST: CLEAR STACK : 1    <----- Stack containing the last element is cleared by calling DefaultEbmlReader::reset()

D/WEBMTEST-ContainerMediaChunk load(): URI: /noChangeWithAds/webm_320p/1.webm
D/WEBMTEST-CacheDataSource: OPEN DATASPEC : /noChangeWithAds/webm_320p/1.webm
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 12  elementEndPosition: 253565  elementId: 524531317  element content size: 253553
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 251939  elementEndPosition: 253565  elementId: 160  element content size: 1626
D/WEBMTEST: masterElement StackSize before popping ID: 160 SIZE: 2
D/WEBMTEST: masterElement StackSize before popping ID: 524531317 SIZE: 1   <--- cluster element is popped when set size is read
D/WEBMTEST-CacheDataSource: ID: 14745 Return end of Input


D/WEBMTEST-ContainerMediaChunk load(): INITIALIZE : /noChangeWithAds/webm_320p/2.webm

D/WEBMTEST: CLEAR STACK : 0  


```

```
WITH CHANGE:
INIT SEGMENT:
D/WEBMTEST-CacheDataSource: OPEN DATASPEC : /withChangeWithAds/webmstream.mpd
D/WEBMTEST-CacheDataSource: OPEN DATASPEC : /withChangeWithAds/webm_640p/init.webm
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 5  elementEndPosition: 36  elementId: 440786851  element content size: 31
D/WEBMTEST: masterElement StackSize before popping ID: 440786851 SIZE: 1
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 48  elementEndPosition: 72057594037927983  elementId: 408125543  element content size: 72057594037927935
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 142  elementEndPosition: 235  elementId: 357149030  element content size: 93
D/WEBMTEST: masterElement StackSize before popping ID: 357149030 SIZE: 2
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 240  elementEndPosition: 294  elementId: 374648427  element content size: 54
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 242  elementEndPosition: 294  elementId: 174  element content size: 52
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 276  elementEndPosition: 294  elementId: 224  element content size: 18
D/WEBMTEST: masterElement StackSize before popping ID: 224 SIZE: 4
D/WEBMTEST: masterElement StackSize before popping ID: 174 SIZE: 3
D/WEBMTEST: masterElement StackSize before popping ID: 374648427 SIZE: 2
D/WEBMTEST-CacheDataSource: ID: 14722 Return end of Input


D/WEBMTEST-ContainerMediaChunk load(): INITIALIZE : /withChangeWithAds/webm_640p/1.webm
 D/WEBMTEST: CLEAR STACK : 1   <----- Same as above

For MediaSegments:
D/WEBMTEST-CacheDataSource: OPEN DATASPEC : /withChangeWithAds/webm_640p/1.webm
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 12  elementEndPosition: 72057594037927947  elementId: 524531317  element content size: 72057594037927935
D/WEBMTEST:  PUSHING INTO MASTERELEMENT STACK - contentPosition: 442588  elementEndPosition: 445730  elementId: 160  element content size: 3142
D/WEBMTEST: masterElement StackSize before popping ID: 160 SIZE: 2
D/WEBMTEST-CacheDataSource: ID: 14722 Return end of Input
D/WEBMTEST-ContainerMediaChunk load(): INITIALIZE : /withChangeWithAds/webm_640p/2.webm
D/WEBMTEST: CLEAR STACK : 1  <---- In this case the cluster element entry is not popped but cleared here
```

Let me know what you think.

sr1990: @kqyang I have created another PR where mkvwriter writes to memory file and dumps the content into webm file in FinalizeSegment.

kqyang: Thanks @sr1990 for the extra work! That PR LGTM with a few nits.

cosmin:
@sr1990 was this replaced by #803 ?

sr1990:
@cosmin yes. closing.

PR #796, opened 2020-06-30
pull request test

koln67:
Pull Request Submission Test

### DIFF
--- a/packager/media/codecs/es_descriptor.h
+++ b/packager/media/codecs/es_descriptor.h
@@ -175,7 +175,7 @@ class ESDescriptor : public BaseDescriptor {
   ESDescriptor() : BaseDescriptor(DescriptorTag::kES) {}
 
   uint16_t esid() const { return esid_; }
-  void set_esid(uint16_t esid) { esid_ = esid; }
+  void set_esid(uint16_t esid) { esid_ = 0; }
 
   const DecoderConfigDescriptor& decoder_config_descriptor() const {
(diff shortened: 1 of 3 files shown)


packager/media/codecs/es_descriptor.h
```diff
@@ -175,7 +175,7 @@ class ESDescriptor : public BaseDescriptor {
   ESDescriptor() : BaseDescriptor(DescriptorTag::kES) {}
 
   uint16_t esid() const { return esid_; }
-  void set_esid(uint16_t esid) { esid_ = esid; }
+  void set_esid(uint16_t esid) { esid_ = 0; }
```
kqyang, line 178: This is not the correct way to address it.

Since we are not expecting the users to modify it, we should remove this function. We should also update ESDescriptor::WriteInternal to write "0" instead of writing |esid_|.

packager/media/formats/mp4/box_definitions.cc
```diff
@@ -1020,7 +1020,7 @@ SampleGroupDescription::SampleGroupDescription() = default;
 SampleGroupDescription::~SampleGroupDescription() = default;
 
 FourCC SampleGroupDescription::BoxType() const {
-  return FOURCC_sgpd;
+  return FOURCC_senc;
```
kqyang, line 1023: That is incorrect way to address the problem as the box type cannot be "senc". It will cause playback problems.

PR #803, opened 2020-07-21
Write to memory file before writing to file for WebM segments generation.

sr1990:


packager/media/formats/webm/multi_segment_segmenter.cc
```diff
@@ -32,10 +32,14 @@ Status MultiSegmentSegmenter::FinalizeSegment(uint64_t start_timestamp,
...
-    const std::string segment_name = writer_->file()->file_name();
-    // Close the file, which also does flushing, to make sure the file is
-    // written before manifest is updated.
-    RETURN_IF_ERROR(writer_->Close());
+    std::string segment_name =
+        GetSegmentName(options().segment_template, start_timestamp,
+                       num_segment_, options().bandwidth);
+
+    if (!writer_->WriteToFile(segment_name))
+      return Status(error::FILE_FAILURE, "Unable to write file.");
```
kqyang, line 40: A simpler approach without the need to modify MkvWriter is to use `File::Copy`. Also, the file needs to be deleted otherwise the memory will not be de-allocated, which could result in memory leak.

// Close the file, which also does flushing, to make sure the file is
// written before manifest is updated.
RETURN_IF_ERROR(writer_->Close());

if (!File::Copy(temp_file_name_.c_str(), segment_name.c_str())) {
  ...
}
if (!File::Delete(temp_file_name_.c_str()) {
  ...
}

packager/media/formats/webm/multi_segment_segmenter.cc
```diff
@@ -77,15 +88,16 @@ Status MultiSegmentSegmenter::DoFinalize() {
 Status MultiSegmentSegmenter::NewSegment(uint64_t start_timestamp,
                                          bool is_subsegment) {
   if (!is_subsegment) {
-    // Create a new file for the new segment.
-    std::string segment_name =
+    temp_file_name_ =
```
kqyang, line 91: It should be either a memory file (prefix with "memory://") or an actual temporary file:

```c++
  if (!TempFilePath(options().temp_dir, &temp_file_name_))
    return Status(error::FILE_FAILURE, "Unable to create temporary file.");
```

This is because `options().segment_template` could be a remote file, e.g. HttpFile. We don't want to create a temporary file there.

sr1990: Sorry about that. I have made the change.

sr1990: @kqyang Do you want to change temp_file_name_ to memory_file_name_?

kqyang: No. It is fine with me. Thanks.

PR #822, opened 2020-08-12
draft change to ts muxer to add frame rate

koln67:
Attempting to add the FRAME-RATE attribute when the output is a segment template in .ts format.


packager/media/formats/mp2t/ts_muxer.cc
```diff
@@ -49,7 +49,8 @@ Status TsMuxer::FinalizeSegment(size_t stream_id,
 void TsMuxer::FireOnMediaStartEvent() {
   if (!muxer_listener())
     return;
-  muxer_listener()->OnMediaStart(options(), *streams().front(), kTsTimescale,
+  muxer_listener()->OnMediaStart(options(), *streams().front(), 
+                                 streams().front()->time_scale(),
```
kqyang, line 53: Don't alter the output timescale. It has to be fixed at kTsTimescale.

packager/media/formats/mp2t/ts_muxer.cc
```diff
@@ -34,12 +30,16 @@ Status TsMuxer::Finalize() {
 
 Status TsMuxer::AddSample(size_t stream_id, const MediaSample& sample) {
   DCHECK_EQ(stream_id, 0u);
+  sample_duration_ = sample.duration();
```
kqyang, line 33: I think you can call OnSampleDurationReady here, i.e.

```c++
if (sample_duration_ == 0) {
  sample_duration_ = sample.duration() * kTsTimescale / streams().front()->time_scale();
  if (muxer_listener()) 
    muxer_listener()->OnSampleDurationReady(sample_duration_);
} 
```

`if (sample_duration_ == 0)` is needed to avoid issuing event multiple times, which is needed only once.

The below code is needed to re-scale the timestamp to output timescale.

```c++
sample.duration() * kTsTimescale / streams().front()->time_scale()
```

packager/media/formats/mp2t/ts_muxer.cc
koln67: Added back constkTsTimeScale

packager/media/formats/mp2t/ts_muxer.cc
koln67: Changed, it seems to work properly with a frame rate of 24. Currently working on unit tests

kqyang: Looks like there isn't a unit-test. I think it is fine to just update the end to end test.

koln67:
The golden files are updated. However, there is another pull request ongoing, which also affects the golden files, so I'm not sure if I should wait until that one is merged.

Note: the 2 non-deterministic files can be reverted, but currently waiting on the other ongoing pull request.

kqyang:
This PR is deprecated by https://github.com/google/shaka-packager/pull/823.

PR #824, opened 2020-08-21
use second sample duration if it exists

koln67:
As seen in #823, in TSMuxer::AddSample from the file ts_muxer.cc, if we always use the first sample duration, there is a case in the end-to-end tests where the frame rate is 9.990. For that special case, if we use the second sample duration, the frame rate is 29.97, which is more reasonable.

This change would use the second sample_duration, if it exists.


packager/media/formats/mp2t/ts_muxer.cc
```diff
@@ -34,10 +34,12 @@ Status TsMuxer::Finalize() {
 
 Status TsMuxer::AddSample(size_t stream_id, const MediaSample& sample) {
   DCHECK_EQ(stream_id, 0u);
-  if (sample_duration_ == 0) {
-    sample_duration_ = sample.duration() * kTsTimescale / streams().front()->time_scale();
-    if (muxer_listener()) 
-      muxer_listener()->OnSampleDurationReady(sample_duration_);
+  if (num_samples_ < 2) {
+    sample_durations_[num_samples_] = sample.duration() * kTsTimescale / 
+                                     streams().front()->time_scale();
```
kqyang, line 39: Fix the alignment.

PR #828, opened 2020-08-27
change major brand from isom to mp41

koln67:
This fixes the following two warnings from the DASH Conformance Tool.

1. ### error: ftyp-1 ### The brand 'isom' can only be a compatible, not major, brand
2. ### Warning: major brand ('isom') not also found in list of compatible brands

Spec reference for this change:

Based on ISO/IEC 14496-15:2004(E), Section 5.3.2, 

"If the extensions are required, then the brand ‘avc1’ should be used. In this case, in a file with extension “.mp4”, the major
brand may be ‘avc1’."

Shaka packager uses those extensions, which includes sgpd (SampleGroupDescriptionBox), and SampleToGroup. The input files have a .mp4 extension as well, which suggests that the major brand should be avc1.


packager/media/formats/mp4/mp4_muxer.cc
```diff
@@ -206,10 +206,11 @@ Status MP4Muxer::DelayInitializeMuxer() {
   std::unique_ptr<FileType> ftyp(new FileType);
   std::unique_ptr<Movie> moov(new Movie);
 
-  ftyp->major_brand = FOURCC_isom;
+  ftyp->major_brand = FOURCC_avc1;
```
kqyang, line 209: What is your rational for using avc1 as the major brand?

koln67: Based on ISO/IEC 14496-15:2004(E), Section 5.3.2,

"If the extensions are required, then the brand ‘avc1’ should be used. In this case, in a file with extension “.mp4”, the major
brand may be ‘avc1’."

Shaka packager uses those extensions, which includes sgpd (SampleGroupDescriptionBox), and SampleToGroup. The input files have a .mp4 extension as well, which suggests that the major brand should be avc1.

koln67: Also, based on this spec, ISO/IEC 14496-12:2015(E)

E.3The ‘avc1’ brand
The	 brand ‘avc1’ shall be used to	indicate that the file	is conformant with the ‘AVC Extensions’	in subclauses 8.6.4	and 8.9.	If used without	other brands, this implies that	support for those extensions is required. The	use of ‘avc1’ as a major‐brand may be permitted by specifications; in that case, that specification defines the file extension and required behaviour.

koln67: In section 4.3, it says mp41 should be considered the major brand if there is no FTYP. However, we have a FTYP.

Files with no file‐type box should be read as if they contained an FTYP box with Major_brand='mp41', minor_version=0, and	the single	compatible brand 'mp41'.

kqyang: From your first two comments, I think we already meet the requirement as we have already used the brand "avc1". I don't see it mentioning that it should be the major brand.

Re your last comment, sounds like we should use major brand "mp41".

koln67:
avc1 was added back to compatible brands. End to end tests were updated, although for some of the files, I am not sure if they were nondeterministic files, or if changing the major brand caused the golden files to have different contents.

packager/media/formats/mp4/mp4_muxer.cc
```diff
@@ -206,10 +206,11 @@ Status MP4Muxer::DelayInitializeMuxer() {
   std::unique_ptr<FileType> ftyp(new FileType);
   std::unique_ptr<Movie> moov(new Movie);
 
-  ftyp->major_brand = FOURCC_isom;
+  ftyp->major_brand = FOURCC_mp41;
   ftyp->compatible_brands.push_back(FOURCC_iso8);
   ftyp->compatible_brands.push_back(FOURCC_mp41);
   ftyp->compatible_brands.push_back(FOURCC_dash);
+  ftyp->compatible_brands.push_back(FOURCC_avc1);
```
kqyang, line 213: I thought you removed a fourcc in the original code. Didn't notice that you hard-coded it yourself. It is already included in line 223.

Sorry for my oversight. 

Can you replace it with FOURCC_isom? And move it after iso8?

PR #829, opened 2020-08-27
don't add edit box if video track

koln67:
This solves the following error from the DASHconformance tool output by skipping the EditListBox if the track is a video track.

**'CMAF check 'cmf2' violated: Section 7.7.2- For video CMAF tracks, the EditListBox SHALL NOT be present', elst box found for Rep 2.

### DIFF
--- a/packager/media/formats/mp4/mp4_muxer.cc
+++ b/packager/media/formats/mp4/mp4_muxer.cc
@@ -269,7 +269,8 @@ Status MP4Muxer::DelayInitializeMuxer() {
 
     // Generate EditList if needed. See UpdateEditListOffsetFromSample() for
     // more information.
-    if (edit_list_offset_.value() > 0) {
+    const bool isVideoStream = stream->stream_type() == kStreamVideo;
+    if (edit_list_offset_.value() > 0 && !isVideoStream) {
       EditListEntry entry;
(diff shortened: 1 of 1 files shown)


packager/media/formats/mp4/mp4_muxer.cc
```diff
@@ -269,7 +269,8 @@ Status MP4Muxer::DelayInitializeMuxer() {
 
     // Generate EditList if needed. See UpdateEditListOffsetFromSample() for
     // more information.
-    if (edit_list_offset_.value() > 0) {
+    const bool isVideoStream = stream->stream_type() == kStreamVideo;
+    if (edit_list_offset_.value() > 0 && !isVideoStream) {
```
kqyang, line 273: We cannot just skip adding the EditList here as it will introduce problems elsewhere. Did you find the clause that EditList shall not be present in the spec?

cosmin:
This is only true for `cmf2` compliance and it's not clear that this is currently targeting `cmf2` (and with `cmfc`  edit lists are fine). https://docs.google.com/document/d/1KplnJ-u2GvfG6pMZyb1VzZF2ZNmBX_evcsbWxV-iDcE has more details

That said some option to specify the CMAF profile to use would be nice, so anyone that wants to target `cmf2` can specify that.

PR #835, opened 2020-09-02
use second sample in webm and mp4 formats

koln67:
Just as in #824 with mp2t, the second sample's duration should be stored instead of the first sample's duration for the webm and mp4 formats.


koln67:
I believe the webm timescale calculation was also supposed to be changed, but I'm not sure how to correctly calculate it.

packager/media/formats/mp4/segmenter.h
```diff
@@ -144,7 +144,8 @@ class Segmenter {
   ProgressListener* progress_listener_ = nullptr;
   uint64_t progress_target_ = 0u;
   uint64_t accumulated_progress_ = 0u;
-  uint32_t sample_duration_ = 0u;
+  uint32_t sample_durations_[2] = {0, 0};
+  uint32_t num_samples_ = 0u;
```
kqyang, line 148: s/uint32_t/size_t/

packager/media/formats/webm/segmenter.h
```diff
@@ -139,7 +139,8 @@ class Segmenter {
   uint64_t progress_target_ = 0;
   uint64_t accumulated_progress_ = 0;
   uint64_t first_timestamp_ = 0;
-  int64_t sample_duration_ = 0;
+  int64_t sample_durations_[2] = {0, 0};
+  int64_t num_samples_ = 0;
```
kqyang, line 143: s/int64_t/size_t/

packager/media/formats/mp4/segmenter.h
```diff
@@ -98,7 +98,7 @@ class Segmenter {
 
   /// @return The sample duration in the timescale of the media.
   ///         Returns 0 if no samples are added yet.
-  uint32_t sample_duration() const { return sample_duration_; }
+  uint32_t sample_duration() const { return sample_durations_[1]; }
```
kqyang, line 101: Can you change uint32_t to int64_t? Same below.

packager/media/formats/webm/segmenter.cc
```diff
@@ -161,11 +161,13 @@ Status Segmenter::Finalize() {
...
-  if (sample_duration_ == 0) {
-    first_timestamp_ = sample->pts();
-    sample_duration_ = sample->duration();
-    if (muxer_listener_)
-      muxer_listener_->OnSampleDurationReady(sample_duration_);
+  if (num_samples_ < 2) {
+    sample_durations_[num_samples_] = sample->duration();
+    if (num_samples_ == 0)
+      first_timestamp_ = sample->pts();
+    else
```
kqyang, line 168: Check `if (muxer_listener_)`

kqyang (commented):
Thanks. A few nits. Can you update the end to end test files?

packager/media/formats/webm/segmenter.h
koln67: Should num_samples be size_t? I'm assuming only sample_durations_[2] needs to be size_t

kqyang: num_samples_ should be size type, so it should be `size_t`. sample_durations_ should still be int64_t.

koln67:
Replaced with #838

Issue #733, opened 2020-03-18
Refactor Period::ProtectedAdaptationSetMap and merge it into AdaptationSet


kqyang:
The current logics in Period::ProtectedAdaptationSetMap should be moved into AdaptationSet class. This can help simplify the code.

PR #844, opened 2020-09-20
refactor: merge Period::ProtectedAdaptationSetMap  into AdaptationSet

sr1990:
This change is related to https://github.com/google/shaka-packager/issues/733.

### DIFF
--- a/packager/mpd/base/adaptation_set.cc
+++ b/packager/mpd/base/adaptation_set.cc
@@ -216,8 +216,8 @@ bool AdaptationSet::MatchAdaptationSet(
   if (!media_info.has_protected_content())
     return false;
 
-  return protected_content_->SerializeAsString() ==
-         media_info.protected_content().SerializeAsString();
+  return ProtectedContentEq(*protected_content_,
+                            media_info.protected_content());
 }
 
 std::set<std::string> GetUUIDs(
--- a/packager/mpd/base/adaptation_set.cc
+++ b/packager/mpd/base/adaptation_set.cc
@@ -177,11 +177,70 @@ AdaptationSet::AdaptationSet(const std::string& language,
                              uint32_t* counter)
     : representation_counter_(counter),
       language_(language),
-      mpd_options_(mpd_options) {
+      mpd_options_(mpd_options),
+      protected_content_(nullptr) {
   DCHECK(counter);
 }
 
-AdaptationSet::~AdaptationSet() {}
+AdaptationSet::~AdaptationSet() {
+  delete protected_content_;
+}
+
+void AdaptationSet::set_protected_content(const MediaInfo& media_info) {
+  DCHECK(!protected_content_);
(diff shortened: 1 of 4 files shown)


sr1990:
Hi @kqyang, any thoughts on this?

cosmin:
I rebased this to latest and at a high level it looks good to me, let's see if all the tests pass.

packager/mpd/base/adaptation_set.cc
```diff
@@ -177,11 +177,70 @@ AdaptationSet::AdaptationSet(const std::string& language,
...
+
+void AdaptationSet::set_protected_content(const MediaInfo& media_info) {
+  DCHECK(!protected_content_);
+  protected_content_ =
+      new MediaInfo::ProtectedContent(media_info.protected_content());
+}
+
+// The easiest way to check whether two protobufs are equal, is to compare the
+// serialized version.
+bool ProtectedContentEq(
```
cosmin, line 197: this method seems unused now

packager/mpd/base/adaptation_set.cc
```diff
@@ -177,11 +177,70 @@ AdaptationSet::AdaptationSet(const std::string& language,
...
+  if (!content_protection_in_adaptation_set)
+    return true;
+
+  if (!protected_content_)
+    return !media_info.has_protected_content();
+
+  if (!media_info.has_protected_content())
+    return false;
+
+  return protected_content_->SerializeAsString() ==
```
cosmin, line 219: should this use `ProtectedContentEq`? Or should `ProtectedContentEq` be deleted?

packager/mpd/base/adaptation_set.cc
```diff
@@ -177,11 +177,70 @@ AdaptationSet::AdaptationSet(const std::string& language,
...
+    const MediaInfo::ProtectedContent* protected_content) {
+  std::set<std::string> uuids;
+  for (const auto& entry : protected_content->content_protection_entry())
+    uuids.insert(entry.uuid());
+  return uuids;
+}
+
+bool AdaptationSet::SwitchableAdaptationSet(
+    const AdaptationSet& adaptation_set) {
+  if (!protected_content_)
```
cosmin, line 233: I think this method could be written in a way that makes it more clear that it's switchable if protected content is false on both, or true on both and matching UUIDs. otherwise not switchable.

cosmin:
lgtm to me overall but could use a second set of eyes

packager/mpd/base/adaptation_set.cc
sr1990: I think ProtectedContentEq should be used here similar to the removed Period::ProtectedAdaptationSetMap::Match. I am ok with removing if you think there is no future use.

cosmin (approved):

joeyparrish (approved):

PR #848, opened 2020-09-26
[DASH] Template based url construction with RepresentationID.

sr1990:
Trying to implement template based url construction that uses $RepresentationID$. This will add SegmentTemplate below adaptationSet instead of every Representation.
I have added a new stream descriptor field called rep_id that takes in the representation id. 

Example:
```
../packager \                                                                  
  'in=S-1024x576.mp4,stream=video,init_segment=$RepresentationID$/init.mp4,segment_template=$RepresentationID$/$Number$.m4s,rep_id=RES-1024x576' \
  'in=S-1280x720.mp4,stream=video,init_segment=$RepresentationID$/init.mp4,segment_template=$RepresentationID$/$Number$.m4s,rep_id=RES-1280x720' \
  --mpd_output test_mpd.mpd
```

and following is the generated mpd: 

```
<?xml version="1.0" encoding="UTF-8"?>                                          
<!--Generated with https://github.com/google/shaka-packager version 5b9fd409a5-debug-->
<MPD xmlns="urn:mpeg:dash:schema:mpd:2011" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:mpeg:dash:schema:mpd:2011 DASH-MPD.xsd" profiles="urn:mpeg:dash:profile:isoff-live:2011" minBufferTime="PT2S" type="dynamic" publishTime="2020-09-26T00:16:44Z" availabilityStartTime="2020-09-26T00:16:43Z" minimumUpdatePeriod="PT5S" timeShiftBufferDepth="PT1800S">
  <Period id="0" start="PT0S">                                                  
    <AdaptationSet id="0" contentType="video" maxWidth="1280" maxHeight="720" maxFrameRate="96/4" segmentAlignment="true" par="16:9">
      <SegmentTemplate timescale="96" initialization="$RepresentationID$/init.mp4" media="$RepresentationID$/$Number$.m4s" startNumber="1">
        <SegmentTimeline>                                                      
          <S t="0" d="576" r="7"/>                                              
          <S t="4608" d="404"/>                                                
        </SegmentTimeline>                                                      
      </SegmentTemplate>                                                        
      <Representation id="RES-1024x576" bandwidth="212490" codecs="avc1.4d401f" mimeType="video/mp4" sar="1:1" width="1024" height="576" frameRate="96/192"/>
      <Representation id="RES-1280x720" bandwidth="2942370" codecs="avc1.4d401f" mimeType="video/mp4" sar="1:1" width="1280" height="720" frameRate="96/4"/>
    </AdaptationSet>                                                            
  </Period>                                                                    
</MPD> 
```

This is similar to MP4Box where ids are specified along with tracks: 
Example:
```
mp4box -dash 12000 -frag 12000 -mem-frags -rap -profile dashavc264:live ^
  -profile-ext "urn:hbbtv:dash:profile:isoff-live:2012" ^
  -mpd-title "" -mpd-info-url "" ^
  -bs-switching no -sample-groups-traf -single-traf -subsegs-per-sidx 1 ^
  -segment-name "$RepresentationID$_$Number$$Init=i$" -segment-timeline ^
  -out "manifest_orig.mpd" ^
  "temp-360p.mp4#trackID=1:id=360p" ^
  "temp-720p.mp4#trackID=1:id=720p" ^
  "temp-1080p.mp4#trackID=1:id=1080p" ^
  "temp-128k.mp4#trackID=1:id=128k"
```

@kqyang please let me know what you think about this approach.

### DIFF
--- a/packager/media/base/muxer_options.h
+++ b/packager/media/base/muxer_options.h
@@ -45,6 +45,10 @@ struct MuxerOptions {
   /// User-specified bit rate for the media stream. If zero, the muxer will
   /// attempt to estimate.
   uint32_t bandwidth = 0;
+
+  /// Used as Representation id for template based url construction using
+  /// $RepresentationID$.
+  std::string rep_id;
 };
 
 }  // namespace media
--- a/packager/mpd/base/xml/xml_node.cc
+++ b/packager/mpd/base/xml/xml_node.cc
@@ -197,6 +197,10 @@ void XmlNode::SetId(uint32_t id) {
   SetIntegerAttribute("id", id);
 }
 
+void XmlNode::SetIdString(std::string id) {
+  SetStringAttribute("id", id);
+}
+
 void XmlNode::SetContent(const std::string& content) {
   DCHECK(node_);
   xmlNodeSetContent(node_.get(), BAD_CAST content.c_str());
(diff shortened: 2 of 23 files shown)


sr1990:
Hi @kqyang, any thoughts on this PR?

kqyang:
@sr1990 Thanks for working on it. Yes, that is a great feature. (And sorry for late reply, super busy recently..)

I'll try to look at the PR some time this week.

sr1990:
> @sr1990 Thanks for working on it. Yes, that is a great feature. (And sorry for late reply, super busy recently..)
> 
> I'll try to look at the PR some time this week.

Sounds good. Thanks. I just wanted to check if the approach is right. If it is correct, then I can move ahead with adding unit tests and documentation.

sr1990:
Ping

ProIcons:
bump

packager/media/base/muxer_options.h
```diff
@@ -45,6 +45,10 @@ struct MuxerOptions {
   /// User-specified bit rate for the media stream. If zero, the muxer will
   /// attempt to estimate.
   uint32_t bandwidth = 0;
+
+  /// Used as Representation id for template based url construction using
+  /// $RepresentationID$.
+  std::string rep_id;
```
sr1990, line 51: This should not be added as muxer option. Will update the PR soon.

joeyparrish:
I'm trying to catch up on the PR backlog as we work to revive the project.

I agree with this feature in principle, but haven't reviewed the code yet.  Please merge main into your branch and fix conflicts, and make whatever other changes you're considering (I see a comment you left about moving `rep_id` out of `muxer_options.h`).  When it's ready for review, let me know, and I'll give it a detailed review.  Thanks!

packager/packager.h
```diff
@@ -133,6 +133,8 @@ struct StreamDescriptor {
   bool dash_only = false;
   /// Set to true to indicate that the stream is for hls only.
   bool hls_only = false;
+
+  std::string rep_id;
```
cosmin, line 137: in my experience allowing non-numeric representation IDs is bound to cause problems, last time I tried I believe ExoPlayer failed to parse non-numeric representation IDs for example.

cosmin: instead of having the user specify the representation ID, what about allowing to specify using the representation ID in the output filename and then substitute in the computed representation ID by packager.

packager/mpd/base/xml/xml_node.h
```diff
@@ -66,6 +66,10 @@ class XmlNode {
   /// @param id is the ID for this element.
   void SetId(uint32_t id);
 
+  /// Sets 'id=rep_id' attribute.
+  /// @param id is the ID for this element.
+  void SetIdString(std::string id);
```
cosmin, line 71: this is a problem for the same reason I mentioned elsewhere, even internally id is expected to be uint. So the option of passing in an alternate string ID isn't great.

packager/mpd/base/xml/xml_node.cc
```diff
@@ -404,6 +408,77 @@ bool RepresentationXmlNode::AddVODOnlyInfo(const MediaInfo& media_info) {
   return true;
 }
 
+scoped_xml_ptr<xmlNode> RepresentationXmlNode::GetLiveOnlyInfo(
```
cosmin, line 411: `GetLiveOnlyInfo` doesn't quite seem like the right name for this, wouldn't something like GetSegmentTemplate be more clear about what it does?

cosmin (changes requested):
Overall the adaptation set level segment template seems fine, but non-numeric representation IDs does not seem fine. We could consider allowing the user to use manually specified numeric representation IDs, but it seems like it would be safer to leave representation IDs automatically generated and instead provide a facility to use the generated representation ID in the output filename.

sr1990:
@cosmin 
> Overall the adaptation set level segment template seems fine, but non-numeric representation IDs does not seem fine. 
As per spec:
Represenation@id,
```
specifies an identifier for this Representation. The
identifier shall be unique within a Period unless the
Representation is functionally identical to another Rep-
resentation in the same Period.
The identifier shall not contain whitespace characters.
**If used in the template-based URL construction as de-
fined in subclause 5.3.9.4.4, the string shall only contain
characters that are permitted within an HTTP-URL
according to IETF RFC 3986.**
```
Also exoplayer seems to be parsing Representation id as a string
https://github.com/google/ExoPlayer/blob/faa296ae3c4ac3ceed514b247284f268953afad7/library/dash/src/main/java/com/google/android/exoplayer2/source/dash/manifest/DashManifestParser.java#L688

> We could consider allowing the user to use manually specified numeric representation IDs, but it seems like it would be safer to leave representation IDs automatically generated and instead provide a facility to use the generated representation ID in the output filename.
Can you give an example command?

joeyparrish:
@cosmin, given the responses of @sr1990, do you still disagree with this change??

sr1990:
@joeyparrish closing this for now.

PR #856, opened 2020-10-16
feat: add support for adding Session Based Descriptor Document Essential Property for DASH

sr1990:
Trying to add support for Session Based Descriptor. 
Use case as per the std: 
"While the DASH viewing session per client is a unique interaction, the MPD is often cached on the CDN and
shared by multiple viewing sessions of the same asset. The session-based description mechanism allows
adding session-specific elements to segment URLs without generating unique per-session MPDs."
"A sidecar document, Session-Based description Document (SBD), includes the time ranges and the
corresponding strings, along with additional metadata. The SBD is referenced in the MPD by its URL. "

mpd example: https://github.com/MPEGGroup/DASHPart8/blob/master/example-mpd.xml

To add the Essential Property, have added the following flags:
1. sbd_url_all/video/audio/text -> comma separated urls pointing to SBDs
sbd_url_all -> add EssentialProperties to all the adaptation sets.
sbd_url_video/audio/text -> add EssentialProperties to all video, audio and text adaptation sets respectively.
2. sbd_template_all/video/audio/text -> comma separated template for urls mentioned in sbd_url*. This flag is optional.
3. sbd_key_all/video/audio/text -> key=default value pairs for each Essential Property. Separated by :. Example: k1=v1,k2=v2:k3=v3,k4=v4 . The default value needs to be set to nil if not present.

Example:
```
`../packager

  'in=Sintel-640x360.mp4,stream=video,init_segment=h264_640p/init.mp4,segment_template=h264_640p/\$Number$.m4s' 
  'in=Sintel-768x432.mp4,stream=video,init_segment=h264_768p/init.mp4,segment_template=h264_768p/\$Number$.m4s' 
  --mpd_output h264.mpd --sbd_url_all "url1,url2" --sbd_template_all "t1,t2" --sbd_key_all "k1=v1,k2=v2:k3=v3,k4=v4"`
```

Have added many flags as SBD Essential Property can different for different adaptation sets, so added flags to indicate different adaptation set based on content types. Also PR assumes that SBD EP will be indicated at adaptation set level.
Have added 1 example in packager_test.py. 
Validations have not been added yet.


cosmin:
I rebased this to `main` while otherwise trying to keep the code as close as possible to the original implementation. We can use this as a starting point to address any feedback.

joeyparrish:
I don't love the command-line syntax.  I find it confusing.  I don't find the docs or the PR description helpful in understanding, either.

Happy to return to this after v3.0.0 is released, if @sr1990 wants to make revisions.

joeyparrish:
Closing for inactivity.  Feel free to reopen and rebase, or create a new one, if you're interested in resuming this work.

PR #818, opened 2020-08-07
[DASH] Get start number from muxer.

sr1990:
This PR will set the start number in representation to the segment index that is sent by muxer.

PR #519, opened 2018-11-28
Specify Initial Sequence Number

sammirata:
With this enhancement, you can now specify the initial sequence number
to be used on the generated segments when calling the packager.
With the old implementation, it was always starting with "1".

Issue #270, opened 2017-09-16
Specify Initial Sequence Number


sammirata:
Here is a sample command line:
packager-linux input=s00015072.ts.25274682549.ts,stream_selector=video,init_segment=video.mp4,segment_template=video2-\$Number\$.mp4

sammirata:
I answered my own question by looking at the code. There is no way to change the default of 1 as starting sequence number. Here is a patch that applies to latest version. It adds a new option to control the initial sequence number of the segmenter:

diff --git a/packager/app/muxer_flags.cc b/packager/app/muxer_flags.cc
index a18a4e7..888d486 100644
--- a/packager/app/muxer_flags.cc
+++ b/packager/app/muxer_flags.cc
@@ -50,3 +50,7 @@ DEFINE_bool(mp4_use_decoding_timestamp_in_timeline,
             "be used when generating media timeline, e.g. timestamps in sidx "
             "and mpd. This is to workaround a Chromium bug that decoding "
             "timestamp is used in buffered range, https://crbug.com/398130.");
+DEFINE_int32(mp4_sequence_number_initial,
+             1,
+             "This number allows for the customization of the "
+             "initial sequence_number in the mp4 moof->mfhd.");
\ No newline at end of file
diff --git a/packager/app/muxer_flags.h b/packager/app/muxer_flags.h
index a88a777..a096be4 100644
--- a/packager/app/muxer_flags.h
+++ b/packager/app/muxer_flags.h
@@ -20,5 +20,6 @@ DECLARE_int32(num_subsegments_per_sidx);
 DECLARE_string(temp_dir);
 DECLARE_bool(mp4_include_pssh_in_stream);
 DECLARE_bool(mp4_use_decoding_timestamp_in_timeline);
+DECLARE_int32(mp4_sequence_number_initial);
 
 #endif  // APP_MUXER_FLAGS_H_
diff --git a/packager/app/packager_main.cc b/packager/app/packager_main.cc
index fa7e034..e9d41be 100644
--- a/packager/app/packager_main.cc
+++ b/packager/app/packager_main.cc
@@ -284,6 +284,7 @@ base::Optional<PackagingParams> GetPackagingParams() {
   mp4_params.use_decoding_timestamp_in_timeline =
       FLAGS_mp4_use_decoding_timestamp_in_timeline;
   mp4_params.include_pssh_in_stream = FLAGS_mp4_include_pssh_in_stream;
[...]

PR #879, opened 2021-01-09
feat: get start number from muxer and specify initial sequence number

sr1990:
@kqyang Have merged https://github.com/google/shaka-packager/pull/818 and https://github.com/google/shaka-packager/pull/519 by @sammirata along with the changes mentioned in #519.
Example:
`../packager 'in=Sintel-1280x720.mp4,stream=video,init_segment=h264_1280p/init.mp4,segment_templa    te=h264_1280p/$Number$.m4s' --mpd_output h264.mpd --mp4_initial_sequence_number 0`
Please review and let me know. 
Thanks.


vish91:
@sr1990 thanks for putting in PR.  Initial testing looks good to me. I can try and run more variants if I get time. 
Command:
```
./packager 'in=/home/vshah/shaka/sample.mp4,stream=video,init_segment=/tmp/shaka/dash/init.m4v,segment_template=/tmp/shaka/dash/output/sample/seg_$Number$.m4s' --segment_duration 6 --fragment_duration 6 --mp4_initial_sequence_number 0 --mpd_output /home/vshah/shaka/output/stream-sr1990.mpd
```
``` 
<SegmentTemplate timescale="12800" initialization="/tmp/shaka/dash/init.m4v" 
   media="/tmp/shaka/dash/output/sample/seg_$Number$.m4s" startNumber="0">
```

jigneshdhruv:
@kqyang  Do you think this PR can be merged? Our initial testing looks good. We are waiting on this patch do deliver dash manifest to ps4 devices. Thanks.

docs/source/options/mp4_output_options.rst
```diff
@@ -17,3 +17,8 @@ MP4 output options
     template).
 
     Default enabled.
+
+--mp4_initial_sequence_number
```
kqyang, line 21: Thanks for working on this feature. Here is my high level feedback.

There are two concepts in this CL.

1. mp4_initial_sequence_number which specifies the initial sequence number in moof->mfhd.

2. the starting segment index.

They are different concepts. They may not have the same value as a segment can contain multiple fragments.


Can we have a separate flag for segment index, e.g. start_segment_number to match the `startNumber` in DASH `SegmentTemplate`?

I think we can remove the initial sequence number from this PR (or move it to a separate PR).

 I also think that it can be a ChunkingParam as it is not restricted to mp4.

sr1990: Thanks for the feedback.
`They are different concepts. They may not have the same value as a segment can contain multiple fragments.`
You are right. My mistake. 
I do not see exoplayer's FragmentedMp4Extractor.java parsing it 
https://github.com/google/ExoPlayer/blob/03263db378392385e290c40505d80e990c85a5cb/library/extractor/src/main/java/com/google/android/exoplayer2/extractor/mp4/FragmentedMp4Extractor.java#L708
What is the use-case for specifying moof->mfhd sequence number? 

```
Can we have a separate flag for segment index, e.g. start_segment_number 
to match the startNumber in DASH SegmentTemplate?
I also think that it can be a ChunkingParam as it is not restricted to mp4.
```
Yes that makes sense. Will look into it.

kqyang: > What is the use-case for specifying moof->mfhd sequence number?

I don't really know. None of the players and devices I know care about the sequence number in mfhd. That is actually the question I have in the original PR (#270) too.

vish91: Yeah that makes sense.. we can keep it out. I don't know of use-cases either where it would be needed to change the initial sequence number of moof -> mfhd. 
From what I understand was that segment index start should be enough to fix console platform issues and align us with the DASH spec, that should be the only validation happening player side. Beyond that it just uses either index ranges or the indexed fragmented files from origin for delivery.

sr1990:
@vish91, @kqyang, I have added a start_segment_number flag to indicate startNumber in SegmentTemplate. 
Example:

```
../packager \                                                                   
  'in=bear-640x360.mp4,stream=video,init_segment=bear_360p/init.mp4,segment_template=bear_360p/$Number$.m4s' \
  'in=bear-640x360.mp4,stream=audio,init_segment=bear_audio/init.mp4,segment_template=bear_audio/$Number$.m4s' \
  'in=bear-english-text.vtt,stream=text,init_segment=bear_text/init.mp4,segment_template=bear_text/$Number$.m4s' \
  --mpd_output bear.mpd --generate_static_live_mpd --start_segment_number 0
```
Please review and let me know.

packager/media/formats/packed_audio/packed_audio_writer_unittest.cc
```diff
@@ -42,8 +42,10 @@ const uint32_t kTimescale = 9000;
 const char kOutputFile[] = "memory://test.aac";
 // For multi-segment mode.
 const char kSegmentTemplate[] = "memory://test_$Number$.aac";
-const char kSegment1Name[] = "memory://test_1.aac";
-const char kSegment2Name[] = "memory://test_2.aac";
+const char kSegment124Name[] = "memory://test_124.aac";
+const char kSegment125Name[] = "memory://test_125.aac";
+const int64_t kSegmentIndex123 = 123;
```
kqyang, line 47: Can we start the segment index from 0 or 1?

packager/media/formats/webm/single_segment_segmenter.cc
```diff
@@ -21,9 +21,11 @@ SingleSegmentSegmenter::~SingleSegmentSegmenter() {}
 
 Status SingleSegmentSegmenter::FinalizeSegment(uint64_t start_timestamp,
                                                uint64_t duration_timestamp,
-                                               bool is_subsegment) {
+                                               bool is_subsegment,
+                                               int64_t segment_index) {
   Status status = Segmenter::FinalizeSegment(start_timestamp,
-                                             duration_timestamp, is_subsegment);
+                                             duration_timestamp, is_subsegment,
+					                         segment_index);
```
kqyang, line 28: Alignment

packager/media/formats/webm/single_segment_segmenter_unittest.cc
```diff
@@ -234,7 +240,7 @@ TEST_F(SingleSegmentSegmenterTest, LargeTimestamp) {
     ASSERT_OK(segmenter_->AddSample(*sample));
   }
   ASSERT_OK(segmenter_->FinalizeSegment(kLargeTimestamp, 5 * kDuration,
-                                        !kSubsegment));
+                           !kSubsegment, (kLargeTimestamp / (5 * kDuration))));
```
kqyang, line 243: alignment

packager/media/formats/webm/single_segment_segmenter_unittest.cc
```diff
@@ -270,7 +276,7 @@ TEST_F(SingleSegmentSegmenterTest, ReallyLargeTimestamp) {
     ASSERT_OK(segmenter_->AddSample(*sample));
   }
   ASSERT_OK(segmenter_->FinalizeSegment(kReallyLargeTimestamp, 5 * kDuration,
-                                        !kSubsegment));
+                     !kSubsegment, (kReallyLargeTimestamp / (5 * kDuration))));
```
kqyang, line 279: alignment

packager/media/formats/webm/webm_muxer.cc
```diff
@@ -80,7 +80,8 @@ Status WebMMuxer::FinalizeSegment(size_t stream_id,
   }
   return segmenter_->FinalizeSegment(segment_info.start_timestamp,
                                      segment_info.duration,
-                                     segment_info.is_subsegment);
+                                     segment_info.is_subsegment,
+				                     segment_info.segment_index);
```
kqyang, line 84: alignment

packager/mpd/base/representation.cc
```diff
@@ -271,8 +268,11 @@ base::Optional<xml::XmlNode> Representation::GetXml() {
   }
 
   if (HasLiveOnlyFields(media_info_) &&
-      !representation.AddLiveOnlyInfo(media_info_, segment_infos_,
-                                      start_number_)) {
+      !representation.AddLiveOnlyInfo(
```
kqyang, line 271: I'd update `AddLiveOnlyInfo` to remove the `start_number_` parameter instead.

packager/mpd/base/representation.h
```diff
@@ -223,9 +227,7 @@ class Representation {
   BandwidthEstimator bandwidth_estimator_;
   const MpdOptions& mpd_options_;
 
-  // startNumber attribute for SegmentTemplate.
-  // Starts from 1.
-  uint32_t start_number_ = 1;
+  bool stream_just_started_ = false;
```
kqyang, line 230: Is it part of this this PR?

packager/mpd/base/representation_unittest.cc
```diff
@@ -456,7 +456,10 @@ class SegmentTemplateTest : public RepresentationTest {
                    int repeat) {
     DCHECK(representation_);
 
-    SegmentInfo s = {start_time, duration, repeat};
+    if (start_segment_index_ == -1)
+      start_segment_index_ = start_time / duration;
```
kqyang, line 460: Is it necessary?

packager/media/chunking/chunking_handler.cc
```diff
@@ -151,6 +153,9 @@ Status ChunkingHandler::EndSegmentIfStarted() const {
   auto segment_info = std::make_shared<SegmentInfo>();
   segment_info->start_timestamp = segment_start_time_.value();
   segment_info->duration = max_segment_time_ - segment_start_time_.value();
+  segment_info->segment_index = current_segment_index_ +
+                                num_segments_before_last_cue_ +
+                                chunking_params_.start_segment_number - 1;
```
kqyang, line 158: There is a potential behavior change.

1. In the original code, the segment_index starts with 0 and is always monotonically increasing. 
2. In the new code, the segment_index is related to the segment timestamp.

Can we move the behavior change to a separate PR and focus this PR on start segment number?

So in this PR, we can just set the initial segment_index to chunking_params_.start_segment_number - 1 and increases by 1 for every new segment.

The comment applies to text_chunker below too.

packager/media/chunking/text_chunker.cc
```diff
@@ -14,8 +14,10 @@ namespace {
 const size_t kStreamIndex = 0;
 }  // namespace
 
-TextChunker::TextChunker(double segment_duration_in_seconds)
-    : segment_duration_in_seconds_(segment_duration_in_seconds){};
+TextChunker::TextChunker(double segment_duration_in_seconds,
+                         int64_t start_segment_number)
+    : segment_duration_in_seconds_(segment_duration_in_seconds),
+      start_segment_number_(start_segment_number){};
```
kqyang, line 20: needs to be reformatted

packager/media/event/combined_muxer_listener.h
```diff
@@ -42,7 +42,8 @@ class CombinedMuxerListener : public MuxerListener {
   void OnNewSegment(const std::string& file_name,
                     int64_t start_time,
                     int64_t duration,
-                    uint64_t segment_file_size) override;
+                    uint64_t segment_file_size,
+                    int64_t segment_index) override;
```
kqyang, line 46: Prefer `segment_number` so it matches with the file_name. What do you think? A number of files will need to be updated.

kqyang (commented):
Thanks for updating the code. It looks mostly good.

I left a number of comments, with most of them as formatting issues.

As I have suggested in your earlier PRs, you can use `clang-format` to format the code properly:

$ git clang-format <original commit>

If `HEAD^` is the original commit:

$ git clang-format HEAD^

packager/media/formats/packed_audio/packed_audio_writer_unittest.cc
sr1990: Can do that. Do you want to pass segment number to MediaHandlerTestBase::GetSegmentInfo ?

packager/mpd/base/representation.cc
sr1990: And getting the start_number in the method? 
```
bool RepresentationXmlNode::AddLiveOnlyInfo(
    const MediaInfo& media_info,
    const std::list<SegmentInfo>& segment_infos) {
  XmlNode segment_template("SegmentTemplate");

  int start_number = segment_infos.empty()? 
	  1 : segment_infos.begin()->start_segment_index;
```

packager/media/chunking/chunking_handler.cc
sr1990: Sure. Let's do that.

packager/media/event/combined_muxer_listener.h
sr1990: @kqyang Do you want to change segment_index in struct SegmentInfo, SegmentEventInfo, and kSegmentIndex in unit tests to kSegmentNumber as well? Or just the methods with file_name parameter in it?

packager/media/formats/packed_audio/packed_audio_writer_unittest.cc
kqyang: Ah, didn't notice that. It is fine to keep it as is then.

packager/mpd/base/representation.cc
kqyang: Or perhaps adding start_number only if segment_infos is not empty, e.g.

```
  if (media_info.has_segment_template_url() && !segment_infos.empty()) {
```

packager/media/event/combined_muxer_listener.h
kqyang: I think we should. There should be as few unnecessary "+1" or "-1" as possible. (If I have overlooked anything, let me know)

kqyang (commented):
Thanks.

packager/mpd/base/representation.h
sr1990: Removed it.

packager/mpd/base/representation_unittest.cc
sr1990: No. Removed it.

vish91:
Hey @sr1990  or @kqyang  just noticed that we are also implementing HLS for this flag. 
Would you know why or if that was a requirement for some scenario ? 
Not worried since it's an optional flag but curious because for HLS I assumed players always look for the #EXT-X-MEDIA-SEQUENCE tag in playlist.https://tools.ietf.org/html/rfc8216#section-4.3.3.2

sr1990:
@vish91 Looks like there is already a flag to specify EXT-X-MEDIA-SEQUENCE explicitly. 
https://github.com/bldsoft/shaka-packager/commit/2584ffb6b8bf63a3659d097794e432a3f34f44dc
Maybe we can set it's value to start_segment_number if the stream is HLS. What do you think @kqyang?

packager/media/chunking/chunking_handler.cc
```diff
@@ -31,6 +31,7 @@ bool IsNewSegmentIndex(int64_t new_index, int64_t current_index) {
 ChunkingHandler::ChunkingHandler(const ChunkingParams& chunking_params)
     : chunking_params_(chunking_params) {
   CHECK_NE(chunking_params.segment_duration_in_seconds, 0u);
+  segment_number_ = chunking_params.start_segment_number - 1;
```
kqyang, line 34: Why is it `start_segment_number - 1`?

It does not seem to be necessary.

Right now `GetSegmentName` is based on `segment_index`. It should be modified to be based on `segment_number`.

packager/mpd/base/mock_mpd_notifier.h
```diff
@@ -26,11 +26,12 @@ class MockMpdNotifier : public MpdNotifier {
...
   MOCK_METHOD2(NotifySampleDuration,
                bool(uint32_t container_id, uint32_t sample_duration));
-  MOCK_METHOD4(NotifyNewSegment,
+  MOCK_METHOD5(NotifyNewSegment,
                bool(uint32_t container_id,
                     uint64_t start_time,
                     uint64_t duration,
-                    uint64_t size));
+                    uint64_t size,
+                    int64_t segment_index));
```
kqyang, line 34: Should it be segment_number here?

packager/mpd/base/representation.cc
```diff
@@ -353,7 +351,8 @@ void Representation::AddSegmentInfo(int64_t start_time, int64_t duration) {
       } else {
         segment_infos_.push_back(
             {previous_segment_end_time,
-             actual_segment_end_time - previous_segment_end_time, kNoRepeat});
+             actual_segment_end_time - previous_segment_end_time, kNoRepeat,
+             segment_number + 1});
```
kqyang, line 355: Why is it +1?

packager/mpd/base/xml/xml_node.cc
```diff
@@ -432,9 +432,12 @@ bool RepresentationXmlNode::AddVODOnlyInfo(const MediaInfo& media_info) {
...
 bool RepresentationXmlNode::AddLiveOnlyInfo(
     const MediaInfo& media_info,
-    const std::list<SegmentInfo>& segment_infos,
-    uint32_t start_number) {
+    const std::list<SegmentInfo>& segment_infos) {
   XmlNode segment_template("SegmentTemplate");
+
+  int start_number =
+      segment_infos.empty() ? 1 : segment_infos.begin()->start_segment_number;
+
```
kqyang, line 440: This is only needed for line 459 below. I'd prefer just skipping the block at line 456 if segment_infos is empty.

kqyang:
> Would you know why or if that was a requirement for some scenario ?

@vish91 Can you clarify what you mean? The current change does not affect the media sequence number.

> @vish91 Looks like there is already a flag to specify EXT-X-MEDIA-SEQUENCE explicitly.
> [bldsoft@2584ffb](https://github.com/bldsoft/shaka-packager/commit/2584ffb6b8bf63a3659d097794e432a3f34f44dc)
> Maybe we can set it's value to start_segment_number if the stream is HLS. What do you think @kqyang?

Yes, there are overlaps. We could set the media sequence number to start_segment_number; we could also just keep them separate. I prefer having them separate for now to avoid affecting the current behavior.

packager/mpd/base/mock_mpd_notifier.h
sr1990: Yes.

packager/media/chunking/chunking_handler.cc
sr1990: `Why is it start_segment_number - 1?`
I was treating segment_number as segment_index. Will change it.

packager/mpd/base/xml/xml_node.cc
sr1990: It is also needed at line 465 as IsTimelineConstantDuration() argument.
Do you prefer having int start_number = ... in IsTimelineConstantDuration() method?

vish91:
> Yes, there are overlaps. We could set the media sequence number to start_segment_number; we could also just keep them separate. I prefer having them separate for now to avoid affecting the current behavior.

Yes agree. I prefer it separate as well.
> Would you know why or if that was a requirement for some scenario ? 
>  @vish91 Can you clarify what you mean? The current change does not affect the media sequence number.

I meant to ask do we even need a feature like this for HLS in first place ? since HLS uses the Media Sequence tag which already starts from 0 and isn't really affected by sequence ID on the segment name. 
Or to put it in other words do we have an example of what will this flag `start_segment_number` do when applied on an HLS packager command ?

packager/mpd/base/xml/xml_node.cc
sr1990: `I'd prefer just skipping the block at line 456 if segment_infos is empty.
`
@kqyang With the above change, how do you want to handle RepresentationClone unit test?

kqyang: I think we can add a call to AddSegments once after calling CopyRepresentation. Can you see what happens with that?

vish91:
@kqyang and @sr1990 please take a look when you have some time.

packager/mpd/base/xml/xml_node.cc
sr1990: @kqyang I just saw your reply. Sorry about the delay. AddSegments is specific to representation_ in SegmentTemplateTest. Do you want to create a new method AddingSegmentsToClone and a new data member cloned_representation?
@vish91 I am still looking into this. Will update the PR after that.

sr1990:
@kqyang I have updated the PR to change segment number as 1 based. Please review and let me know.

packager/mpd/base/xml/xml_node.cc
sr1990: I have not addressed this in the PR update.

ProIcons:
Looks good to me.  And it would be useful when the packager restarts for any reason to "continue" from where it left off.
Also a good other solution would be the ability to define the segment offsetting in unix epoch timestamps. 
While that would increase the manifest size, it would allow you in case of an interruption instead of resetting the offsets to 0 , to have them consecutively so you won't have to change periods on the manifest if you want to keep on streaming.

Vinayk93:
When can we expect to be merged.

joeyparrish:
This can't be merged as-is.  There are many merge conflicts, and that doesn't even account for the changes coming from the `cmake` branch when that work is complete.

Please be patient with us as we clean up and modernize the code, its dependencies, and its build system.  Then we can work with PR authors to fix/reimplement their changes and merge them.

cosmin:
This needs to be rebased, in particular changes in #934 for single file TS have changed some of the underlying code in ts_segmenter.cc

cosmin:
I rebased this and made a few updates and all tests seem to pass locally.

packager/app/test/packager_test.py
```diff
@@ -575,6 +575,9 @@ def _GetFlags(self,
     elif force_cl_index is False:
       flags += ['--noforce_cl_index']
 
+    if start_segment_number:
```
joeyparrish, line 578: Maybe `if start_segment_number is not None:` so that a start number of 0 is permitted.

packager/app/test/packager_test.py
```diff
@@ -823,6 +826,13 @@ def testForcedSubtitle(self):
                                                       output_hls=True))
     self._CheckTestResults('forced-subtitle')
 
+  def testDashStartNumber(self):
+    audio_video_streams = self._GetStreams(['audio', 'video'], segmented=True)
+    streams = audio_video_streams
+    self.assertPackageSuccess(streams, self._GetFlags(output_dash=True,
+                              start_segment_number='0'))
```
joeyparrish, line 833: With "is not None" above, this wouldn't have to be a string.

joeyparrish (approved):

PR #885, opened 2021-01-13
Fix wrong _create_bin_int implementation in pssh-box.py

qingquanwang:
The implementation is wrong for some value, which leads to wrong pssh box generation from pssh data.

Example: value = 132, below is the result in python3 shell. (python 3.6.9)

```
>>> value=132
>>> (chr(value >> 24) + chr((value >> 16) & 0xff) + chr((value >> 8) & 0xff) + chr(value & 0xff)).encode()
b'\x00\x00\x00\xc2\x84'
>>> import struct
>>> struct.pack('>i', value)
b'\x00\x00\x00\x84'
```
The previous implementation outputs 5 bytes which is wrong.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -45,3 +45,4 @@ Sergio Ammirata <sergio@ammirata.net>
 Thomas Inskip <tinskip@google.com>
 Tim Lansen <tim.lansen@gmail.com>
 Weiguo Shao <weiguo.shao@dolby.com>
+Qingquan Wang <wangqq1103@gmail.com>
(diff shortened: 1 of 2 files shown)


CONTRIBUTORS
```diff
@@ -45,3 +45,4 @@ Sergio Ammirata <sergio@ammirata.net>
 Thomas Inskip <tinskip@google.com>
 Tim Lansen <tim.lansen@gmail.com>
 Weiguo Shao <weiguo.shao@dolby.com>
+Qingquan Wang <wangqq1103@gmail.com>
```
kqyang, line 48: Can you order it in alphabetical order?

kqyang (commented):
Thanks for the fix. Just one comment.

CONTRIBUTORS
qingquanwang: Thanks. I have changed it.

kqyang (approved):

kqyang:
Thanks @qingquanwang. I have merged the change.

Issue #291, opened 2017-11-14
Support Single File TS for HLS


zshenker:
Support output of a single TS file for HLS media, instead of segmenting.
Use #EXT-X-BYTERANGE in the rendition manifests. May make sense to do this work at the same time as i-frame index playlists for HLS.

https://developer.apple.com/library/content/technotes/tn2288/_index.html#//apple_ref/doc/uid/DTS40012238-CH1-BYTE_RANGE_SUPPORT_FOR_SEGMENTS
@kqyang

joeyparrish:
Closed by #934

PR #891, opened 2021-01-29
[HLS] Support Single File TS for HLS.

sr1990:
This PR is related to https://github.com/google/shaka-packager/issues/291.
Example:
```
../packager 'in=TOS.ts,stream=video,output=tos_video.ts,playlist_name=tos_video.m3u8' \
            'in=TOS.ts,stream=audio,output=tos_audio.ts,playlist_name=tos_audio.m3u8' \
           --hls_master_playlist_output tos.m3u8 
```
          
Tested on Exoplayer.


packager/app/test/packager_test.py
```diff
@@ -1538,6 +1538,13 @@ def testEc3AndHlsSingleSegmentMp4Encrypted(self):
             ['audio', 'video'], hls=True, test_files=['bear-640x360-ec3.mp4']),
         self._GetFlags(encryption=True, output_hls=True))
     self._CheckTestResults('ec3-and-hls-single-segment-mp4-encrypted')
+  
+  def testHlsByterange(self):
```
kqyang, line 1542: I'd prefer calling it: testHlsSingleSegmentTs, as "hls-byte-range" could be for mp4 or ts.

packager/media/formats/mp2t/mp2t.gyp
```diff
@@ -33,12 +33,16 @@
         'mp2t_media_parser.h',
         'mpeg1_header.cc',
         'mpeg1_header.h',
+        'multi_segment_ts_segmenter.cc',
+	    'multi_segment_ts_segmenter.h',
```
kqyang, line 37: Can you align it with the line above it?

packager/packager.cc
```diff
@@ -219,23 +219,8 @@ Status ValidateStreamDescriptor(bool dump_stream_info,
...
-    // content. However, for TS all segments must be self-initializing so
-    // there cannot be an init segment.
-    if (stream.output.length()) {
-      return Status(error::INVALID_ARGUMENT,
-                    "All TS segments must be self-initializing. Stream "
-                    "descriptors 'output' or 'init_segment' are not allowed.");
-    }
-  } else if (output_format == CONTAINER_WEBVTT ||
+  
+  if (output_format == CONTAINER_WEBVTT ||
```
kqyang, line 223: The condition should include `output_format == CONTAINER_MPEG2TS ||` too, as TS does not have init segment. Can you also update line 232 too?

packager/media/formats/mp2t/ts_muxer.cc
```diff
@@ -21,7 +23,12 @@ Status TsMuxer::InitializeMuxer() {
   if (streams().size() > 1u)
     return Status(error::MUXER_FAILURE, "Cannot handle more than one streams.");
 
-  segmenter_.reset(new TsSegmenter(options(), muxer_listener()));
+  if (options().segment_template.empty()) {
```
kqyang, line 26: Prefer using the approach in PackedAudio: https://github.com/google/shaka-packager/tree/56908a83a76c85450d6b5834f3aa16e0660f4648/packager/media/formats/packed_audio.

Here is the summary:

- Having only one segmenter class instead of having one for SingleSegment and one for MultiSegment.
- The segmenter class writes the data into a temporary buffer first, which is already the case in the current TS segmenter implementation.
- Do not output the temporary buffer to the file in the segmenter, but in the muxer instead.
- In the muxer, the code can decide whether to write to a single file or a segment file based on the configuration. See https://github.com/google/shaka-packager/blob/56908a83a76c85450d6b5834f3aa16e0660f4648/packager/media/formats/packed_audio/packed_audio_writer.cc#L100 for an example.
- The muxer listener events are handled in the muxer. There is no need to handle that in the segmenter.

I think the code is cleaner and easier to maintain with this structure. 

What do you think?

kqyang (commented):
Thanks for working on this feature.

Had a suggestion on using PackedAudio approach instead, which can result in a better and simpler code.

See the inline comments for details.

packager/media/formats/mp2t/ts_muxer.cc
sr1990: Hi @kqyang, thanks for the review. Yes that makes sense. Let me try it out.

sr1990:
I had issues with this branch. 
So, I have created a new PR and branch at https://github.com/google/shaka-packager/pull/934

cosmin:
Replaced by #934

Issue #893, opened 2021-02-02
Reserved bits in PCR field in TS Adaptation Field not asserted


JPeMu:
## System info
**Operating System**: Linux Mint 19.3
**Shaka Packager Version**: v2.4.3-dd98700-release

## Issue and steps to reproduce the problem

**Packager Command**: 
packager-linux \
    'in=temp/shaka/u-1111-m-512x288-350.mp4,stream=video,playlist_name=temp/shaka/u-6400-m-512x288-350-1.m3u8,segment_template=temp/shaka/u-6400-m-512x288-350-1-$Number$.ts' \
    'in=temp/shaka/u-1111-a-128-1.mp4,stream=audio,playlist_name=temp/shaka/u-6400-a-128-1.m3u8,segment_template=temp/shaka/u-6400-a-128-1-$Number$.ts' \
    --hls_playlist_type live \
    --transport_stream_timestamp_offset_ms 1400 \
    --segment_duration 5.0 \
    --hls_master_playlist_output temp/shaka/u-6400-manifest.m3u8

Extra steps to reproduce the problem?
N/A

What is the expected result?
All generated TS segments should contain a PCR field in the TS Adaptation Field where the 6 reserved bits (between the base and extension) are set to '1' not '0' (As per ISO 13818-1 section 2.1.46 "reserved" ... Unless otherwise specified, all reserved bits shall be set to '1') and the Adapation Field spec section 2.3.4.5, table 2-6)

What happens instead?
The bits are set to '0' instead of '1'

<Please attach the input files or email to shaka-packager-issues@google.com.>
<Please provide any additional information below.>

In src/media/formats/mp2t/ts_packet_writer_util.cc in WriteAdaptationField() function, the local variable "pcr_last_bit_reserved_and_pcr_extension" takes only the LSB of the base << 15 and does not also assert the reserved bits.

I believe the code line should read:
    const uint16_t pcr_last_bit_reserved_and_pcr_extension =
        ((pcr_base & 1) << 15) | 0x7e00; // Assert reserved bits

JPeMu:
This is my first PR, so apologies for the unittest fails. I've also had to update the golden .ts files to accomodate the change made now I learned how to run the full test suite. Hope this is the correct approach here.

PR #895, opened 2021-02-02
Fix PCR reserved bits not being set correctly

JPeMu:
This is a PR to address the issue reported in #893

The PCR value inside the TS Packet's Adaptation Field (where present) should be represented as:

program_clock_reference_base (33 bits)
reserved (6 bits)
program_clock_reference_extension (9 bits)

```
(Bit Positions)
.... ...4 .... .... .3.. .... ...2 .... .... .1.. .... ....
7654 3210 9876 5432 1098 7654 3210 9876 5432 1098 7654 3210
BBBB BBBB BBBB BBBB BBBB BBBB BBBB BBBB Brrr rrrE EEEE EEEE
----- most_significant_32bits_pcr -----|pcr_last_bit_reserved_and_pcr_extension
            (uint32_t)                         (uint16_t)

B = Base, r = reserved, E = Extension
```

The reserved bits should always be set to '1' per ISO 13818-1 therefore OR'ing a value of 0x7e00 is required to set these reserved bits imho.

Tony


JPeMu:
This is my first PR here, so I apologise for the unittest fail back-and-forth. I've also had to update the golden files for .ts now that I understand how to run the full test suite as obviously this change affects some bits in the output files. Hope this is the correct approach?

kqyang (approved):

kqyang:
@JPeMu Thanks for the fix! And yes, that is the correct approach. 

I'll merge it once the build-bots completes successfully.

PR #892, opened 2021-02-01
Audible/mpd segment list

samidhtalsania:
This adds request made in https://github.com/google/shaka-packager/issues/455

The solution adds an additional flag for mpd generation called --dash_force_segment_list. If it is set, SegmentList would be used instead of SegmentBase and sidx atom would not be included in the generated audio file. When the flag is set, the intermediate media_info file is populated with subsegment ranges which is then used to create SegmentURLs within SegmentList. Without the flag set, the intermediate media_info is not populated with the subsegment ranges.

PR #897, opened 2021-02-03
Fix a bug in Mp4 segmenter where the last segment does not get the ri…

samidhtalsania:
Since segments are contiguous and start of new segment > end of previous segment, the previous segment's offset is subtracted by 1. This is not necessary for the last segment. This change fixes that and all the associated tests that were affected by this change.


samidhtalsania:
@kqyang I came across this bug while working on https://github.com/google/shaka-packager/pull/892. Since this is a small change and not related to the other PR i created a separate one.

kqyang:
> Since segments are contiguous and start of new segment > end of previous segment, the previous segment's offset is subtracted by 1. This is not necessary for the last segment. This change fixes that and all the associated tests that were affected by this change.

Sorry for the confusion. The start and end indicate the starting and ending index of the range, *inclusive*, so it does not matter whether it is the last segment or not. 

For example, for a mp4 with init segment size of 100, two segments with 100 each. 

The range for the first segment is: [100, 100 + 100 - 1].

The range for the second segment is: [200, 200 + 100 - 1].

Let me know if there are still confusions.

samidhtalsania:
That makes sense. Thanks! Closing this pr in that case.

Issue #900, opened 2021-02-16
Shaka packager seg faults if no media is feed


murillo128:
## System info
**Operating System**:  Ubuntu 20.4
**Shaka Packager Version**:  master head 95089593fc05af4c297e6edd02d70379a284cd78

## Issue and steps to reproduce the problem
I use shaka packager for live recording a streaming session. Sometimes, there are issues on the streaming and there is no data for a while causing several seg faults.

This happens both in mp4 and webm. Will send one PR for each.

PR #902, opened 2021-02-16
prevent seg fault if webm fragment is not initialized or last frame is EOS

murillo128:
as per #900

### DIFF
--- a/packager/media/formats/webm/webm_muxer.cc
+++ b/packager/media/formats/webm/webm_muxer.cc
@@ -47,7 +47,8 @@ Status WebMMuxer::InitializeMuxer() {
 }
 
 Status WebMMuxer::Finalize() {
-  DCHECK(segmenter_);
+  if (!segmenter_)
+    return Status::OK;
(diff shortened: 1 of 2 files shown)


packager/media/formats/webm/webm_muxer.cc
```diff
@@ -47,7 +47,8 @@ Status WebMMuxer::InitializeMuxer() {
 }
 
 Status WebMMuxer::Finalize() {
-  DCHECK(segmenter_);
+  if (!segmenter_)
```
kqyang, line 50: That happens only when Finalize is called without calling Initialize.

@murillo128 Have you checked why it happens?

murillo128: This seems to be happening when no media is received during the duration of the segmenter.

I am using shaka packager integrated into my media server to record media coming via streaming, so it may happen (due to streaming client malfunction, network disruption or reconnections) that there are periods of time in which media is not received.

kqyang (approved):

PR #907, opened 2021-03-05
Change to DEPS to potentially run on linux ARM 64

lovgrandma:
Running gclient on linux AWS arm64 architecture produced an error related to an unexpected architecture name. Run linux cmd uname -m
Change can potentially fix issue on some systems


lovgrandma:
Made a mistake heh. Meant to use my fork

PR #909, opened 2021-03-05
TTML: change "imagetype" attribute to camel case.

### LINKED ISSUE #908: TTML: wrong attribute "imagetype", camel case expected.
## System info
**Operating System**: Ubuntu 20.04
**Shaka Packager Version**: f0a52cbbf2

While testing some DVBSUB-to-TTML conversions with shaka packager, I could see at playback time the following message using [DASH reference player](https://reference.dashif.org/dash.js/v3.2.1/samples/dash-if-reference-player/index.html):
`    Error : SMPTE 2052-1:2013 defines the attribute name as "imageType" and does not define "imagetype"`

Checking out some text segment/chunk, I can see the attrubute (note the lowercase `t`):
```
$ cat text_5.mp4 
(...)
  <smpte:image imagetype="PNG" encoding="Base64" xml:id="img_3">
(...)
```

**Packager Command**:
```
packager \
'in=file4.ts,stream=0,init_segment=1080p_init.mp4,segment_template=1080p_$Number$.m4s' \
'in=file4.ts,stream=1,skip_encryption=1,init_segment=audio_init.mp4,segment_template=audio_$Number$.m4s,language=POL' \
'in=file4.ts,stream=2,skip_encryption=1,format=ttml+mp4,init_segment=text_init.mp4,segment_template=text_$Number$.mp4,language=ENG' \
--mpd_output test.mpd
```

**What is the expected result?:**
The same, but with that attribute in camel case format, as expected by the spec (which I didn't read: I'm trusting the player here).

**What happens instead?**
The error quoted above. 

Looks like an easy thing to fix.

Canta:
This should fix #908


kqyang:
Hey @Canta, thanks for fixing this bug. The PR has been merged.

PR #920, opened 2021-03-21
Update build_instructions.md

surajkumar-sk:
For windows running `GYP_DEFINES='target_arch=X64'` shows an error GYP_DEFINES is not recognized as an internal or external command.
the actual command is to set the value of GYP_DEFINES. so it should be `SET GYP_DEFINES='target_arch=X64'` After browsing through all the issues I found that docs were wrong so updating it.


kqyang (approved):
Thanks for the fix.

Issue #336, opened 2018-03-02
Make travis and appveyor run faster


kqyang:
We have a few failures due to test taking too long. There are two optimizations we can do to make travis and appveyor run faster:

- Run gtest unittests in parallel using https://github.com/google/gtest-parallel
- Run packager_test.py tests in parallel using pytest-xdict

joeyparrish:
I'm replacing Travis and Appveyor with a single GitHub Actions workflow.

Issue #318, opened 2018-01-19
Shared Library Build Broken on Windows


steji113:
## System info
**Operating System**: Windows 10 x64 1709
**Shaka Packager Version**: 435d69c3a6452cd868a09fa7c48e762004358664

## Issue and steps to reproduce the problem
```
set GYP_DEFINES=target_arch=x64 libpackager_type=shared_library
gclient runhooks
cd src
ninja -C out/Debug_x64
```

What is the expected result?
libpackager should compile as a DLL successfully.

What happens instead?
Errors such as the following occur:
```
[1/47] CXX obj\status.status.obj
FAILED: obj/status.status.obj 
ninja -t msvc -e environment.x64 -- "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\cl.exe" /nologo /showIncludes /FC @obj\status.status.obj.rsp /c ..\..\packager\status.cc /Foobj\status.status.obj /Fdobj\status.cc.pdb 
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(110): error C2143: syntax error: missing ')' before '('
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(110): error C2079: 'visibility' uses undefined class 'shaka::__attribute__'
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(110): error C2059: syntax error: ')'
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(110): error C2440: 'initializing': cannot convert from 'const char [8]' to 'int'
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(110): note: There is no context in which this conversion is possible
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(110): error C2146: syntax error: missing ';' before identifier 'Status'
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(110): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(111): error C2059: syntax error: 'public'
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(113): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
c:\users\stephen.daudell\documents\code\shaka_me\src\packager\status.h(113): error C2365: 'shaka::Status': redefinition; previous definition was 'data variable'
[...]

PR #957, opened 2021-06-15
Fix shared library test runs on Windows

joeyparrish:
Shared library builds worked, but failed tests because they were made
with the wrong CRT linker settings.  Strings allocated within the
library could not be freed outside the library because the dynamic CRT
was not used.

This sets necessary gyp variables to link with a dynamic CRT on
Windows, thereby fixing tests running in shared library mode that
otherwise hung in a GitHub Actions environment.

Related to #318 (shared_library builds on Windows)

Issue #336 (progress toward GitHub Actions workflow to replace Travis
  and Appveyor, where we need to build and test shared_library on all
  platforms)

b/190743862 (internal; tracking replacement of Travis)

### DIFF
--- a/packager/common.gypi
+++ b/packager/common.gypi
@@ -13,13 +13,29 @@
       # musl is a lightweight C standard library used in Alpine Linux.
       'musl%': 0,
     },
+
     'shaka_code%': '<(shaka_code)',
     'musl%': '<(musl)',
     'libpackager_type%': 'static_library',
+
     'conditions': [
       ['shaka_code==1', {
         # This enable warnings and warnings-as-errors.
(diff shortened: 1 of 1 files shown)


TheModMaker (approved):
Another option would be to stop using `<(component)` for internal components and just use it on `libpackager`.  Then we can just set component builds and get this the normal way.  But this works too.

joeyparrish:
> Another option would be to stop using <(component) for internal components and just use it on libpackager. Then we can just set component builds and get this the normal way. But this works too.

Good point.  I think for now, I'd like to do it this way.  I'd like us to replace gyp entirely in the next 12 months.

PR #959, opened 2021-06-16
CI overhaul based on GitHub Actions

### LINKED ISSUE #336: Make travis and appveyor run faster

joeyparrish:
This replaces Travis (for Linux & Mac) and Appveyor (for Windows) with
GitHub Actions.  In addition to using GitHub Actions to test PRs, this
also expands the automation of releases so that the only manual steps
are:

 1. Create a new CHANGELOG.md entry
 2. Create a release tag

Workflows have been created for building and testing PRs and releases,
for publishing releases to GitHub, NPM, and Docker Hub, and for
updating documentation on GitHub Pages.

When a new PR is created, GitHub Actions will:
 - Build and test on all combinations of OS, release type, and library
   type

Appveyor's workflow took ~2 hours, whereas the new GitHub Actions
workflow takes ~30 minutes.

When a new release tag is created, GitHub Actions will:
 - Create a draft release on GitHub
 - Extract release notes from CHANGELOG.md & attach them to the
   draft release
 - Build and test on all combinations of OS, release type, and library
   type, aborting if any build or test fails
 - Attach release artifacts to the draft release
 - Fully publish the draft release on GitHub
 - Publish the same release to NPM (triggered by GitHub release)
 - Publish the same release to Docker Hub (triggered by GitHub release)
 - Update the docs on GitHub pages (triggered by GitHub release)

Closes #336 (GitHub Actions workflow to replace Travis and Appveyor)

b/190743862 (internal; tracking replacement of Travis)

Change-Id: Ic53eef60a8587c5d1487769a0cefaa16eb9b46e7


.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,180 @@
...
+      - name: Test
+        shell: bash
+        run: |
+          # NOTE: Some of these tests must be run from the "src" directory.
+          cd src
+          if [[ '${{ matrix.os_name }}' == 'osx' ]]; then
+            export DYLD_FALLBACK_LIBRARY_PATH=out/${{ matrix.build_type }}${{ matrix.build_type_suffix }}
+          fi
+          for i in out/${{ matrix.build_type }}${{ matrix.build_type_suffix }}/*test${{ matrix.exe_ext }}; do "$i" || exit 1; done
+          python out/${{ matrix.build_type }}${{ matrix.build_type_suffix }}/packager_test.py -v --libpackager_type=${{ matrix.lib_type }}_library
```
TheModMaker, line 133: You may want to add `echo` statements to show which thing is being run.  I don't think the tests print which executable it is coming from.

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,180 @@
...
+            src/out/Release${{ matrix.build_type_suffix }}/mpd_generator${{ matrix.exe_ext }} \
+            artifacts/mpd_generator-${{ matrix.os_name }}${{ matrix.exe_ext }}
+          if [[ '${{ matrix.os_name }}' == 'win' ]]; then
+            (
+              cd src/out/Release${{ matrix.build_type_suffix }}
+              7z a ../../../pssh-box.py.zip pyproto pssh-box.py
+            )
+            mv pssh-box.py.zip artifacts/pssh-box-${{ matrix.os_name }}.py.zip
+          else
+            tar -czf pssh-box.py.tar.gz -C src/out/Release${{ matrix.build_type_suffix }} pyproto pssh-box.py
```
TheModMaker, line 153: Can't you just put the output in the correct location?  Or would that affect the output layout?

.github/workflows/draft_github_release.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Compute ref
+        # We could be building from a workflow dispatch (manual run)
+        # or from a pushed tag.  If triggered from a pushed tag, we would like
+        # to strip refs/tags/ off of the incoming ref and just use the tag
+        # name.  Subsequent steps can refer to $TARGET_REF to determine the
+        # correct ref in all cases.
+        run: |
+          echo "TARGET_REF=${{ github.event.inputs.tag || github.ref }}" | \
+              sed -e 's@refs/tags/@@' >> $GITHUB_ENV
```
TheModMaker, line 44: Can't you just do `TARGET_REF=refs/tags/{{...}}`?

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,180 @@
...
+            mv pssh-box.py.tar.gz artifacts/pssh-box-${{ matrix.os_name }}.py.tar.gz
+          fi
+
+      - name: Attach artifacts to release
+        if: matrix.build_type == 'Release' && matrix.lib_type == 'static' && github.event.inputs.release_id
+        uses: dwenegar/upload-release-assets@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.SHAKA_BOT_TOKEN }}
```
TheModMaker, line 161: I'd suggest adding a README with what these secrets are, where they come from, and how to update them.

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,180 @@
...
+          mv \
+            src/out/Release${{ matrix.build_type_suffix }}/packager${{ matrix.exe_ext }} \
+            artifacts/packager-${{ matrix.os_name }}${{ matrix.exe_ext }}
+          mv \
+            src/out/Release${{ matrix.build_type_suffix }}/mpd_generator${{ matrix.exe_ext }} \
+            artifacts/mpd_generator-${{ matrix.os_name }}${{ matrix.exe_ext }}
+          if [[ '${{ matrix.os_name }}' == 'win' ]]; then
+            (
+              cd src/out/Release${{ matrix.build_type_suffix }}
+              7z a ../../../pssh-box.py.zip pyproto pssh-box.py
```
TheModMaker, line 149: I think this also requires other files for the protobuf parsing.

.github/workflows/build_and_test.yaml
joeyparrish: Good call.  If this is all executed as one shell script, then `set -x` would do the trick nicely.

.github/workflows/build_and_test.yaml
joeyparrish: You're right, I could.  And I guess I probably should.  I was initially porting the original travis configs, which did it this way.  I just never stopped to re-evaluate that.  I'll make that change and test it on my fork to make sure it's working.

.github/workflows/draft_github_release.yaml
joeyparrish: No, you've got it backwards.  I'm stripping off the `refs/tags/` part from the input to get the tag name only.  I suppose this needs a more local comment.

.github/workflows/build_and_test.yaml
joeyparrish: I double-checked, and both the tarball and zip versions of this are complete.  The protos are all in the `pyproto/` folder.

.github/workflows/build_and_test.yaml
joeyparrish: Done.  I added a README.md file to the workflows folder.  PTAL!

joeyparrish:
I made changes to address your feedback and tested them on my fork.  Everything appears to be working well still.  PTAL!

joeyparrish:
I've refactored this after learning more about GitHub Actions.  The workflows are better organized now, using custom actions as units of reusability for build & test.  PTAL!

joeyparrish:
Made one more change to move hard-coded package names out to GitHub Secrets to make it easier to test in a fork without accidentally releasing packages.  PTAL!

TheModMaker (approved):

.github/workflows/README.md
```diff
@@ -0,0 +1,49 @@
...
+ - `custom-actions/build-packager`:
+   Builds Shaka Packager.  Leaves build artifacts in the "artifacts" folder.
+   Requires OS-dependent and build-dependent inputs.
+ - `custom-actions/test-packager`:
+   Tests Shaka Packager.  Requires OS-dependent and build-dependent inputs.
+ - `custom-actions/build-docs`:
+   Builds Shaka Packager docs.
+
+## Workflows
+ - On PR:
```
kqyang, line 13: Does it apply to non-release commits too?

It seems to be redundant for PRs, but it will be useful for commits submitted by the internal Gerrit (the test configs for the Gerrit is not exactly the same as the test configs here).

.github/workflows/README.md
```diff
@@ -0,0 +1,49 @@
...
+     docs.
+ - On release tag:
+   - `github_release.yaml`:
+     Creates a draft release on GitHub, builds and tests all combinations of OS
+     & build settings, builds docs on all OSes, attaches static release binaries
+     to the draft release, then fully publishes the release.
+ - On release published:
+   - `docker_hub_release.yaml`:
+     Builds a Docker image to match the published GitHub release, then pushes it
+     to Docker Hub.
```
kqyang, line 25: It will be useful to publish the latest ToT image to Docker Hub too with :latest tag on all commits. We used to do that but it was stopped due to permission issues about 2 years ago. With GitHub actions, it seems like we can do it again.

.github/workflows/custom-actions/build-docs/action.yaml
```diff
@@ -0,0 +1,46 @@
...
+  Only runs in Linux due to the dependency on doxygen, which we install with
+  apt.
+
+runs:
+  using: composite
+  steps:
+    - name: Install dependencies
+      shell: bash
+      run: |
+        echo "::group::Install dependencies"
```
kqyang, line 15: Just a nit: Can we enable the echoing of the commands instead of using custom messages? I think it is more readable here and it is also more useful for debugging.

.github/workflows/custom-actions/build-packager/action.yaml
```diff
@@ -0,0 +1,109 @@
...
+    description: A suffix to append to the build type in the output path.
+    required: false
+    default: ""
+  exe_ext:
+    description: The extension on executable files.
+    required: false
+    default: ""
+
+runs:
+  using: composite
```
kqyang, line 27: Just to confirm, for composite runs, the next step will not run if the current step fails, right?

kqyang (commented):
Thanks. Just a few nits and comments.

.github/workflows/README.md
joeyparrish: We can run workflows on push of new commits, too, if we wish.  It can be the same workflow we use for PRs, or it can be an independent workflow with many of the same steps.  Let's discuss and decide if we want to do this or not.

.github/workflows/README.md
joeyparrish: Absolutely.  Right now, `:latest` is being set on the new release images.

.github/workflows/custom-actions/build-docs/action.yaml
joeyparrish: I tried that first, but I found that for long-running commands with lots of output, groups were more useful in the GitHub UI to review what happened.  Locating a command echoed with set -x in the middle of 10k lines of output is hard.  Finding a group and expanding it is easy.

.github/workflows/custom-actions/build-packager/action.yaml
joeyparrish: Correct.

.github/workflows/README.md
kqyang: And related to this: the build status needs to be updated: https://github.com/google/shaka-packager/blame/master/README.md#L11. I don't know if GitHub Actions provide a similar feature.

joeyparrish: GitHub Actions does offer badges.  Having build and test status badges for ToT would suggest that we should have a build-and-test workflow run on pushes, too.  I'll take care of that.

PR #971, opened 2021-08-04
fix: misleading log output when HLS target duration updates (fixes #969)

### LINKED ISSUE #969: Misleading log output
Have I misunderstood something or is there an error in this log output?

https://github.com/google/shaka-packager/blob/6c8ad30217c286d4eecadc9df12420767d389942/packager/hls/base/media_playlist.cc#L513

Shouldn't it be the other way around? Like:
```c++
    VLOG(1) << "Updating target duration from " << target_duration_ << " to "
            << target_duration;
```

Not a big issue, but still a bit misleading. Can create a PR if you can confirm my understanding.

MarcusWichelmann:
Minor fix. The parameters were swapped.

### DIFF
--- a/packager/hls/base/media_playlist.cc
+++ b/packager/hls/base/media_playlist.cc
@@ -513,8 +513,8 @@ void MediaPlaylist::SetTargetDuration(int32_t target_duration) {
   if (target_duration_set_) {
     if (target_duration_ == target_duration)
       return;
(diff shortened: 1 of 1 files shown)


MarcusWichelmann:
@googlebot I signed it!

cosmin (approved):

PR #973, opened 2021-08-04
feat: Add support for the EXT-X-START tag

### LINKED ISSUE #970: Add support for the EXT-X-START tag
The EXT-X-START tag would allow to specify the location where the player starts playing, either from start (positive value) or from end (negative value).
This is especially useful in case of livestreams where this tag can be used to set the target latency of the playback.

https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.5.2

I'm working on a PR.

MarcusWichelmann:
This PR adds an optional `--hls_start_time_offset` parameter, that, when used, adds `EXT-X-START` tags to the HLS media playlists.

The EXT-X-START tag allows to specify the location where the player starts playing, either from start (positive value) or from end (negative value).
This is especially useful in case of livestreams where this tag can be used to set the target latency of the playback.

Reference: https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.5.2

The RFC says, that the `EXT-X-START` tag could also be added to the master playlist, but my tests have shown that most players only respect it when it's in the media playlists.

Fixes #970


joeyparrish:
I know this is very old, but we have added it to the list of things we'd like to finish and merge before v3.0.0 is released.  I will rebase this and make some modifications before merging.

joeyparrish:
I am also adding some unit tests.

joeyparrish:
LGTM, but I made all the changes to update it and write the tests... @cosmin, what do you think?

cosmin (approved):

PR #975, opened 2021-08-10
fix: set array_completeness in HEVCDecoderConfigurationRecord correctly

MarcusWichelmann:
ISO/IEC 14496-15 says about the `HEVCDecoderConfigurationRecord`:
> **array_completeness** when equal to 1 indicates that all NAL units of the given type are in the following
array and none are in the stream; when equal to 0 indicates that additional NAL units of the indicated type may be in the stream; the default and permitted values are constrained by the sample entry name;

This PR sets `array_completeness` to 0 if parameter NAL units may appear in the stream when they are not stripped by `--strip_parameter_set_nalus`.

This should increase player-compatibiltity for streams with mid-stream SAR changes.

### DIFF
--- a/packager/media/codecs/h265_byte_to_unit_stream_converter.cc
+++ b/packager/media/codecs/h265_byte_to_unit_stream_converter.cc
@@ -75,22 +75,22 @@ bool H265ByteToUnitStreamConverter::GetDecoderConfigurationRecord(
   buffer.AppendInt(static_cast<uint8_t>(3) /* numOfArrays */);
 
   // More parameter set NALUs may follow when strip_parameter_set_nalus is disabled.
-  const uint8_t kArrayCompleteness = strip_parameter_set_nalus() ? 0x80 : 0;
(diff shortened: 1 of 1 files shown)


packager/media/codecs/h265_byte_to_unit_stream_converter.cc
```diff
@@ -72,8 +72,10 @@ bool H265ByteToUnitStreamConverter::GetDecoderConfigurationRecord(
   buffer.AppendInt(static_cast<uint8_t>(kUnitStreamNaluLengthSize - 1));
   buffer.AppendInt(static_cast<uint8_t>(3) /* numOfArrays */);
 
+  // More parameter set NALUs may follow when strip_parameter_set_nalus is disabled.
+  const uint8_t kArrayCompleteness = strip_parameter_set_nalus() ? 0x80 : 0;
```
joeyparrish, line 76: Style nit: Since this is no longer a compile-time constant, it should be renamed to something like array_completeness.

joeyparrish (changes requested):

cosmin (approved):

PR #976, opened 2021-08-12
Insert EXT-X-DISCONTINUITY when decoder config parameters change

MarcusWichelmann:
When a HLS stream has mid-stream changes in its decoding parameters, like an aspect ratio switch, this change is not always applied by some players (iOS, VLC).

This PR fixes this behaviour by inserting a `#EXT-X-DISCONTINUITY` into the media playlists every time a decoder config change happens and SPS/PPS/.. NALUs are sent.

See [RFC8216 Section 4.3.2.3](https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.3):
> The EXT-X-DISCONTINUITY tag SHOULD be present if there is a change in any of the following characteristics:
>   -  encoding parameters
>   -  encoding sequence

I'm not yet sure how this should play together with the `--strip_parameter_set_nalus` parameter. Maybe the discontinuity should only be injected if the SPS NALUs are not stripped?

### DIFF
--- a/packager/hls/base/hls_notifier.h
+++ b/packager/hls/base/hls_notifier.h
@@ -81,6 +81,9 @@ class HlsNotifier {
   /// @return true on success, false otherwise.
   virtual bool NotifyCueEvent(uint32_t stream_id, int64_t timestamp) = 0;
 
+  /// Called when the decoder config parameters have changed.
+  virtual void NotifyDecoderConfigChanged(uint32_t stream_id) = 0;
+
   /// @param stream_id is the value set by NotifyNewStream().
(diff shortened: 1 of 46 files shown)


MarcusWichelmann:
This PR doesn't fully solve the issue. There is a probability for a race condition where more discontinuieties are inserted into the audio playlist than they are into the video playlists. This causes playback zu fail.

MarcusWichelmann:
This is superseded by https://github.com/MarcusWichelmann/shaka-packager/tree/discontinuity-on-parameters-change-2. But I'm not planning to create a PR for that because it introcudes many controversial changes and could need some cleanups.

PR #978, opened 2021-08-12
ci: Produce static release executables on Linux

### LINKED ISSUE #965: Standalone/Static packager-linux
## System info
**Operating System**: Linux (RedHat, CentOS)
**Shaka Packager Version**: 2.5.0

Is it possible that you provide or building the standalone/static packager-linux with static libraries included without requiring external libraries for execution?

**Packager Command**:

(1) Copy the packager-linux in a directory on RedHat
(2) Try to execute. 

What is the expected result?

Packager is executing.

What happens instead?

error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory

joeyparrish:
We never produced static release executables on Linux before, but the dynamic libraries they depended on were universal enough that nobody noticed.  Now that we have released v2.5 and switched to GitHub Actions for CI builds, the Linux executables depend on libatomic, which is causing issues for some users.

Although we can't create fully-static executables on macOS or Windows, we can at least do so on Linux.

This adds a GYP variable `static_link_binaries` which can be set to request full-static binaries on Linux.  This also exposes the Chromium build variable `disable_fatal_linker_warnings`, which is necessary when static linking on Linux due to static-link-related warnings generated by libcurl for its use of `getaddrinfo`.  Finally, this enforces the definition of `__UCLIBC__` with static linking on Linux, which is the only way to disable malloc hooks in Chromium base.  Those hooks cause linker failures when linking statically on Linux.

A new check has been added to the release workflow to ensure that the builds we create are statically linked on Linux.

Closes #965

### DIFF
--- a/packager/common.gypi
+++ b/packager/common.gypi
@@ -12,12 +12,18 @@
       'shaka_code%': 0,
       # musl is a lightweight C standard library used in Alpine Linux.
       'musl%': 0,
+      # This is a flag from build/common.gypi to allow linker warnings.
+      # This may be necessary with static_link_binaries=1.
+      'disable_fatal_linker_warnings%': '0',
       'libpackager_type%': 'static_library',
+      'static_link_binaries%': '0',
     },
 
     'shaka_code%': '<(shaka_code)',
     'musl%': '<(musl)',
+    'disable_fatal_linker_warnings%': '<(disable_fatal_linker_warnings)',
     'libpackager_type%': '<(libpackager_type)',
+    'static_link_binaries%': '<(static_link_binaries)',
 
     'conditions': [
       ['shaka_code==1', {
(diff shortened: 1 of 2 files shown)


packager/common.gypi
```diff
@@ -122,6 +128,22 @@
...
       }],
+      ['static_link_binaries==1', {
+        'conditions': [
+          ['OS=="linux"', {
+            'defines': [
+              # Even when we are not using musl or uClibc, pretending to use
+              # uClibc on Linux is the only way to disable certain Chromium
+              # base features, such as hooking into malloc.  Hooking into
+              # malloc, in turn, fails when we are linking statically.
+              '__UCLIBC__',
```
kqyang, line 139: Is there side effects by defining `__UCLIBC__`, i.e. effects other than disabling malloc hooking?

packager/common.gypi
```diff
@@ -122,6 +128,22 @@
...
+            ],
+          }],
+        ],
+        'ldflags': [
+          '-static',
```
kqyang, line 144: What is the binary size impact of the "fully" static build?

kqyang (approved):
LGTM with two questions.

packager/common.gypi
joeyparrish: Answering this question, I found out that we have not been stripping the release binaries.  So I'll be adding that to the workflow, as well.

Here are the sizes from Linux arm64 as an example:

| libpackager | build config | exe linking | stripped | size |
| ----- | ----- | ----- | ----- | ----- |
| static | release | dynamic | no | 6.7MB |
| static | release | dynamic | yes | 5.1MB |
| static | release | static | no | 8.5MB |
| static | release | static | yes | 5.8MB |

So the increase for stripped builds moving to static linking is 700kB, but since we didn't strip the executables before, the size will actually go down by 900kB once we do.

kqyang: Ah, thanks!

kqyang (approved):

PR #985, opened 2021-08-30
Fix for gap size warning in Low Latency mode

CaitlinOCallaghan:
## The issue
- With LL-DASH mode enabled, the gap size warning was hit and printed to the console every time a new segment was registered to the manifest.
- This occurred because the first chunk's size and duration were being stored for each segment, rather than the full segment size and duration. Note, only the first chunk's metrics are known at first because in low latency mode, the segment is registered to the manifest before it is finished being processed and written.
- Because of this, the gap size check was comparing the end time of the first chunk in the previous segment to the beginning time of the current segment, causing the check to fail every time.

## The Fix
- Update a low latency segment's duration and size once the segment file has been fully written.
- The full segment size and duration will be used to update the bandwidth estimator and the segment info list. 
- Updating the segment info list to hold the full duration is necessary for satisfying [the gap size check found in Represenation.cc](https://github.com/google/shaka-packager/blob/master/packager/mpd/base/representation.cc#L391).
- NOTE: bandwidth estimation is currently only used in HLS

## Testing
- `./mpd_unittest --gtest_filter="SegmentTemplateTest.OneSegmentLowLatency"`
image

- manual
image

### DIFF
--- a/packager/media/event/combined_muxer_listener.cc
+++ b/packager/media/event/combined_muxer_listener.cc
@@ -77,6 +77,13 @@ void CombinedMuxerListener::OnNewSegment(const std::string& file_name,
   }
 }
 
+void CombinedMuxerListener::OnCompletedSegment(int64_t duration,
+                                               uint64_t segment_file_size) {
+  for (auto& listener : muxer_listeners_) {
+    listener->OnCompletedSegment(duration, segment_file_size);
+  }
+}
+
 void CombinedMuxerListener::OnKeyFrame(int64_t timestamp,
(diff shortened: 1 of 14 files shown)


joeyparrish (approved):

joeyparrish:
Thank you!

CaitlinOCallaghan:
Thank you for your help and support Joey!

joeyparrish:
My pleasure!

kevleyski:
Thanks again Caitlin

PR #1012, opened 2021-12-27
Hex encode key id for FairPlay.

wjywbs:
Instead of base64.


joeyparrish:
Wouldn't this be a breaking change for someone?  Why is hex correct?  Why is base64 wrong?

joeyparrish:
Closing due to inactivity.

PR #1020, opened 2022-01-14
feat: HLS / DASH support forced subtitle

### LINKED ISSUE #988: HLS / DASH support for forced subtitles
Does shaka-packager support setting forces subtitles in HLS and DASH.  Mainly DASH ? 

In HLS it's the stream media tag `FORCED`
and in DASH its a ROLE `forced-subtitles`

https://github.com/Dash-Industry-Forum/DASH-IF-IOP/issues/165

https://github.com/Dash-Industry-Forum/Ingest/issues/40

Based on the first issue it looks like Shaka-player and ExoPlayer do support it for HLS and DASH

vish91:
Closes #988


vish91:
> If I understand this correctly, to get forced-subs and output both DASH & HLS, one would have to indicate a dash role for forced subs **and** set the forced attribute to get the same effect in HLS?

@joeyparrish  yeah I sync'd this branch with master to re-run actions and see if we can get some logs from the builds. 
meanwhile some background on this is in here: https://github.com/shaka-project/shaka-packager/issues/988
Yeah dash does need support for a new role for this

joeyparrish:
This seems fine, but I would strongly prefer it (from a usability standpoint) if `forced=1` implied adding `forced-subtitle` to `dash_roles`.  I don't think you should have to use both if you're already setting `forced=1` for HLS.

vish91:
> This seems fine, but I would strongly prefer it (from a usability standpoint) if `forced=1` implied adding `forced-subtitle` to `dash_roles`. I don't think you should have to use both if you're already setting `forced=1` for HLS.

@joeyparrish  yeah we discussed that here #988 . The problem is dash has roles field to signal vs HLS doesn't have anything like that so far. Hence the differentiation between how to trigger these for HLS vs DASH.

joeyparrish:
> The problem is dash has roles field to signal vs HLS doesn't have anything like that so far. Hence the differentiation between how to trigger these for HLS vs DASH.

I understand, but why couldn't you check for the HLS field in the DASH code, and insert the role automatically?

vish91:
> > The problem is dash has roles field to signal vs HLS doesn't have anything like that so far. Hence the differentiation between how to trigger these for HLS vs DASH.
> 
> I understand, but why couldn't you check for the HLS field in the DASH code, and insert the role automatically?

yeah we could technically. Was discussing it with KQ initially, but that drifts us away for this being a DASH role from the spec prespective and users having to change their command to support a new parameter instead of actually using the dash_role param for setting the role to `forced-subtitle`.

vish91:
> I'm not suggesting you drop the dash_role part. But if you're generating both HLS & DASH in one command, why not say that `forced=1` implies the appropriate role on the DASH side?

hmm.. would you mind sharing an example for this.. I wanna see it and test it before I can say we can change it or not. This is an input paramter and not packager parameter so thinking out loud here 
would you want to allow a user to pass either dash_role or forced on the input parameter and be able to support this feature ?

vish91:
@joeyparrish  any examples or sample commmand on what you were explaining ?

vish91:
@joeyparrish  if you wanna take a pass at this PR , I updated it with the main branch so all the changes are in here

cosmin:
I think we should call the stream descriptor forced_subtitle and have it be generic for both dash and hls, if set it would apply the dash role (in addition to being able to directly set the dash_roles manually), set FORCED=YES for HLS, and validate it's only supplied on subtitle streams. 

And remove all the various hls only referendes in the code.

vish91:
@cosmin  @joeyparrish  yeah sure i am ok with that, however in that case what happens when someone doesn't set stream_descriptor and has just set the DASH role value in stream_descriptor ? Will it fail silently or error ? 
So for packager behavior will we ask users to ensure that they set `dash_role=forced-subtitle` and `forced_subtitle=1` to achieve success for DASH and for HLS they only have to set `forced_subtitle=1`. 

If that would be the expectation we can go that route. I might not have enough time to rework this PR by tonight. Might need some time to look into DASH flow for that.

joeyparrish:
@cosmin wrote:
> I think we should call the stream descriptor forced_subtitle and have it be generic for both dash and hls, if set it would apply the dash role (in addition to being able to directly set the dash_roles manually), set FORCED=YES for HLS, and validate it's only supplied on subtitle streams.

@vish91 wrote:
> So for packager behavior will we ask users to ensure that they set `dash_role=forced-subtitle` and `forced_subtitle=1` to achieve success for DASH and for HLS they only have to set `forced_subtitle=1`.

I think what Cosmin is saying is that `forced_subtitle=1` in the stream descriptor would imply the appropriate DASH role.  If no `dash_role=` is given, `forced_subtitle=1` would set `dash_role=forced-subtitle` internally.  `dash_role` is low-level, `forced_subtitle` is high-level.  For those who just want forced subs, the high level flag should be enough.  @cosmin, am I interpreting you correctly?

cosmin:
@joeyparrish that's correct

@vish91  I can make the updates if you're ok with it or can wait if you prefer to do it.

vish91:
yep go for it please. 🙇🏽‍♂️  Just one note if you do can you please adjust the documentation as well. I think we missed changing the  force_cl_index as well so was meaning to update that as well in here.

cosmin:
Made the updates to do this as we discussed with `forced_subtitle` stream descriptor being the high level way of doing this, and having it set role internally for dash if needed.

packager/app/test/testdata/forced-subtitle/bear-english-text.vtt
```diff
@@ -0,0 +1,11 @@
+WEBVTT
+
+STYLE
+::cue { color:lime }
+
+00:00:00.000 --> 00:00:00.800 align:center
+Yup, that's a bear, eh.
+
+00:00:01.000 --> 00:00:04.700 align:center
+He 's... um... doing bear-like stuff.
```
joeyparrish, line 10: :rofl:

packager/hls/base/media_playlist.cc
```diff
@@ -373,6 +373,10 @@ void MediaPlaylist::SetCharacteristicsForTesting(
   characteristics_ = characteristics;
 }
 
+void MediaPlaylist::SetForcedSubtitleForTesting(const bool forced_subtitle) {
+  forced_subittle_ = forced_subtitle;
```
joeyparrish, line 377: typo: forced_subittle => forced_subtitle

joeyparrish (changes requested):
Looks great otherwise

joeyparrish (approved):

PR #1050, opened 2022-03-17
fix: PTS diverge DTS when DTS close to 2pow33 and PTS more than 0

### LINKED ISSUE #1049: [MPEGTS] Special case when unrolled DTS and PTS diverge
## System info
**Operating System**: all
**Shaka Packager Version**:  2.4.x

mpeg-ts has 33bit timestamps. After reaching 2^33 it starts from (close to) 0 values. It's known in Shaka so during parsing mpeg-ts unrolling can occur. It happens when PTS and DTS are about to pass 2^33 boudary. Next values read from mpeg-ts are unrolled that is they are continued and incremented beyond 2^33 boundary for DTS to be monotonic increasing.

In very rare case there can be a situation when PTS is very close to 2^33 (right before passing the boudary) and PTS already passed so it has a value a bit above 0. This causes for mpeg-ts parser to unroll only DTS so it next values are very big (and above 2^33) and PTS remain small, right above 0. This causes lot's of problems including:
* `PTS - DTS`  is very big 
* `PTS < DTS` as PTS is close to 0 and DTS is more than 2^33

During mpeg-ts reading, in this special case PTS should be forced to unroll to maintain PTS close to  DTS.

I've prepared a PR #1050  fixing this.

**Packager Command**:

(1) Parse attached bear-640x360_ptszero_dtswraparound.ts

It has set DTS close to 2^33 and PTS a bit above 0. I had to attach zip as you did not enable 'ts' files.

What is the expected result?
Properly parse.

What happens instead?

<Please attach the input files or email to shaka-packager-issues@google.com.>
<Please provide any additional information below.>

zdanek:
Fixes #1049


packager/media/formats/mp2t/mp2t_media_parser_unittest.cc
```diff
@@ -163,6 +172,22 @@ TEST_F(Mp2tMediaParserTest, TimestampWrapAround) {
   EXPECT_GT(video_max_dts_, static_cast<int64_t>(1) << 33);
 }
 
+TEST_F(Mp2tMediaParserTest, PtsZeroDtsWrapAround) {
+  // "bear-640x360.ts" has been transcoded from bear-640x360.mp4 by applying a
```
joeyparrish, line 176: It appears you quote the wrong file name.  This should be "bear-640x360_ptszero_dtswraparound.ts".

joeyparrish (changes requested):
Looks good otherwise.  I'll make a quick correction to the test comment and merge it.

joeyparrish (approved):

zdanek:
Thank you.

PR #1055, opened 2022-04-22
feat: Solution for adding startwithSAP/subsegmentstartswithSAP for audio tracks

XingzhaoYun:
This pull request adds startwithSAP/subsegmentstartswithSAP for aac, ac3, ec3 and ac4 audio tracks according to LIVE or VOD profile.

### DIFF
--- a/packager/mpd/base/adaptation_set.cc
+++ b/packager/mpd/base/adaptation_set.cc
@@ -274,6 +274,14 @@ base::Optional<xml::XmlNode> AdaptationSet::GetXml() {
       return base::nullopt;
     }
   }
+  if (subsegment_start_with_sap_ == 1) {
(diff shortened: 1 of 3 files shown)


packager/mpd/base/adaptation_set.cc
```diff
@@ -274,6 +274,14 @@ base::Optional<xml::XmlNode> AdaptationSet::GetXml() {
       return base::nullopt;
     }
   }
+  if (subsegment_start_with_sap_ == 1) {
```
vish91, line 277: could you please add some tests and potentially since this is for spec compliance all the existing golden tests will also be regenrated once you can run those tests as well. 

assuming this is for audio only and hence the subsegment check for just value == 1 ?

joeyparrish:
Please merge with `main`, which has changed a lot in the recent `cmake` port.  Please also add/update tests.  Thanks!

PR #1092, opened 2022-08-31
feat: Add xHE-AAC support

geoffjukes:
Note:
* An xHE-AAC capable encoder will auto adjust the user-specified SAP/RAP
  value to the allowed grid where SAP/RAPs can occur.
  e.g.: `-rapInterval 5000` (5s) may result in actual SAPs/RAPs every 4.984s.
* To ensure SAP/RAP starts a new segment, Shaka needs to executed with a
  "--segment_duration" is less than or equal to that adjusted value.
* If every SAP/RAP should trigger a new segment, just set the segment
  length to a very low value e.g.: `--segment_duration 0.1`

### DIFF
--- a/packager/media/codecs/aac_audio_specific_config.cc
+++ b/packager/media/codecs/aac_audio_specific_config.cc
@@ -129,7 +129,7 @@ bool AACAudioSpecificConfig::Parse(const std::vector<uint8_t>& data) {
     extension_frequency_ = kSampleRates[extension_frequency_index];
   }
   
-  if (audio_object_type_ == 42) {
+  if (audio_object_type_ == AOT_USAC) {
     return frequency_ != 0 && num_channels_ != 0 && channel_config_ <= 7;
   } else {
     return frequency_ != 0 && num_channels_ != 0 && audio_object_type_ >= 1 &&
@@ -209,7 +209,7 @@ uint8_t AACAudioSpecificConfig::GetNumChannels() const {
 bool AACAudioSpecificConfig::ParseAudioObjectType(BitReader* bit_reader) {
   RCHECK(bit_reader->ReadBits(5, &audio_object_type_));
   
-  if (audio_object_type_ == 31) {
+  if (audio_object_type_ == AOT_ESCAPE) {
     uint8_t audioObjectTypeExt;
     RCHECK(bit_reader->ReadBits(6, &audioObjectTypeExt));
     audio_object_type_ = static_cast<AudioObjectType>(32 + audioObjectTypeExt);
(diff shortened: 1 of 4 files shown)


cosmin:
I verified that this patch does package xhe-aac sample files from https://www2.iis.fraunhofer.de/AAC/xhe-aac-abr.html whereas without this patch the packaging errors.

packager/media/codecs/aac_audio_specific_config.cc
```diff
@@ -205,6 +206,18 @@ uint8_t AACAudioSpecificConfig::GetNumChannels() const {
   return num_channels_;
 }
 
+bool AACAudioSpecificConfig::ParseAudioObjectType(BitReader* bit_reader) {
+  RCHECK(bit_reader->ReadBits(5, &audio_object_type_));
+  
+  if (audio_object_type_ == 31) {
```
joeyparrish, line 212: Should we maybe use the constant name here?  (Is this AOT_ESCAPE?)

packager/media/codecs/aac_audio_specific_config.cc
```diff
@@ -131,10 +128,14 @@ bool AACAudioSpecificConfig::Parse(const std::vector<uint8_t>& data) {
     RCHECK(extension_frequency_index < arraysize(kSampleRates));
     extension_frequency_ = kSampleRates[extension_frequency_index];
   }
-
-  return frequency_ != 0 && num_channels_ != 0 && audio_object_type_ >= 1 &&
-         audio_object_type_ <= 4 && frequency_index_ != 0xf &&
-         channel_config_ <= 7;
+  
+  if (audio_object_type_ == 42) {
```
joeyparrish, line 132: Should we maybe use the constant name here?  (Is this AOT_USAC?)

joeyparrish (changes requested):
Looks good otherwise.  Thanks!

packager/media/codecs/aac_audio_specific_config.cc
geoffjukes: there are other integer references in the file (lines 286, 290, 295, 296, for example). While I'm in there, would you like me to switch those to constants?

geoffjukes:
@joeyparrish Did the changes come through. TBH I've not often contributed to projects, and so I'm not sure I did it correctly.

geoffjukes:
@joeyparrish Apologies for bugging you. Is there a timeline for when this might get merged (assuming it's acceptable)? I have a downstream vendor that depends on the Shaka packager, and we can't use their service until it has xHE-AAC support.

cosmin:
It does look like the requested changes have been addressed. +1 for merging this.

geoffjukes:
@joeyparrish Feedback from a downstream application that uses Shaka Packager is that it works for them, but they are unable to merge the patch with their code until this PR is merged.

joeyparrish:
Apologies for dropping this.  I'll review tomorrow (Tuesday, US/Pacific time).

joeyparrish (approved):

PR #1096, opened 2022-09-08
correct to following ISO BMFF

AUGxhub:
fourcc for mp4 container is defined in ISO BMFF  
all packger should follow it


joeyparrish:
Please note:
> Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

Beyond the CLA, it looks like you're changing the definition of the constant, but that constant is used in both reading inputs and writing outputs.  If the old output it wrong, we would need a second constant for the new fourcc, and to accept both versions on input.

It would also be useful to have a reference to the ISO BMFF spec where it says that the fourcc for vp9 should be capitalized, even though the precedent for other fourccs seems to be lowercase.

Thanks!

joeyparrish:
Closing due to inactivity.

PR #1098, opened 2022-09-09
feat(HTTP): Retry failed HTTP requests with exponential backoff

### LINKED ISSUE #1095: Implement exponential backoff for HTTP PUT
A feature request to track adding an exponential backoff strategy for HTTP PUTs.
https://github.com/shaka-project/shaka-packager/blob/31129eed64d60372435403fd33073e126caaa74b/packager/file/http_file.cc#L212-L213
Not backing is sometimes problematic, as seen in here: https://github.com/shaka-project/shaka-streamer/pull/103#issuecomment-1111246197

mariocynicys:
This PR implements exponential backoff retrials for http files.

I am not quite familiar with the code base & how packager works, so I could use some help in clarifying the following:
- In case of PUT or POST requests, the `upload_cache_` would be drained, right? Does this mean we have to keep a copy of it?
- Are there any other cases we should perform retries in? Other than client timeouts and these specific response code.

Tests yet to be added.
Closes #1095

### DIFF
--- a/packager/file/http_file.cc
+++ b/packager/file/http_file.cc
@@ -6,6 +6,8 @@
 
 #include <packager/file/http_file.h>
 
+#include <thread>
+
 #include <absl/flags/declare.h>
 #include <absl/flags/flag.h>
 #include <absl/log/check.h>
@@ -54,6 +56,10 @@ namespace {
 
 constexpr const char* kBinaryContentType = "application/octet-stream";
 constexpr const int kMinLogLevelForCurlDebugFunction = 2;
+
+// The number of times to retry requesting the server before we give up.
+constexpr const int kNumRequestErrorRetries = 5;
+constexpr const int kFirstRetryDelayMilliseconds = 1000;
 
 size_t CurlWriteCallback(char* buffer, size_t size, size_t nmemb, void* user) {
   IoCache* cache = reinterpret_cast<IoCache*>(user);
@@ -135,13 +141,9 @@ int CurlDebugCallback(CURL* /* handle */,
 
 class LibCurlInitializer {
  public:
-  LibCurlInitializer() {
-    curl_global_init(CURL_GLOBAL_DEFAULT);
-  }
+  LibCurlInitializer() { curl_global_init(CURL_GLOBAL_DEFAULT); }
 
-  ~LibCurlInitializer() {
-    curl_global_cleanup();
-  }
+  ~LibCurlInitializer() { curl_global_cleanup(); }
 
   LibCurlInitializer(const LibCurlInitializer&) = delete;
   LibCurlInitializer& operator=(const LibCurlInitializer&) = delete;
@@ -223,9 +225,6 @@ bool HttpFile::Open() {
     return false;
   }
   // TODO: Try to connect initially so we can return connection error here.
-
-  // TODO: Implement retrying with exponential backoff, see
-  // "widevine_key_source.cc"
 
   ThreadPool::instance.PostTask(std::bind(&HttpFile::ThreadMain, this));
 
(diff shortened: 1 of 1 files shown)


packager/file/http_file.cc
```diff
@@ -333,8 +334,39 @@ void HttpFile::SetupRequest() {
...
+
+  // Try sending the request multiple times if it fails.
+  for (int i = 0; i < kNumRequestErrorRetries; ++i) {
+    res = curl_easy_perform(curl_.get());
+
+    if (res == CURLE_HTTP_RETURNED_ERROR) {
+      long res_code = 0;
+      curl_easy_getinfo(curl_.get(), CURLINFO_RESPONSE_CODE, &res_code);
+      if (res_code == 408 || res_code == 429 || res_code >= 500) {
+        // FIXME: Should we clear download_cache_ and upload_cache_ here??
```
joeyparrish, line 348: Yes, you'll need to ensure that the inside of the loop has no side-effects.  I haven't analyzed to see what the caches do here, but it's worth looking into more closely.

You may want to write test cases that verify the correct behavior for POST/PUT with data.  In the `cmake` branch, you'll find HTTP file tests are all enabled and working.  (In `main`, they are all disabled.)  You should be able to verify retry behavior with httpbin, and see that the final failed retry still has a response that reflects the data from the POST/PUT.

I recommend you try the `cmake` branch in a separate clone, without using gclient.  We have yet to update the docs for that, but this is what you'll want to try:

```sh
# Clone
git clone ...
# Use cmake branch
git checkout -b cmake upstream/cmake
# Get submodules
git submodule init && git submodule update
# Gen build files
cmake -S . -B build/ -DCMAKE_BUILD_TYPE=Debug
# Build
cmake --build build/ --config Debug --parallel
# Test
(cd build && ctest -C Debug -V)
```

If you have time, you could make a PR against the `cmake` branch, which is where most development is happening right now as we rewrite the build system and dependencies.

joeyparrish (changes requested):

joeyparrish:

cosmin:
This has been rebased to `main` and the tests do pass, but from some digging into the implications of download and upload cache I'm not sure whether this can be done safely. For example let's think of the case of using this to upload packaging output to some HTTP server.

As soon as CURL calls `CurlReadCallback` the underlying circular buffer is going to notify potential writers that were blocked on writing to the circular cache that there is room to write. And those writers may start writing right away, overwriting the data in the cache. Meanwhile the request fails and needs to be retried, but the underlying circular buffer was altered so the request cannot be retried.

I don't think this can be done safely with the currently IoCache interface as the backing buffers. I believe the data from the upload_cache would have to written to some other intermediary buffer (without triggering read events) so that curl and retry safely, and only upon success should the IoCache circular buffer signal the read event and advance the pointers. 

Likewise for download cache, only upon successful download should the data be written to `download_cache` otherwise a consumer may have already consumed the bad response body by the time we retry.

Alternatively to avoid the memory copies I think we'd have to take a lock on IoCache around the `curl_easy_perform` loop and use some lower level interfaces than `Read()` and `Write()` (perhaps something like `MaybeRead()` and `MaybeWrite()`) and then explicitly release the lock and trigger read event once the request was successful.

mariocynicys:
@cosmin I am fiddling with this but I can see this is not fixable via `IoCache` (please correct me if I am wrong). `CurlReadCallback` might be called multiple times during a single request, but if the request fails we start all over again and not from where we left of. This means that the circular buffer must keep the data written on it since the start of the request, and if the circular buffer size is less than the request body size, then we have some cases where we must overwrite part of our `upload_cache_` and if we the request fails after this point, we can't retry it.

For `download_cache_` we shouldn't want readers to read before the request is successful. If part of the response was read but `curl_easy_perform` fails, then we should try again and seek to the very same byte we stopped reading at and continue from there? This might be problematic if different data is returned.

Copying/backing up the data elsewhere until the request is successful would work fine, but it wouldn't be memory efficient if we are dealing with big files/requests.

PR #1110, opened 2022-10-21
ci: Fix repo settings for self-hosted runners and debugging

joeyparrish:
The hack of using "secrets" to store per-repo settings was not working.  The main reason is that pull_request workflows don't have access to secrets no matter what you do.  So it was impossible to make this work for settings like "ENABLE_SELF_HOSTED" for PR tests.

This change replaces that old hack with a new one.  Now a repo owner must create a "GitHub Environment" with the name of the setting they want to enable.  Currently supported values are "self_hosted", to add self-hosted runners to the build/test matrix, and "debug", to start an SSH server for debugging when a workflow fails.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -60,7 +60,11 @@
    pushing bogus packages to NPM during CI testing from a fork
    - In a fork, set to a private name which differs from the production one
 
-## Optional Repo Secrets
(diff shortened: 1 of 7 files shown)


joeyparrish:
Note that because the settings were not working, we did not realize the arm64 build workflow was broken.  There will be a follow-up PR with a fix for that.

theodab (approved):

Issue #1047, opened 2022-03-09
Switch to CMake and git submodules


joeyparrish:
The use of our ancient chromium build infrastructure has been increasingly problematic.  We use old versions of chromium components so that they are compatible with gyp (gyp has been removed upstream), but this means also an older version of depot_tools, and lack of support for newer OSes, SDKs, and Linux distros.

We should replace gyp with CMake, gclient with submodules, and depot_tools should go away completely.  Dependencies (like protobuf, etc) should be updated to the latest versions, and if possible, system versions should be used instead of building from source.

I'm not sure what order these should be done in, or how much work they will be.  I'll experiment with it a bit to figure that out, but I may or may not be the one to finish the work.  Volunteers are welcome!

joeyparrish:
Oh, and of course, some of the old Python build scripts in our old forks of Chromium repos will only run with Python 2, which is not even installed by default any more in many places.

joeyparrish:
In a pure python3-environment, such as my workstation at the office, I can't even run gyp any more.  So replacing gyp is a high priority in this list.

Replacing gclient with submodules is relatively easy and straightforward.  See https://github.com/joeyparrish/shaka-packager/commit/f4d560ed924c54c7ec97389d19bd5b5d25f02b9e  Some of those deps could be dropped by the time this work is complete (such as gyp, packager/build).

zdanek:
@joeyparrish I managed to fix some problems with generating CMakeLists.txt from gyp.
It was discussed some time ago at https://github.com/shaka-project/shaka-packager/pull/774
I wished to make a PR to fix issues with generation of CMakeLists.txt but I failed when discovered that some tools are part of Chromium ecosystem and is very hard to fix because those tools were from ancient revision of Chromium tools.

But still - I can tell how to fix locally some files in order to generate CMakeLists.txt.

And by the way - I use Python3 with Shaka gyp and it works (AFAIK I had to fix only about two places regarding py2->py3)

joeyparrish:
@zdanek, thank you!  You shouldn't need to make permanent changes to gyp.  You could make temporary changes, generate CMakeLists.txt from gyp, then clean up CMakeLists.txt by hand and commit it.  The goal is to get rid of gyp, so it should be fine to hack it up as needed if it is easier than writing CMakeLists.txt completely from scratch.

zdanek:
@joeyparrish I got it. 

[...]

Issue #346, opened 2018-03-07
Switch to abseil (https://github.com/abseil/abseil-cpp) as base library


kqyang:
Right now we use base library from Chromium and use GYP as the meta build system. There are a few problems:

- Chromium base is not really designed to be a common base library. There are a few things enabled by default, which we do not really need. 
- GYP is already deprecated in Chromium two years ago. Because of this, we cannot update our base library without switching to Chromium's new meta build system: GN.

This issue tracks the effort to switch to abseil as base library and bazel as the meta build system for Shaka Packager. It will also enable us to add a CMAKE build.

joeyparrish:
As part of this work, C++14 will now be required to build Shaka Packager.  This will reduce our dependencies, and we will rely more on the standard library.

joeyparrish:
If you would like to help with this effort, please see https://github.com/shaka-project/shaka-packager/issues/1047#issuecomment-1282870835 to volunteer.  Thanks!

joeyparrish:
There are no longer any references to chromium base in the `cmake` branch.  Once we clean up documentation, logging, and some automated workflows, we can merge the `cmake` branch into `main`.  But since the absl porting away from base is complete, I'm closing this issue.

PR #1116, opened 2022-10-24
feat: port media/test

joeyparrish:
Rewrite test_data_util.cc to locate files relative to the source file itself, rather than using a service from chromium `base::`.

Issue #1047 (CMake porting)
Issue #346 (absl porting)


kevleyski:
Ah great, thanks Joey

joeyparrish:
This is working now.  After this lands, I have another PR ready with the crypto code from media/base, which needed a lot of changes.

theodab (approved):

packager/media/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Copyright 2022 Google Inc. All rights reserved.
```
zdanek, line 1: what is right business name for Google? LLC or Inc? as I see both in different files

packager/media/test/test_data_util.cc
```diff
@@ -4,41 +4,42 @@
...
-  std::string buffer;
-  CHECK(base::ReadFileToString(GetTestDataFilePath(name), &buffer));
-  return std::vector<uint8_t>(buffer.begin(), buffer.end());
+  std::filesystem::path path = GetTestDataFilePath(name);
+
+  std::vector<uint8_t> data;
+  data.resize(std::filesystem::file_size(path));
+
+  FILE* f = fopen(path.string().c_str(), "rb");
+  if (!f) {
```
zdanek, line 33: shall we use [errno](https://cplusplus.com/reference/cerrno/errno/) and/or [perror](https://cplusplus.com/perror)?

packager/media/test/test_data_util.cc
```diff
@@ -4,41 +4,42 @@
...
+    LOG(FATAL) << "Failed to read test data from " << path;
+    return std::vector<uint8_t>();
```
zdanek, line 35: if you'd move `data.resize` to line 37 you could return here `data` as it would be already created and empty
still resizing makes sense in line 37 as we are sure that file open was successful

zdanek (changes requested):
please consider my notices

packager/media/CMakeLists.txt
joeyparrish: It should be LLC.  We did a mass-rename operation in this branch not long ago, and this slipped back in.  A minor copy-paste bug.  :-)

packager/media/test/test_data_util.cc
joeyparrish: We certainly could.  I don't think it's important at the level of a test data file.  This is a case of "should not fail", and is only used in tests, not production code.

packager/media/test/test_data_util.cc
joeyparrish: Moved.  Thanks!

packager/media/CMakeLists.txt
joeyparrish: Fixed!

PR #1119, opened 2022-10-25
feat: Port AES and RSA crypto to mbedtls

joeyparrish:
mbedtls works very differently from BoringSSL, and many changes had to be made in the details of AES decryption to accomodate this.

Beyond the basic changes required for mbedtls, part of the CTS padding implementation had to be rewritten.  I believe this is because of an assumption that held for BoringSSL, but not for mbedtls.  I was unable to determine what it was, so I rewrote the CTS decryption using reference materials.  After this, tests passed.

The deterministc PRNG I used with mbedtls in the RSA tests differs somewhat from the old one, so the expected vectors had to be regenerated.  The old determinstic tests were also disabled, and are now re-enabled.

Since cryptography is sensitive code, and because there were far more changes needed here than just updating some headers and utility function calls, this has been split into its own PR for separate review from the rest of the media/base porting work.

Issue #1047 (CMake porting)
Issue #346 (absl porting)


kqyang:
OOC, what is the reason of switching to mbedtls? Is [boringssl's cmake](https://github.com/google/boringssl/blob/master/CMakeLists.txt) not working?

joeyparrish:
mbedtls was significantly lighter, easier to build, and easier to integrate through CMake.  boringssl required both go and perl to generate sources, and it was difficult to make it work with libcurl's CMake configs (at the time).  I evaluated the other TLS libraries supported by libcurl, and landed on mbedtls, which we are already using in our cross-platform static binaries for ffmpeg (https://github.com/shaka-project/static-ffmpeg-binaries).

NOTE: I see that libcurl's CMake build now has support for boringssl (added in July), but that was not the case when I began this project.

joeyparrish:
I've fixed some minor, platform-specific compiler errors, and now I'm debugging a crash in the RSA tests on macOS & Windows.  I'll keep changes in fresh commits and avoid a force-push so that the review can still proceed in the meantime.  Thanks!

joeyparrish:
The crashing test has been fixed.

joeyparrish:
I also ran these tests under valgrind to ensure that we're not doing anything stupid.  In an early revision of this work, I had overflowed some buffers with mbedtls by not reading the docs carefully enough.  :-)

zdanek:
@joeyparrish I built it on Ubuntu 20.04, cmake 3.16.3
All test passed
```
4: [----------] Global test environment tear-down
4: [==========] 19 tests from 2 test suites ran. (0 ms total)
4: [  PASSED  ] 19 tests.
4/4 Test #4: status_unittest ..................   Passed    0.00 sec

100% tests passed, 0 tests failed out of 4

```

packager/media/base/test/rsa_test_data.cc
```diff
@@ -326,122 +326,125 @@ const char kTestMessage[] =
     "A fool thinks himself to be wise, but a"
     " wise man knows himself to be a fool.";
 
+// Self-generated test vector.  Used to verify algorithm stability.
 const uint8_t kTestEncryptedMessage_3072[] = {
```
kqyang, line 330: Can you keep the original encrypted message to verify it can still be decrypted successfully? The same for the signature verification. It can confirm "partially" that the algorithm is compatible with the original implementation.

packager/media/base/aes_encryptor.cc
```diff
@@ -76,7 +68,14 @@ bool AesCtrEncryptor::CryptInternal(const uint8_t* plaintext,
 
   for (size_t i = 0; i < plaintext_size; ++i) {
     if (block_offset_ == 0) {
-      AES_encrypt(&counter_[0], &encrypted_counter_[0], aes_key());
+      size_t ignored_output_size;
+      uint8_t unused_iv[AES_BLOCK_SIZE];
+      CHECK_EQ(
+          mbedtls_cipher_crypt(&cipher_ctx_, unused_iv, AES_BLOCK_SIZE,
```
kqyang, line 74: nit: Does it work by setting iv to NULL and iv_size to 0? mbedtls implementation seems to ignore the IV for ECB mode anyway: https://siliconlabs.github.io/Gecko_SDK_Doc/mbedtls/html/cipher_8h.html#a1c249f6ee1a0d2c906927c7790c41dc5 and https://github.com/Mbed-TLS/mbedtls/issues/1091.

packager/media/base/aes_cryptor.cc
```diff
@@ -30,20 +26,21 @@ namespace shaka {
...
+  mbedtls_cipher_free(&cipher_ctx_);
+}
 
 bool AesCryptor::Crypt(const std::vector<uint8_t>& text,
                        std::vector<uint8_t>* crypt_text) {
   // Save text size to make it work for in-place conversion, since the
   // next statement will update the text size.
   const size_t text_size = text.size();
-  crypt_text->resize(text_size + NumPaddingBytes(text_size));
+  // mbedtls requires an extra block's worth of output buffer available.
```
kqyang, line 42: Is it a bug? Is it mentioned in the mbedtls documentation?

I glanced through the mbedtls_cipher_crypt implementation quickly but I don't see that an extra block is needed for ECB and CBC modes. (But I may have overlooked it)

Also, I am not sure if it works everywhere it is used. It might not always be possible to require the output size to be larger than the input size (especially for in-place encryption / decryption).

packager/media/base/aes_decryptor.cc
```diff
@@ -112,44 +115,49 @@ bool AesCbcDecryptor::CryptInternal(const uint8_t* ciphertext,
...
+  // Decrypt the last two blocks.
+  internal_iv_ = backup_iv;
+  // mbedtls requires a buffer large enough for one extra block.
+  std::vector<uint8_t> final_output_blocks(AES_BLOCK_SIZE * 3);
+  CbcDecryptBlocks(reconstructed_blocks.data(), AES_BLOCK_SIZE * 2,
+                   final_output_blocks.data());
+
+  // Copy the final output.
+  memcpy(next_to_last_plaintext_block, final_output_blocks.data(),
+         AES_BLOCK_SIZE + residual_block_size);
```
kqyang, line 160: Just two nits (feel free to ignore):

1. The original implementation decrypts one less block: the new implementation decrypts the residual block twice while the old implementation decrypts the residual block only once.
2. CTS is only needed for Widevine classic packaging, which is already deprecated.

kqyang (commented):
That is a lot of changes!

Mostly looks good to me with a few comments!

packager/media/base/aes_cryptor.cc
joeyparrish: It is mentioned in the documentation.  I overlooked it initially, but after valgrind reported overflows, I re-read the docs and found it there.

https://github.com/Mbed-TLS/mbedtls/blob/55fd0b9fc1a2a8e0720c8ab15e06cea579aa4f4b/include/mbedtls/cipher.h#L1035-L1037

```
 * \param output        The buffer for the output data. This must be able to
 *                      hold at least `ilen + block_size`. 
```

packager/media/base/aes_decryptor.cc
joeyparrish: I understand.  I couldn't make the original implementation work, though, possibly due to mbedtls treating overlapping input/output buffers differently, or the way it treats IV updates differently.

I never did figure out what was wrong with my attempt to port the old implementation more directly, but this re-implementation from descriptions of CTS in Wikipedia allowed me to create this, which passes all the original tests.  I'm inclined to keep it for now, though, at least until we choose to separately remove Widevine Classic support.  (It's still in there, right?)

packager/media/base/aes_encryptor.cc
joeyparrish: I thought it was required, but I double-checked, and it's not.  This might have been left over from some older revision where I had the cipher context set up wrong.

Removed.  Thanks!

packager/media/base/test/rsa_test_data.cc
joeyparrish: No, because I can't make the PRNG from mbedtls produce _exactly_ the same outputs as the one from OpenSSL.  Even with the same seed data, their outputs are different.

kqyang: Sorry for not making it clear. 

What I meant is to keep the original data but don't perform the data verification on that data.

Let's say the original gold encrypted data is called kOpensslGoldEncryptedData.

There can be two tests:

Test_1:
- Verify that kOpensslGoldEncryptedData can be decrypted and the decrypted message matches kTestMessage.

Test_2a:
- Verify that kTestMessage can be encrypted and then decrypted and the decrypted message matches kTestMessage.

Test_2b:
- Verify that the kTestMessage can be encrypted and the encrypted message matches the new gold encrypted data and then the encrypted message can be decrypted and the decrypted message matches kTestMessage.

Either test_2a or test_2b can be used.

packager/media/base/rsa_key.cc
```diff
@@ -132,45 +150,85 @@ bool RsaPrivateKey::GenerateSignature(const std::string& message,
...
-  int signature_size = RSA_private_encrypt(
-      padded_digest.size(), &padded_digest[0],
-      reinterpret_cast<uint8_t*>(&(*signature)[0]), rsa_key_, RSA_NO_PADDING);
 
-  if (signature_size != static_cast<int>(rsa_size)) {
-    LOG(ERROR) << "RSA private encrypt failure: " << ERR_error_string(
-                                                         ERR_get_error(), NULL);
+  // Because we set the digest mode in the previous step, here we use
+  // MBEDTLS_MD_NONE to signal that we're passing the raw message, which
+  // mbedtls should hash for us using MBEDTLS_MD_SHA1 (set above).
```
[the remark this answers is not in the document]
joeyparrish: mbedtls_rsa_set_padding docs say:

https://github.com/Mbed-TLS/mbedtls/blob/55fd0b9fc1a2a8e0720c8ab15e06cea579aa4f4b/include/mbedtls/rsa.h#L165-L168

```
 * \note           The hash selected in \p hash_id is always used for OEAP
 *                 encryption. For PSS signatures, it is always used for
 *                 making signatures, but can be overridden for verifying them.
 *                 If set to #MBEDTLS_MD_NONE, it is always overridden.
```

I don't believe you're looking at the correct source repo.  Here's the implementation:

https://github.com/Mbed-TLS/mbedtls/blob/development/library/rsa.c#L1679

packager/media/base/aes_cryptor.cc
kqyang: I see. Thanks for confirming! It also says that "This must not be the same buffer as \p input.", which I assume it means that it does not support in-place encryption/decryption. I think we do use in-place encryption / decryption in a few places.

packager/media/base/test/rsa_test_data.cc
joeyparrish: I see.  Yes, I could do that.  I'm out of time for today, but I'll make that change soon.

packager/media/base/aes_decryptor.cc
kqyang: Yah, it is a nit. Feel free to ignore.

> (It's still in there, right?)

IIRC, there are still a few long tail use cases, but we don't expect new contents to be packaged. The plan is to fully deprecate it by the end of 2022 or in early 2023.

packager/media/base/aes_cryptor.cc
joeyparrish: We do, but I've been able to verify that they still work.  Some did not, and were replaced.  I believe it depends on the cipher mode in practice, even though the header docs state that it is not supported in general.  All the tests work, and valgrind is happy.

packager/media/base/aes_decryptor.cc
joeyparrish: OK, sounds good.  Thanks!

packager/media/base/rsa_key.cc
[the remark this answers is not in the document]
kqyang: The implementation in https://github.com/Mbed-TLS/mbedtls/blob/development/library/rsa.c#L1679 is the same: 

It does not perform the first hashing: "Generate the message digest: mHash = Hash(M)", so it still has to be done outside of mbedtls_rsa_rsassa_pss_sign_ext just like in the original implementation.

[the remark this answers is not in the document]
joeyparrish: You were right!  When I reverted the test vectors, I found that I couldn't verify the old signatures without doing the hash explicitly first.  Fixed now.

packager/media/base/test/rsa_test_data.cc
joeyparrish: Done!  This exposed my mistake in the signature code, and I fixed it by doing an explicit hash first.

kqyang (approved):

PR #1122, opened 2022-10-27
ci: Fix main workflows

joeyparrish:
This brings some workflow improvements and fixes from the `cmake` branch to `main`, as well as some unique fixes to keep gclient working, so that we can continue to accept contributions in `main` until the `cmake` merge is ready.

 - Fix docs build in GitHub Actions (from `cmake` branch)
 - Cancel workflow when a PR is updated (from `cmake` branch)
 - Fix docker failures caused by running as root (from `cmake` branch)
 - Work around exception in depot_tools on Windows
 - Use Windows 2019 images in GitHub Actions for compatibility with gyp
 - Remove Docker build on ArchLinux, which no longer supports python2 at all
   - (NOTE: The `cmake` branch is still building on ArchLinux.  Docker builds for Arch will be restored to the `main` branch when the `cmake` branch is finally merged to `main`.)


.github/workflows/build_and_test.yaml
```diff
@@ -42,7 +49,7 @@ jobs:
         # NOTE: macos-10.15 is required for now, to work around issues with our
         # build system.  See related comments in
         # .github/workflows/custom-actions/build-packager/action.yaml
-        os: ["ubuntu-latest", "macos-10.15", "windows-latest", "self-hosted-linux-arm64"]
+        os: ["ubuntu-latest", "macos-10.15", "windows-2019", "self-hosted-linux-arm64"]
```
theodab, line 52: Much like how you have a comment for why we use `macos-10.15`, you should have a comment for why we use `windows-2019`.

theodab (changes requested):

.github/workflows/build_and_test.yaml
joeyparrish: Good point.  Done.

theodab (approved):

PR #1127, opened 2022-11-02
feat: port media/base

joeyparrish:
This removes all chromium dependencies from media/base/ and completes the build system in CMake.

The ClosureThread class and its classes were removed, as they were specific to chromium base.  ClosureThread has been replaced by std::thread.

The byte-swapping utilities in network_util.cc have been removed and replaced with absl.

generate_unique_temp_path() was split out of file_unittest.cc into file_test_util.cc, where other test suites could make use of it.

WARN_UNUSED_RESULT was replaced with the C++ standard attribute [[nodiscard]].

The base::Clock interface was replaced with a typedef for a function pointer that returns the current time.

This re-enables the tests in http_key_fetcher_unittest.cc by using httpbin.org.

Issue #1047 (CMake porting)
Issue #346 (absl porting)


packager/media/base/container_names.cc
```diff
@@ -1726,41 +1727,44 @@ MediaContainerName DetermineContainer(const uint8_t* buffer, int buffer_size) {
 
 MediaContainerName DetermineContainerFromFormatName(
     const std::string& format_name) {
-  if (base::EqualsCaseInsensitiveASCII(format_name, "aac") ||
-      base::EqualsCaseInsensitiveASCII(format_name, "adts")) {
+  std::string normalized_format_name = format_name;
+  std::transform(format_name.begin(), format_name.end(),
+                 normalized_format_name.begin(), ::tolower);
+
+  if (normalized_format_name == "aac" || normalized_format_name == "adts") {
```
theodab, line 1734: Now that this is no longer using a custom comparison method, couldn't it now be a `switch` statement?

theodab (changes requested):

joeyparrish:
This includes the changes in PR #1132 and #1133.  With that, builds are (finally) all passing.

theodab (approved):

packager/media/base/macros.h
```diff
@@ -46,9 +46,7 @@
 #endif
 
 #ifndef FALLTHROUGH_INTENDED
-#define FALLTHROUGH_INTENDED \
-  do {                       \
-  } while (0)
+#define FALLTHROUGH_INTENDED
```
kqyang, line 49: What is the reason of removing the original definition?

packager/media/base/playready_pssh_generator.cc
```diff
@@ -59,6 +57,45 @@ std::vector<uint8_t> ConvertGuidEndianness(const std::vector<uint8_t>& input) {
...
+  if (location != std::string::npos) {
+    str->replace(location, from.size(), to);
+  }
+}
+
+void AesEcbEncrypt(const std::vector<uint8_t>& key,
+                   const std::vector<uint8_t>& plaintext,
+                   std::vector<uint8_t>* ciphertext) {
+  CHECK_EQ(plaintext.size() % AES_BLOCK_SIZE, 0u);
+  ciphertext->resize(plaintext.size());
```
kqyang, line 73: So for ECB, the output size does not need to be one block more than the input size?

kqyang (approved):
lgtm

packager/media/base/macros.h
joeyparrish: In an earlier PR, you advised me to remove the do/while(0) around some other macro definition.  So I did the same here.  I don't recall it causing any issues.

packager/media/base/playready_pssh_generator.cc
joeyparrish: I double-checked the tests with valgrind, and got no errors.  Then I noticed that this code is not actually covered by tests...

I'll add an extra block.  :-(

I think we will need to do some code coverage analysis at some point.  I'll file an issue for that, too.

PR #1129, opened 2022-11-03
feat: Replace arraysize with std::size

joeyparrish:
This does not depend on absl, which frees macros.h from any library deps that would make the CMake dependency tree more complicated.  This also fixes build errors in some environments.


joeyparrish:
This seems to explain some portion of the build errors in the first version of PR #1127.

theodab (approved):

PR #1159, opened 2022-12-15
ci: Parallelize and cache builds

joeyparrish:
Previously, we disabled parallel builds in CI because we were running out of memory.  This re-enables parallel builds with limited parallelism in linking (since linking debug builds was the major cause of OOM conditions).

Limiting parallelism for linking only requires the use of the Ninja generator, which requires some additional setup in the CI environment.

This also enables caching with ccache to speed up the build steps. This should make CI feedback much faster when small changes are made to a PR.


joeyparrish:
Hrm... Not caching on Windows yet.  Needs more work.

PR #1162, opened 2022-12-23
feat: Allow language code "und"

### LINKED ISSUE #1161: Consider supporting language tag "und"
Currently (on v2.6.1) Shaka Packager will error out when the language tag is specified to be "und":
```
Failed to initialize packager: 3 (INVALID_ARGUMENT): Unknown/invalid language specified: und
```

However, since both HLS and DASH allow this value, it makes sense to allow it also in Shaka, i.e. pass the value to the according tags in m3u8 and mpd.

Not a huge feature I guess, but it will make Shaka a little bit easier to use.

unext-wendong:
This allows the stream descriptor "language" to be specified as "und".

In this change, the convention of using return value "und" to indicate invalid language code in function LanguageToISO_639_2 has been changed to using "qaa" which is a locally scoped code per ISO-639 for the same purpose.

Closes #1161

### DIFF
--- a/docs/source/options/stream_descriptors.rst
+++ b/docs/source/options/stream_descriptors.rst
@@ -49,7 +49,9 @@ These are the available fields:
 :language (lang):
 
     Optional value which contains a user-specified language tag. If specified,
-    this value overrides any language metadata in the input stream.
+    this value overrides any language metadata in the input stream. Invalid
+    language tags will be treated as 'und' which indicates the language is
+    undetermined.
 
 :output_format (format):
 
--- a/packager/media/base/language_utils.cc
+++ b/packager/media/base/language_utils.cc
@@ -133,7 +133,7 @@ std::string LanguageToISO_639_2(const std::string& language) {
   LOG(WARNING) << "No equivalent 3-letter language code for " << main_language;
   // This is probably a mistake on the part of the user and should be treated
   // as invalid input.
-  return "und";
+  return "qaa"; // Use a locally scoped ISO-639-2 code to denote invalid input.
 }
 
 }  // namespace shaka
--- a/packager/media/base/language_utils.cc
+++ b/packager/media/base/language_utils.cc
@@ -133,7 +133,7 @@ std::string LanguageToISO_639_2(const std::string& language) {
   LOG(WARNING) << "No equivalent 3-letter language code for " << main_language;
   // This is probably a mistake on the part of the user and should be treated
   // as invalid input.
-  return "qaa"; // Use a locally scoped ISO-639-2 code to denote invalid input.
+  return "und";
 }
 
 }  // namespace shaka
--- a/packager/packager.cc
+++ b/packager/packager.cc
@@ -943,10 +943,10 @@ Status Packager::Initialize(
     // Update language to ISO_639_2 code if set.
     if (!copy.language.empty()) {
       copy.language = LanguageToISO_639_2(descriptor.language);
-      if (copy.language == "qaa") {
-        return Status(
-            error::INVALID_ARGUMENT,
-            "Unknown/invalid language specified: " + descriptor.language);
+      if (copy.language == "und" && descriptor.language != "und") {
+        LOG(WARNING)
+            << "Unknown/invalid language specified: " << descriptor.language
+            << ". Treat the language as undetermined.";
       }
     }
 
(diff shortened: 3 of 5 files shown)


joeyparrish:
> In this change, the convention of using return value "und" to indicate invalid language code in function LanguageToISO_639_2 has been changed to using "qaa" which is a locally scoped code per ISO-639 for the same purpose.

I disagree with your interpretation of "qaa".  It appears to be reserved for local use, but without any specification about its meaning.  Therefore I don't think it's right to use it in place of "und", which has a meaning and was being used for its intended purpose.

I think the change from "und" to "qaa" is also unrelated to your goal: making "und" valid input.  For that, you could just remove the check that rejects "und" as invalid input.  (In packager.cc)

unext-wendong:
Thanks for your comment.

To me, there are two different cases. 1) the language is undetermined and the user specified "und" explicitly, and 2) the user specified an invalid language code.

The current LanguageToISO_639_2 implementation returns "und" for both cases, and this makes it difficult for the caller to differentiate them. Therefore they end up with the same result, e.g. erroring out currently.

Case 1) seems a valid use case to me, that's why in #1161 I made the suggestion. For case 2) however, I feel the current behavior is appropriate.

```
   // This is probably a mistake on the part of the user and should be treated
   // as invalid input.
```
From this comment in the code (originally from commit https://github.com/shaka-project/shaka-packager/commit/77ec23afe4103d0d7f1029a3682033e9115ea741), I can tell that it was indeed intentional to fail case 2) at least.

Simply removing the check would be much simpler but it also means case 2) will be treated as a valid use case too.

I can make that change if that's what you prefer. For the record, I don't think that's better, but you are the repo owner, and it's your call.


> has been changed to using "qaa" which is a locally scoped code per ISO-639 for the same purpose.

Btw, sorry for not being clear in my change description, what I wanted to say is: it has been changed to use "qaa" as a return value to indicate invalid language code, i.e. same purpose as currently by using "und". I did not mean to say "qaa" per spec has such meaning. It would be much clearer if I wrote it like this:
```
has been changed to using "qaa" (which is a locally scoped code per ISO-639) for the same purpose.
```

joeyparrish:
I still think the approach here is wrong.  We should not replace all instances of "und" with "qaa".  Instead, the check to disallow "und" as input should simply be removed.

Since "qaa" is reserved for local use, it should be accepted as input but never fabricated by Packager itself.  (We are not the authority on anything "local" to your application.)

We can and should still generate the three-letter code "und" as output for unrecognized two-letter language inputs.  [ISO 639-2](http://www.loc.gov/standards/iso639-2/faq.html#25) says: "If there is language content, but the specific language cannot be determined, a special identifier is provided by ISO 639-2: und".

@unext-wendong, can you make these changes to your PR?

unext-wendong:
Will make the change as you suggested.

Just for logging purpose:

> and 2) the user specified an invalid language code.

I interpret that your intention is to treat the language as undetermined in this case.

> "If there is language content, but the specific language cannot be determined, a special identifier is provided by ISO 639-2: und"

I have a different understanding here. Since the app chose to specify the tag explicitly, the app is telling us they know what language the content contains. It could be "und", in that case we accept it trusting the app. However, if it's something we don't recognize, there is a chance that the app has specified an invalid input, and this to me is not same as we got a content with an undetermined language. I still think it's better to keep the current behavior in this case, where we error out to let the app be aware of it.

About the "qaa", I was seeing it differently.
From the app's point of view, shaka-packager as a 3rd-party library shouldn't be seen as "local" to the app.
But reading the spec again now, I'm not really sure if that's the indented meaning for "local".

unext-wendong:
If to go with the original approach, I guess returning an empty string instead of "qaa" from `LanguageToISO_639_2` should clear your concern about the "qaa" usage.

unext-wendong:
@joeyparrish Change is done. Could you copy it over and merge it in? As I don't have CLA signed.

joeyparrish:
If you haven't signed the CLA, I can't copy your code and merge it.  The CLA is a copyright assignment to Google.  Without that, I can't just copy your code.

joeyparrish:
Let me know if there's anything I can do to help you with the CLA.

unext-wendong:
@joeyparrish Thanks. Will check with my employer.

joeyparrish:
@unext-wendong, are you able to sign the CLA?

unext-wendong:
@joeyparrish I'm still waiting for the response from my employer. We've busy preparing a release of a new feature recently. But hopefully I can get an update next week.

cosmin:
Closing due to no CLA for over 6 months.

PR #1187, opened 2023-03-15
fix: Fix type error in pssh-box.py with Python 3.10+

Brainiarc7:
A single-line change on #L170 to `wv.protection_scheme = struct.unpack('>L', bytes(protection_scheme, encoding='utf-8'))[0]`, needed to work around this issue on Ubuntu 22.04LTS+ running Python 3.10+:

```sh
TypeError: a bytes-like object is required, not 'str'
```
On line 170.


joeyparrish:
Looks good.  Thanks for contributing!

joeyparrish (approved):

PR #1188, opened 2023-03-15
fix: Fix issues with `collections.abc` in Python 3.10+

### LINKED ISSUE #1192: Run-time issues on Ubuntu 22.04LTS due to python changes related to `abc.collections`, patch included.
## System info
**Operating System**: Ubuntu 22.04LTS
**Shaka Packager Version**: git tip master

On Ubuntu 22.04LTS, running an invocation of `pssh-box.py` tool would result in an error similar to:
```sh
AttributeError: module 'collections' has no attribute 'abc'
```
I've included a patch #1188 that resolves this.
Please review.

Brainiarc7:
This issue is observed on Python 3.10+ and above.
This workaround addresses a major backwards compatibility break with a major Python release version where collections.abc isn't available.

Fixes #1192

### DIFF
--- a/packager/third_party/protobuf/python/google/protobuf/internal/containers.py
+++ b/packager/third_party/protobuf/python/google/protobuf/internal/containers.py
@@ -42,7 +42,14 @@ are:
 __author__ = 'petar@google.com (Petar Petrov)'
 
 import collections
+
+try:
+    packagerCor = collections.abc
+except AttributeError:
+    packagerCor = collections
+
 import sys
+
 
 if sys.version_info[0] < 3:
   # We would use collections.MutableMapping all the time, but in Python 2 it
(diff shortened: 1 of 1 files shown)


Brainiarc7:
A cleaner solution to this would be updating the third-party module `protobuf` to v.4.0.x.
That release already incorporates a cleaner fix for this.
The release used in `shaka-packager` (at present) is `3.3.0` from 5 years ago.

joeyparrish:
The `cmake` branch, where we are overhauling the dependencies and the build system, uses source from around Protobuf v3.21.9 via git submodules.  Does that version have the correct fixes, or should we bump to v4 in that branch?

I'll go ahead and merge this into `main`, but since it touches third-party code, and the `cmake` branch uses submodules, this can't be merged over to the `cmake` branch.

PR #1201, opened 2023-04-24
fix: Don't close upstream on HttpFile::Flush

### LINKED ISSUE #1196: WEBM: Init segment with HTTP PUT destination hangs
## System info
**Operating System**: Ubuntu Focal
**Shaka Packager Version**: v2.6.1

**Packager Command**:

```bash
packager 'input=opus_96.opus.webm,stream=0,init_segment=http://localhost:35559/http/init.webm,segment_template=http://localhost:35559/http/$Number%05d$.webm'
``` 

(1) Spin up a local HTTP server supporting chunk transfer encoding
(2) Any supported input file will probably do, but I have only tested with opus in webm so far.

What is the expected result?

Expected shaka to produce and upload the init segment and the segments to the local http server.

What happens instead?

The local http server does get a couple of PUT requests for the init segment path, but they contain 0 bytes of data.
Shaka packager then appears to hang, but if one enables more verbose logging it can be observed that there appears to be an infinite loop of write attempts to the init segment internally. See atached log.

What happens is that when the `HttpFile::Write` call is made the `upload_cache_` is already close, so it returns 0 bytes written and this will be retried over and over again but will never succeed. I think the reason that the `upload_cache_` is closed is due to the `Flush` happening in [ThreadedIoFile::Seek](https://github.com/shaka-project/shaka-packager/blob/main/packager/file/threaded_io_file.cc#L123) causing the `HttpFile::Flush` method to [close the upload cache](https://github.com/shaka-project/shaka-packager/blob/main/packager/file/http_file.cc#L255). Somewhere, someone needs to Reopen the upload cache for this to break out of the infinite retry loop.

petzeb:
Closing the upstream on flush will effectively terminate the ongoing curl connection. This means that we would need re-establish the connection in order to resume writing, this is not what we want. In the spirit of the documentation of File::Flush

```c++
/// Flush the file so that recently written data will survive an 
/// application crash (but not necessarily an OS crash). For 
/// instance, in LocalFile the data is flushed into the OS but not 
/// necessarily to disk.
```

We will instead wait for the curl thread to finish consuming what ever might be in the upload cache, but leave the connection open for subsequent writes.

Fixes #1196

### DIFF
--- a/packager/file/http_file.cc
+++ b/packager/file/http_file.cc
@@ -243,6 +243,7 @@ int64_t HttpFile::Read(void* buffer, uint64_t length) {
 }
 
 int64_t HttpFile::Write(const void* buffer, uint64_t length) {
+  DCHECK(!upload_cache_.closed());
   VLOG(2) << "Writing to " << url_ << ", length=" << length;
   return upload_cache_.Write(buffer, length);
 }
@@ -253,7 +254,8 @@ int64_t HttpFile::Size() {
 }
 
 bool HttpFile::Flush() {
-  upload_cache_.Close();
+  // Wait for curl to read any data we may have buffered.
+  upload_cache_.WaitUntilEmptyOrClosed();
   return true;
 }
 


petzeb:
This is an attempt to fix https://github.com/shaka-project/shaka-packager/issues/1196
Open for other suggestions.

petzeb:
> Should I tag this PR as closing issue #1196?

Yes, if this approach seems reasonable. I am quite new to the shaka-packager code base.

> Could you add a unit test for this?

I'll see what I can do 👍

petzeb:
@joeyparrish Sorry for the delay, I was just about to start looking at adding a unit test but I noticed now that [all the unit tests for http_file](https://github.com/shaka-project/shaka-packager/blob/main/packager/file/http_file_unittest.cc) seem to be disabled. Any idea why this is?

joeyparrish:
I don't know the history for those tests being disabled, but in the `cmake` branch, they have been re-enabled.  When the `cmake` porting work is complete and merges back into `main`, everything should be enabled and working.

I will merge this as is, but follow up with a new test in the `cmake` branch.

joeyparrish:
Assigned myself https://github.com/shaka-project/shaka-packager/issues/1224 to follow up with a new test in the `cmake` branch, but if you want to do it yourself, please feel free to comment on #1224 so I can assign it to you.

Thanks!

joeyparrish (approved):

joeyparrish:
This is causing issues with upload tests.  Essentially, they hang forever because curl thinks there will be more chunks to upload.

I see how your change is more in line with the original intent of Flush(), and I have yet to find evidence of Flush() being used in the normal pipeline to write data to the server during an upload.  It should perhaps be Close() that signals the last chunk has been sent, but there will need to be more changes to make that work, since right now Close() terminates the HTTP connection, possibly without reading the response.

joeyparrish:
I believe I have a solution now.  Please see #1230, #1231, and #1232.

PR #1203, opened 2023-05-01
test: Retry HTTP file tests on temporary httpbin failure

joeyparrish:


packager/file/http_file_unittest.cc
```diff
@@ -64,176 +66,230 @@ nlohmann::json HandleResponse(const FilePtr& file) {
...
-  ASSERT_JSON_STRING(json, "method", "GET");
+  RetryTest(
+      []() -> HttpFile* {
+        return new HttpFile(HttpMethod::kGet, "https://httpbin.org/anything");
+      },
+      [](FilePtr&) -> void {},
+      [](FilePtr& file, nlohmann::json json) -> void {
+        ASSERT_TRUE(json.is_object());
+        ASSERT_TRUE(file.release()->Close());
+        ASSERT_JSON_STRING(json, "method", "GET");
```
kqyang, line 114: nit: a slightly cleaner approach here and below: line 112 and 113 can be moved into RetryTest implementation, and line 114 can be checked after RetryTest. This eliminates the post_read() function callback.

kqyang (approved):
lgtm with nits.

packager/file/http_file_unittest.cc
joeyparrish: We can't move 112 into RetryTest, because the error tests will produce is_object() == false.

If we're making changes to the test structure, though, I would go farther with 113 and eliminate it.  If we don't care about testing Close() return value, it's not providing value.  FilePtr being a unique_ptr with a closer will take care of closing the HttpFile object no matter what.  I'll drop it completely.  If you feel we need to test Close() returning true, I'll add an explicit test case for that.  But it doesn't seem necessary to me at a glance.

Checking 114 after RetryTest would require us to return the json object and check results after the retry callback.  That only works if we don't need to do ASSERT_* in RetryTest, though.  ASSERT_* macros can only be used in void functions, since they return early on failure.  I currently have ASSERT_TRUE(file->Open()) in RetryTest, so I can't have json be a return value.

Also, some test cases want to verify things on file (error cases), while some want to verify things about json (success cases).  So I prefer the callback that provides both, while letting the RetryTest function manage the lifetime of the HttpFile object.

packager/file/http_file.cc
```diff
@@ -231,6 +232,10 @@ bool HttpFile::Open() {
   return true;
 }
 
+int HttpFile::HttpStatusCode() {
```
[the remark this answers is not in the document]
joeyparrish: Will do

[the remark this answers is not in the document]
joeyparrish: Done

packager/file/http_file_unittest.cc
joeyparrish: I dropped the Close() check, but the other changes you suggested didn't work

kqyang: sg.

kqyang (approved):

PR #1210, opened 2023-05-27
feat(DASH): Add video transfer characteristics.

sr1990:
This PR is related to https://github.com/shaka-project/shaka-packager/issues/1035


packager/mpd/base/mpd_utils.cc
```diff
@@ -162,6 +162,15 @@ std::string GetAdaptationSetKey(const MediaInfo& media_info,
   if (!ignore_codec) {
     key.append(":");
     key.append(GetBaseCodec(media_info));
+
+    if (GetBaseCodec(media_info).find("dvh") == 0) {
```
vish91, line 166: @sr1990  could you explain the logic here a bit ? 
what's this conditional logic for dvh for ?

sr1990: @vish91 This is similar to https://github.com/shaka-project/shaka-packager/blob/80e024013df87a4bfeb265c8ea83cfa2a0c5db0f/packager/hls/base/media_playlist.cc#L554
which is used to get HLS VIDEO-RANGE.
VUI which contains the transfer characteritics is an optional section of syntax within the SPS.
As per spec at https://www.itu.int/rec/T-REC-H.265-202108-I/en
```
vui_parameters_present_flag equal to 1 specifies that the vui_parameters( ) syntax structure as specified in Annex E is
present. vui_parameters_present_flag equal to 0 specifies that the vui_parameters( ) syntax structure as specified in
Annex E is not present.
```

This check is done while parsing SPS
H265Parser::ParseSps https://github.com/shaka-project/shaka-packager/blob/80e024013df87a4bfeb265c8ea83cfa2a0c5db0f/packager/media/codecs/h265_parser.cc#L628
and H264Parser::ParseSps
https://github.com/shaka-project/shaka-packager/blob/80e024013df87a4bfeb265c8ea83cfa2a0c5db0f/packager/media/codecs/h264_parser.cc#L695

In case of dolby vision content, even if the VUI is not present, the transfer characteristics are set to PQ.
I will go ahead and add
`  // Dolby Vision (dvh1 or dvhe) is always HDR.`
comment which is also present in MediaPlaylist::GetVideoRange().

packager/mpd/base/period.cc
```diff
@@ -269,6 +269,20 @@ bool Period::SetNewAdaptationSetAttributes(
...
 
+    // Set transfer characteristics.
+    // https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf - 4.2.5.1
+    // ISO/IEC 23001-8 MPEG systems technologies — Part 8: Coding-independent
+    // code points. https://en.wikipedia.org/wiki/Coding-independent_code_points
+    // - Common CCIP values.
+    // Dolby vision:
+    // https://professionalsupport.dolby.com/s/article/How-to-signal-Dolby-Vision-in-MPEG-DASH
+    if (new_adaptation_set->codec().find("dvh") == 0) {
+      new_adaptation_set->set_transfer_characteristics(16);
```
joeyparrish, line 280: Please define a reasonably-named constant for this instead of hard-coding 16.  It will be more readable with a name.  What does 16 represent?  The constant's name should answer that question.

packager/mpd/base/mpd_utils.cc
joeyparrish: @vish91, does that satisfy your concern?

packager/mpd/base/mpd_utils.cc
```diff
@@ -162,6 +162,15 @@ std::string GetAdaptationSetKey(const MediaInfo& media_info,
...
+      key.append(":");
+      key.append(std::to_string(16));
```
joeyparrish, line 168: Please use a named constant here for readability.

joeyparrish (changes requested):

packager/mpd/base/mpd_utils.cc
vish91: yeah this makes sense.  :shipit:

packager/mpd/base/period.cc
sr1990: Updated.

joeyparrish:
The robot is not smart enough to recognize new commits as activity.  I'll update it soon.  Reopening the PR.

packager/mpd/base/mpd_utils.cc
```diff
@@ -162,6 +162,16 @@ std::string GetAdaptationSetKey(const MediaInfo& media_info,
   if (!ignore_codec) {
     key.append(":");
     key.append(GetBaseCodec(media_info));
+
+    if (media_info.video_info().has_transfer_characteristics()) {
```
joeyparrish, line 166: @sr1990, I added a comment to the Dolby section of this, and made it so that explicit transfer characteristics take priority over the assumption of PQ for Dolby.  Please take a look, and let me know if you agree.

packager/mpd/base/period.cc
```diff
@@ -269,6 +269,21 @@ bool Period::SetNewAdaptationSetAttributes(
...
     }
 
+    // Set transfer characteristics.
+    // https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf - 4.2.5.1
+    // ISO/IEC 23001-8 MPEG systems technologies — Part 8: Coding-independent
+    // code points. https://en.wikipedia.org/wiki/Coding-independent_code_points
+    // - Common CCIP values.
+    // Dolby vision:
+    // https://professionalsupport.dolby.com/s/article/How-to-signal-Dolby-Vision-in-MPEG-DASH
+    if (media_info.video_info().has_transfer_characteristics()) {
```
joeyparrish, line 279: @sr1990, I made the same change here.  Please take a look.

joeyparrish (approved):

joeyparrish:
@sr1990, I will merge this if you approve of my edits.  If not, we can discuss it further.  Thanks!

joeyparrish:
My change caused a test failure where the golden output changed from transfer characteristics of 16 to 2.  Why is that? 
 Should a Dolby codec string really override an explicit setting where `media_info.video_info().has_transfer_characteristics()`?  Is the test input mp4 wrong for this scenario?

joeyparrish:
> I don't think the input mp4 is wrong as I had checked Dolby Profile 5 content at https://developer.dolby.com/tools-media/sample-media/video-streams/dolby-vision-streams/ which also has the transfer function set to 2 (unspecified).
> 
> Found https://developer.apple.com/av-foundation/High-Dynamic-Range-Metadata-for-Apple-Devices.pdf which mentions that for DV Profile 5: • Color Transfer Function Index shall be set to 2 (Unspecified) in the colr atom. Not sure why it is set to 2.
>
> I think we should give the codec string higher priority over transfer characteristics in case of Dolby.
> What do you think?

Okay, then this is a peculiarity of Dolby, and Dolby codec strings really should have higher priority.  But we should add a comment to the code in those places to explain it, because it's not at all obvious.  I suggest something like:

```c++
  // Apple's docs for DV profile 5 state: "Color Transfer Function Index shall
  // be set to 2 (Unspecified) in the colr atom."  So we can't trust the colr
  // atom for Dolby inputs, and must set the transfer function based on the
  // codec string alone.
```

sr1990:
@joeyparrish Sounds good. Right now we are getting the transfer characteristics from SPS VUI and not colr atom. Looks like https://github.com/shaka-project/shaka-packager/pull/1205 is working on reading and and mentions writing the colr atom.
I think we can complete this PR right now without the comment and once PRs for colr reading and writing are done, add the colr box for dolby vision as per the spec (even if the input does not contain the colr box) and with the comment. 
OR if you prefer, we can complete this PR after https://github.com/shaka-project/shaka-packager/pull/1205 and colr atom writing is done.

sr1990:
@joeyparrish, based on conversation at https://github.com/shaka-project/shaka-packager/pull/1255
```
Dolby Vision profile 5 must be PQ. The VUI information is optional, 
usually we don't care the value of transfer_characteristics, 
both 2(UNSPECIFIED) and 16 (PQ) are valid.
```
I think we should keep the original code and add the above comment w/o writing the colr atom. 

joeyparrish:
Sounds good.

sr1990:
> Sounds good.

Updated.

joeyparrish (approved):

joeyparrish:
@sr1990, thank you for your patience and persistence!

PR #1211, opened 2023-05-30
fix(MP4): Add compatible brand dby1 for Dolby content.

sr1990:
This PR adds dby1 compatible brand to dolby content as per https://professional.dolby.com/siteassets/content-creation/dolby-vision-for-content-creators/dolby_vision_bitstreams_within_the_iso_base_media_file_format_dec2017.pdf


packager/media/formats/mp4/mp4_muxer.cc
```diff
@@ -227,6 +227,12 @@ Status MP4Muxer::DelayInitializeMuxer() {
                             ->h26x_stream_format());
       if (codec_fourcc != FOURCC_NULL)
         ftyp->compatible_brands.push_back(codec_fourcc);
+
+      // https://professional.dolby.com/siteassets/content-creation/
+      // dolby-vision-for-content-creators/
+      // dolby_vision_bitstreams_within_the_iso_base_media_file_format_dec2017.pdf
```
joeyparrish, line 233: I believe URLs should be exempt from column limits.  (Tell me if I'm wrong.)  Please put this back together where someone could click it or copy it more easily.

joeyparrish (changes requested):

packager/media/formats/mp4/mp4_muxer.cc
sr1990: Will update. https://google.github.io/styleguide/cppguide.html#Line_Length mentions when line may exceed 80 characters.

sr1990: Updated.

joeyparrish (approved):

PR #1148, opened 2022-11-27
feat: Cmake port of Media crypto

zdanek:
One of commits is cherry picked Codecs port as Media Crypto depends on this. Once Codecs will be merged into cmake branch I will rebase and get rid of this technical cherry pick.

PR #1221, opened 2023-06-28
feat: Cmake port of Media crypto

cosmin:
Rebasing #1148 on top of current cmake branch

Related to #1047


joeyparrish:
@cosmin, could you fix the linter errors?

To run the linter against the changes since your local `cmake` branch, run:

```sh
python3 packager/tools/git/check_formatting.py cmake
```

To fix the same errors, run:

```sh
git clang-format --style Chromium cmake
```

For both of those, make sure your local `cmake` branch is up-to-date with the upstream `cmake` branch.

Thanks!

packager/media/base/aes_encryptor.cc
```diff
@@ -129,17 +129,40 @@ bool AesCbcEncryptor::CryptInternal(const uint8_t* plaintext,
...
-  const size_t residual_block_size = plaintext_size % AES_BLOCK_SIZE;
+
   const size_t num_padding_bytes = NumPaddingBytes(plaintext_size);
-  // mbedtls requires a buffer large enough for one extra block.
   const size_t required_ciphertext_size =
       plaintext_size + num_padding_bytes + AES_BLOCK_SIZE;
+
   if (*ciphertext_size < required_ciphertext_size) {
+    VLOG(1) << "Expected output size of at least "
+               << required_ciphertext_size << " bytes. Using teporary buffer.";
```
joeyparrish, line 139: typo: "temporary"

packager/media/base/aes_encryptor.cc
```diff
@@ -129,17 +129,40 @@ bool AesCbcEncryptor::CryptInternal(const uint8_t* plaintext,
                                     size_t plaintext_size,
                                     uint8_t* ciphertext,
                                     size_t* ciphertext_size) {
-  const size_t residual_block_size = plaintext_size % AES_BLOCK_SIZE;
+
   const size_t num_padding_bytes = NumPaddingBytes(plaintext_size);
-  // mbedtls requires a buffer large enough for one extra block.
```
joeyparrish, line 134: I think this was still useful context on the calculation of required_ciphertext_size.  Please restore this!

packager/media/base/aes_encryptor.cc
```diff
@@ -129,17 +129,40 @@ bool AesCbcEncryptor::CryptInternal(const uint8_t* plaintext,
...
   const size_t required_ciphertext_size =
       plaintext_size + num_padding_bytes + AES_BLOCK_SIZE;
+
   if (*ciphertext_size < required_ciphertext_size) {
+    VLOG(1) << "Expected output size of at least "
+               << required_ciphertext_size << " bytes. Using teporary buffer.";
+
+    std::unique_ptr<uint8_t[]> ciphertext_buffer(new uint8_t[required_ciphertext_size]);
```
joeyparrish, line 141: I don't recall having to do this anywhere else.  I think it is expected that we will allocate the right size buffer before calling this method, so that we aren't churning the allocator unnecessarily.  We should control all call sites, and any encryption failure should be fixed at the caller.

At least, this is how we've done it so far.  Please make your case if you disagree.

packager/media/base/aes_encryptor.cc
```diff
@@ -129,17 +129,40 @@ bool AesCbcEncryptor::CryptInternal(const uint8_t* plaintext,
...
+}
+
+bool AesCbcEncryptor::CryptInternal0(const uint8_t* plaintext,
+                                    size_t plaintext_size,
+                                    uint8_t* ciphertext,
+                                    size_t* ciphertext_size) {
+  const size_t residual_block_size = plaintext_size % AES_BLOCK_SIZE;
+  const size_t num_padding_bytes = NumPaddingBytes(plaintext_size);
+  // mbedtls requires a buffer large enough for one extra block.
+  if (*ciphertext_size < plaintext_size + num_padding_bytes) {
```
joeyparrish, line 159: I think it will be more readable to put this back into a named variable like required_ciphertext_size: plaintext_size + num_padding_bytes.  You have to repeat it immediately below in the error log anyway.

packager/media/crypto/encryption_handler.cc
```diff
@@ -390,7 +390,8 @@ void EncryptionHandler::EncryptBytes(const uint8_t* source,
   DCHECK(source);
   DCHECK(dest);
   DCHECK(encryptor_);
-  CHECK(encryptor_->Crypt(source, source_size, dest));
+  CHECK(encryptor_->Crypt(source, source_size, dest, &source_size));
+//bzd  CHECK(encryptor_->Crypt(source, dest));
```
joeyparrish, line 394: Please remove

packager/media/crypto/encryption_handler.cc
```diff
@@ -390,7 +390,8 @@ void EncryptionHandler::EncryptBytes(const uint8_t* source,
   DCHECK(source);
   DCHECK(dest);
   DCHECK(encryptor_);
-  CHECK(encryptor_->Crypt(source, source_size, dest));
+  CHECK(encryptor_->Crypt(source, source_size, dest, &source_size));
```
joeyparrish, line 393: Why the extra parameter?

packager/media/base/aes_encryptor.cc
cosmin: @zdanek thoughts?

packager/media/crypto/encryption_handler.cc
cosmin: I was curious about this as well, but I figured I'd start with @zdanek's changes and we can iterate from there.

cosmin: Reverted the changes, as it seemed this wasn't needed, since the 3 param version calls the 4 param version internally.

packager/media/base/aes_encryptor.cc
cosmin: I reverted this to the previous behavior, to the extent we need to deal with callers passing in the wrong size buffer we can do that in a separate pull request, keeping this one focused on just changes needed to move to CMake.

packager/media/base/aes_encryptor.h
```diff
@@ -89,6 +89,10 @@ class AesCbcEncryptor : public AesCryptor {
                      size_t plaintext_size,
                      uint8_t* ciphertext,
                      size_t* ciphertext_size) override;
+  bool CryptInternal0(const uint8_t* plaintext,
```
joeyparrish, line 92: I think this needs to be dropped now, since you reverted the implementation.

packager/third_party/mbedtls/CMakeLists.txt
```diff
@@ -14,4 +14,5 @@ set(ENABLE_TESTING OFF)
 set(GEN_FILES OFF)
 
 # With these set in scope of this folder, load the library's own CMakeLists.txt.
+add_compile_options(-Wno-unused-but-set-variable)
```
joeyparrish, line 17: This is causing a build error on Windows:

`cl : command line  error D8021: invalid numeric argument '/Wno-unused-but-set-variable' [D:\a\shaka-packager\shaka-packager\build\packager\third_party\mbedtls\source\library\mbedcrypto.vcxproj]`

For similar situations in the past, we have had to use `if(NOT MSVC)` ... `endif()` to make the flag conditional on the compiler.

joeyparrish (changes requested):

joeyparrish (approved):

cosmin:
Now that I reverted the change with dealt with calling `CryptInternal` with an insufficiently large ciphertext size I'm seeing a test failure related to this. Taking a look at that.

cosmin:
https://github.com/cosmin/shaka-packager/blob/75987351976cd4a84b24cf1590c77482aab9250a/packager/media/crypto/encryption_handler.cc#L302

ok, I think this is an example of a callsite that sets up the cipher text buffer to be the same size as the plaintext, but when using `AesCbcEncryptor` the required ciphertext size is larger due to padding. This then fails the check, which the earlier commit worked around by re-allocating a larger buffer.

cosmin:
@joeyparrish take a look at the most recent commit, the only way I could think to workaround this issue is to introduce a function to return the expected ciphertext size so the caller can allocate a properly sized buffer for it. Open to suggestions though if there's a better way to solve this.

packager/media/base/aes_encryptor.cc
```diff
@@ -125,15 +125,20 @@ bool AesCbcEncryptor::InitializeWithIv(const std::vector<uint8_t>& key,
   return SetIv(iv);
 }
 
+size_t AesCbcEncryptor::RequiredCiphertextSize(size_t plaintext_size) {
```
joeyparrish, line 128: I'm embarrassed for not thinking of this myself.  This is so much better!  We shouldn't have to have this knowledge distributed to all the call sites.  Thank you!

packager/media/crypto/encryption_handler.cc
```diff
@@ -314,15 +317,15 @@ Status EncryptionHandler::ProcessMediaSample(
         total_size += subsample.clear_bytes;
       }
       if (subsample.cipher_bytes > 0) {
-        EncryptBytes(source, subsample.cipher_bytes, dest);
+        EncryptBytes(source, subsample.clear_bytes, dest, ciphertext_size);
```
joeyparrish, line 320: I think this should be cipher_bytes, not clear_bytes.  IIUC, the parameter is the number of input bytes to encrypt, which is cipher_bytes (to be encrypted), not clear_bytes (left clear).

Since it seems easy to make a mistake with that here, perhaps you should both revert that parameter and add a comment explaining.

joeyparrish (changes requested):
Such an improvement, thanks!

packager/media/crypto/encryption_handler.cc
cosmin: ah, ok, that makes sense, reverted and added a comment as you suggested.

joeyparrish (approved):

joeyparrish:
I have now fixed the HTTP tests, so I'm going to merge your branch with `cmake` and update the PR again.  Then we should see a fairly straightforward test pass and know that the crypto tests enabled by this PR are passing.

joeyparrish:
Build failure on Windows:

```
packager\media\crypto\subsample_generator.cc(273,28): error C2220: the following warning is treated as an error
packager\media\crypto\subsample_generator.cc(273,28): warning C4457: declaration of 'frame' hides function parameter
packager\media\crypto\subsample_generator.cc(262,20): message : see declaration of 'frame'
```

That sounds like it could be a real bug.

cosmin:
As far as I can tell this code has not changed recently. I don't think it's a bug in this case because by the time

```
  for (const VPxFrameInfo& frame : vpx_frames) {
```

happens which shadows the frame function parameter it was already parsed into `vpx_frames` by

```
  if (!vpx_parser_->Parse(frame, frame_size, &vpx_frames))
```

however it's trivial to fix while I'm at it.

cosmin:
All tests seem to pass now except for Arch which should be fixed by #1233

PR #1231, opened 2023-07-13
test: Add Mongoose embedded HTTP server

joeyparrish:
This adds Mongoose as a third-party library, and builds on top of that an embedded HTTP server for our unit tests.

We are using a fork of Mongoose pending the merging of this PR: https://github.com/cesanta/mongoose/pull/2301

The embedded web server will make our HTTP-based tests independent of httpbin.org, which will make them quick and reliable.


packager/media/test/test_web_server.cc
```diff
@@ -0,0 +1,261 @@
...
+// embedded web server can:
+//
+// 1. Reflect the request method, body, and headers
+// 2. Return a requested status code
+// 3. Delay a response by a requested amount of time
+
+namespace {
+
+// Get a string_view on mongoose's mg_string, which may not be nul-terminated.
+std::string_view view_mg_str(const mg_str& mg_string) {
```
kqyang, line 27: nit: use mixed case https://google.github.io/styleguide/cppguide.html#Function_Names

The same below.

packager/media/test/test_web_server.cc
```diff
@@ -0,0 +1,261 @@
...
+    {
+      absl::MutexLock lock(&mutex_);
+      stopped = stopped_;
+    }
+  }
+
+  {
+    absl::MutexLock lock(&mutex_);
+    status_ = kStopped;
+  }
```
kqyang, line 142: nit: why not just do it after line 134,

```
if (stopped)
  status_ = kStopped;
```

Avoids one extra lock.

kqyang (approved):
LGTM with a few nits.

Good call to switch to use a local web server framework

packager/media/test/test_web_server.cc
joeyparrish: Thanks!  It's so hard to keep that rule straight in my head when all the libraries we use are in snake_case.

packager/media/test/test_web_server.cc
joeyparrish: Ah, good point.  Done.

PR #1232, opened 2023-07-13
fix: Fix flush/close semantics for HTTP files, improve testing

### LINKED ISSUE #1224: Add a test for HttpFile::Flush
PR #1201 did not add a unit test because the HttpFile tests are all disabled in the `main` branch.  They have been re-enabled in the `cmake` branch, so we should add a new test there.

joeyparrish:
All HTTP-based tests now use an embedded test server instead of
httpbin.org, which makes them much faster and more reliable.

These more reliable tests also exposed some issues that began recently
with PR #1201.  HttpFile's Flush() semantics were different than those
documented for files in general.  Flush() used to close the file for
uploading, so that no further writes were allowed, but the documentation
stated that it would only flush data to its destination.  PR #1201
brought HttpFile's Flush() in line with the docs, but gave us no way to
terminate a chunked upload.

This adds a new method to File called CloseForWriting(), which
terminates a chunked upload for HttpFile.  The only other implementation
that does anything is UdpFile, which uses the socket library function
shutdown() to terminate writes while allowing reads.

This also tweaks HttpFile::CloseWithStatus() so that it will not
generate an error if the file is closed before the HTTP response is
written to the download cache.

This modifies the test HttpFileTest.MultipleWrites so that the file is
Flushed after each chunk.  This adds test coverage for the changes
introduced in PR #1201.

Fixes #1224 (missing test coverage for HttpFile::Flush)

### DIFF
--- a/packager/file/http_file_unittest.cc
+++ b/packager/file/http_file_unittest.cc
@@ -137,9 +137,7 @@ TEST_F(HttpFileTest, BasicPost) {
 
   ASSERT_EQ(file->Write(data.data(), data.size()),
             static_cast<int64_t>(data.size()));
-  ASSERT_TRUE(file->Flush());
-
-  // Tells the server in a chunked upload that there will be no more chunks.
+  // Signal that there will be no more writes.
   // If we don't do this, the request can hang in libcurl.
   file->CloseForWriting();
 
@@ -173,9 +171,7 @@ TEST_F(HttpFileTest, BasicPut) {
 
   ASSERT_EQ(file->Write(data.data(), data.size()),
             static_cast<int64_t>(data.size()));
-  ASSERT_TRUE(file->Flush());
-
-  // Tells the server in a chunked upload that there will be no more chunks.
+  // Signal that there will be no more writes.
   // If we don't do this, the request can hang in libcurl.
   file->CloseForWriting();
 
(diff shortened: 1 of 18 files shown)


packager/file/http_file_unittest.cc
```diff
@@ -165,21 +214,35 @@ TEST(HttpFileTest, MultipleWrites) {
 
   ASSERT_EQ(file->Write(data1.data(), data1.size()),
             static_cast<int64_t>(data1.size()));
+  ASSERT_TRUE(file->Flush());
```
kqyang: Why do we call Flush after every write in this test? Is it needed after every write?

kqyang (approved):
LGTM % one question

packager/file/http_file_unittest.cc
joeyparrish: It adds coverage for the change in Flush() semantics from #1201.  Now that Flush() only flushes, this shows the new behavior.  Before #1201, you could not Write() after Flush().

In the test server, you can see those pieces written as separate chunks now.  This isn't used yet, but it may be an important detail for chunked uploads for LL streaming in the future.

kqyang: I see. It is probably better to add another MultipleWritesWithFlush to test this case. Otherwise consecutive Writes without Flush is not tested.

joeyparrish: Good point.

joeyparrish: Done

PR #1233, opened 2023-07-14
fix: Fix compilation on Arch Linux

joeyparrish:
- Update protobuf to v23.4, to fix the cstdint error on Arch Linux, and make some related changes:
   - Silence additional compiler warnings for the new protobuf
 - Update absl to 20230125.3, to support the protobuf update, and make some related changes:
   - Silence additional compiler warnings for the new protobuf
   - Replace GOOGLE_CHECK_OK with ABSL_CHECK_OK
   - Replace GUARDED_BY with ABSL_GUARDED_BY
 - Update other instances of cstdint missing in our own code
 - Always pull the latest docker images, to avoid stale results when your workstation has pulled an older image at the same label

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -103,7 +103,7 @@ jobs:
         # NOTE: cmake is already installed in GitHub Actions VMs, but not
         # necessarily in a self-hosted runner.
         run: |
-          sudo apt install -y \
+          sudo apt update && sudo apt install -y \
               cmake \
               libc-ares-dev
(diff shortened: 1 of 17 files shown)


cosmin:
Windows tests are failing with another warning that needs to be silenced in protobuf

> protobuf\text_format.cc(905,11): warning C4805: '==': unsafe mix of type 'bool' and type 'uint64_t' in operation

kqyang (approved):

PR #1234, opened 2023-07-14
feat: convert mpd module to cmake

cosmin:
Issue #1047


packager/mpd/base/xml/xml_node.cc
```diff
@@ -540,16 +544,18 @@ bool RepresentationXmlNode::AddAudioChannelInfo(const AudioInfo& audio_info) {
       // Convert EC3 channel map into string of hexadecimal digits. Spec: DASH-IF
       // Interoperability Points v3.0 9.2.1.2.
       const uint16_t ec3_channel_map =
-        base::HostToNet16(codec_data.channel_mask());
-      audio_channel_config_value =
-        base::HexEncode(&ec3_channel_map, sizeof(ec3_channel_map));
+          absl::big_endian::FromHost16(codec_data.channel_mask());
+      uint8_t byte1 = ec3_channel_map & 0xFF;
+      uint8_t byte2 = ec3_channel_map >> 8;
+      audio_channel_config_value = absl::StrFormat("%02X%02X", byte1, byte2);
```
cosmin, line 550: couldn't find a better way to get the `base::HexEncode` here in big endian format, is this the best way to proceed?

packager/mpd/base/xml/xml_node.cc
```diff
@@ -583,9 +589,12 @@ bool RepresentationXmlNode::AddAudioChannelInfo(const AudioInfo& audio_info) {
...
       // audio G.3.1.
       const uint32_t ac4_channel_mask =
-        base::HostToNet32(codec_data.channel_mask() << 8);
+          absl::big_endian::FromHost32(codec_data.channel_mask() << 8);
+      absl::Span<const uint8_t> bytes(
+          reinterpret_cast<const uint8_t*>(&ac4_channel_mask),
+          sizeof(ac4_channel_mask));
       audio_channel_config_value =
-        base::HexEncode(&ac4_channel_mask, sizeof(ac4_channel_mask) - 1);
+          absl::StrFormat("%02X%02X%02X", bytes[0], bytes[1], bytes[2]);
```
cosmin, line 597: likewise here for the `uint32_t`, perhaps I should the same approach for both AC3 and AC4 though.

packager/mpd/test/mpd_builder_test_helper.cc
```diff
@@ -6,46 +6,42 @@
...
-      .Append(FILE_PATH_LITERAL("schema"))
-      .Append(FILE_PATH_LITERAL("DASH-MPD.xsd"));
-  return file_path;
+std::filesystem::path GetSchemaPath() {
+  std::filesystem::path header_path(__FILE__);
+  return header_path.parent_path() / "schema" / "DASH-MPD.xsd";
 }
 
-std::string GetPathContent(const base::FilePath& file_path) {
+std::string GetPathContent(std::filesystem::path& file_path) {
```
cosmin, line 31: should this go in file_test_util perhaps?

packager/mpd/test/xml_compare.cc
```diff
@@ -80,10 +78,17 @@ bool CompareContents(xmlNodePtr node1, xmlNodePtr node2) {
...
-  base::TrimString(node2_content, " ", &node2_content);
+
+  node1_content.erase(
+      std::remove(node1_content.begin(), node1_content.end(), '\n'),
+      node1_content.end());
+  node2_content.erase(
+      std::remove(node2_content.begin(), node2_content.end(), '\n'),
+      node2_content.end());
+
+  node1_content = absl::StripAsciiWhitespace(node1_content);
```
cosmin, line 89: I think `base::TrimString` and `absl::StripAsciiWhitespace` are equivalent but I'm not sure

packager/mpd/test/xml_compare.cc
```diff
@@ -80,10 +78,17 @@ bool CompareContents(xmlNodePtr node1, xmlNodePtr node2) {
       reinterpret_cast<const char*>(node1_content_ptr.get());
   std::string node2_content =
       reinterpret_cast<const char*>(node2_content_ptr.get());
-  base::ReplaceChars(node1_content, "\n", "", &node1_content);
-  base::TrimString(node1_content, " ", &node1_content);
-  base::ReplaceChars(node2_content, "\n", "", &node2_content);
-  base::TrimString(node2_content, " ", &node2_content);
+
+  node1_content.erase(
```
cosmin, line 82: this doesn't feel like the best way to do the equivalent of `base::ReplaceChars` but it works so I started with this

cosmin:
`static_cast<uint64_t>(ceil(total_size_in_bits_ / total_duration_))` in bandwidth_estimator.cc is throwing an error on windows due to narrowing conversion from double to uint64. Since this one seems quite intentional, what is the right way to do this without tripping on `C4244`?

packager/mpd/base/mpd_builder.h
```diff
@@ -32,6 +32,15 @@ class AdaptationSet;
 class MediaInfo;
 class Period;
 
+class Clock {
```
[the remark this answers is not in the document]
joeyparrish: I had to create something similar already in Muxer in b1095f6bc36f38892c848689135b3d1dab1e7e1b.  See packager/media/base/muxer.h.

We can always create a new folder to hold these things if we want to dedup them.  I agree that this doesn't really belong in mpd.

packager/mpd/base/mpd_builder_unittest.cc
```diff
@@ -117,17 +124,14 @@ class LiveMpdBuilderTest : public MpdBuilderTest<DashProfile::kLive> {
...
-                                       11,  // day_of_month
-                                       15,  // hour.
-                                       10,  // minute.
-                                       24,  // second.
-                                       0 };  // millisecond.
-    ASSERT_TRUE(test_time.HasValidValues());
-    mpd_.InjectClockForTesting(std::unique_ptr<base::Clock>(
-        new TestClock(base::Time::FromUTCExploded(test_time))));
+    std::tm test_time{};
+    test_time.tm_year = 2016 - 1900;  // Years since 1900
```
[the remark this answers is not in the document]
joeyparrish: I tend to find ISO 8601 time strings to be pretty readable.  If there's an easy way to parse those into the appropriate value type, I would use that.  Ex: "2016-01-11T15:10:24Z"

packager/mpd/base/mpd_builder_unittest.cc
```diff
@@ -20,14 +22,19 @@ namespace shaka {
...
+class TestClock : public Clock {
  public:
-  explicit TestClock(const base::Time& t) : time_(t) {}
-  ~TestClock() override {}
-  base::Time Now() override { return time_; }
+  explicit TestClock(std::tm& utc_time) {
+    std::tm local_time = utc_time;
+    std::time_t utc_time_t = std::mktime(&local_time);
+    std::time_t offset = utc_time_t - std::mktime(std::gmtime(&utc_time_t));
+    mock_time_ = std::chrono::system_clock::from_time_t(utc_time_t + offset);
```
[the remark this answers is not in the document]
joeyparrish: In muxer, I have the clock abstraction return uint64_t in seconds since 1970 UTC.  (Essentially the same as time_t.)  Depending on how it is consumed, you could do the same here, and avoid time_point and std::chrono.

packager/mpd/base/xml/xml_node.cc
joeyparrish: Maybe add another helper to packager/utils/hex_parser.  I'm not certain you need to decompose it into individual bytes if you've already done a swap with absl.  Just FromHost16 and StrFormat("%04X") should be fine, right?  Or FromHost16 and BytesToHexString?

packager/mpd/base/xml/xml_node.cc
joeyparrish: Same as above, I think.  Swap and print/format/hexify without needing Span or individual bytes.

Also, it took me a while to see the "- 1" on the original.  Please add a comment about us needing just 3 bytes.

packager/mpd/test/xml_compare.cc
joeyparrish: I expect so, yes

packager/mpd/test/xml_compare.cc
joeyparrish: LGTM.  It's a test helper, so it's fine.

joeyparrish:
> bandwidth_estimator.cc is throwing an error on windows due to narrowing conversion from double to uint64. Since this one seems quite intentional, what is the right way to do this without tripping on `C4244`?

Is this the `return 0.0;` line?  That's easy, just replace with `return 0`.  There's no reason that should be a floating-point literal in a function returning uint64_t.

packager/mpd/base/mpd_builder.cc
```diff
@@ -75,57 +73,60 @@ bool Positive(double d) {
...
-    relative_path = child_path;
-  return relative_path.NormalizePathSeparatorsTo('/').AsUTF8Unsafe();
+bool IsBasePathParent(const std::filesystem::path& path,
+                      const std::filesystem::path& base_path) {
+  std::string path_str = path.string();
+  std::string base_path_str = base_path.string();
+  return path_str.substr(0, base_path_str.length()) == base_path_str;
+}
+
+std::string MakePathRelative(const std::filesystem::path& media_path,
```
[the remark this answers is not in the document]
joeyparrish: Right.  That's what the snippet I gave should do.

```c++
  // media_path = "foo/bar/baz/manifest.mpd", parent_path = "foo/bar"
  auto relative_path = std::filesystem::relative(media_path, parent_path);
  // relative_path = "baz/manifest.mpd"

  if (relative_path.empty() || *relative_path.begin() == "..") {
    // Not related.
    return media_path;
  }

  // returns "baz/manifest.mpd" here
  return relative_path;
```

If it seems like this method should only ever get inputs that follow the happy path, we could treat the other possibilities as failures instead:

```c++
  auto relative_path = std::filesystem::relative(media_path, parent_path);
  CHECK(!relative_path.empty());
  CHECK(*relative_path.begin() != "..");
  return relative_path;
```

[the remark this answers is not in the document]
cosmin: Ah, yes, I should have read more closely. I did try that approach as well of using `std::filesystem::relative` but I wasn't sure whether this would work in a cross-platform way. I can switch to this though if this should work on Windows as well.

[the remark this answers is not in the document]
joeyparrish: It should... I haven't tested any of the snippets I wrote.  But std::filesystem is supposed to be platform-aware, and many of the examples show APIs that would only do useful things on Windows and not Linux.  (Like splitting off drive letters from paths, etc).

packager/mpd/base/xml/xml_node.cc
cosmin: That's where I started and that didn't work, the byte order was reversed, but looks like I was overthinking it. `StrFormat("%04X", ...)` works by itself, without `FromHost16`, going to big endian order is only important if we go byte at a time like the manual formatting.

packager/mpd/base/mpd_builder.cc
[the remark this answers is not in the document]
cosmin: ok, updated to use this, we'll see if the windows tests pass

packager/mpd/base/mpd_builder.h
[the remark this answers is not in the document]
cosmin: moved it to a separate file. This is used in the end to print ISO-8601 dates so I think it's better to keep this in std::tm instead of uint64_t timestamps. I updated the MuxerClock to use this Clock abstraction as well.

packager/mpd/base/mpd_builder_unittest.cc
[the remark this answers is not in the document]
cosmin: ok, updated TestClock to use ISO-8601 format as the input

packager/mpd/base/xml/xml_node.cc
cosmin: done

packager/mpd/base/mpd_utils_unittest.cc
```diff
@@ -122,8 +122,13 @@ TEST_F(MpdUtilsTest, ContentProtectionPlayReadyCencMspr) {
                                "98404286AB92E65BE0885F9500000001"
                                "11223344556677889900AABBCCDDEEFF"
                                "0000000430313233");
-    std::vector<uint8_t> pssh;
-    base::HexStringToBytes(pssh_str, &pssh);
+
+    std::string pssh_hex_str = absl::HexStringToBytes(pssh_str);
```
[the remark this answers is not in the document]
joeyparrish: I'd still like to clean this up.

packager/mpd/base/representation_unittest.cc
```diff
@@ -234,7 +235,7 @@ TEST_F(RepresentationTest, CheckVideoInfoVp9CodecInWebm) {
 }
 
 TEST_F(RepresentationTest, CheckVideoInfoLegacyVp9CodecInWebm) {
-  FLAGS_use_legacy_vp9_codec_string = true;
+  absl::SetFlag(&FLAGS_use_legacy_vp9_codec_string, true);
```
joeyparrish, line 238: To avoid polluting global state in a test, please use `FlagSaver` from `packager/flag_saver.h`.  See examples of usage in `packager/file/udp_options_unittest.cc` and `packager/file/file_unittest.cc`.

packager/mpd/base/xml/xml_node_unittest.cc
```diff
@@ -354,11 +353,13 @@ TEST(XmlNodeTest, AddAC4AudioInfoMPEGSchemeIMS) {
 class LiveSegmentTimelineTest : public ::testing::Test {
  protected:
   void SetUp() override {
-    FLAGS_segment_template_constant_duration = true;
+    absl::SetFlag(&FLAGS_segment_template_constant_duration, true);
```
joeyparrish, line 356: Same here.  Please use FlagSaver.  Since this is SetUp, you can make it a member variable.

packager/mpd/base/xml/xml_node_unittest.cc
```diff
@@ -354,11 +353,13 @@ TEST(XmlNodeTest, AddAC4AudioInfoMPEGSchemeIMS) {
...
     media_info_.set_segment_template_url("$Number$.m4s");
   }
 
-  void TearDown() override { FLAGS_segment_template_constant_duration = false; }
+  void TearDown() override {
+    absl::SetFlag(&FLAGS_segment_template_constant_duration, false);
```
joeyparrish, line 361: This becomes unnecessary with FlagSaver.

packager/mpd/base/xml/xml_node_unittest.cc
```diff
@@ -536,7 +537,7 @@ TEST_F(LiveSegmentTimelineTest, LastSegmentNumberSupplementalProperty) {
       {kStartTime, kDuration, kRepeat},
   };
   RepresentationXmlNode representation;
-  FLAGS_dash_add_last_segment_number_when_needed = true;
+  absl::SetFlag(&FLAGS_dash_add_last_segment_number_when_needed, true);
```
joeyparrish, line 540: Use it here, too.

packager/mpd/base/xml/xml_node_unittest.cc
```diff
@@ -549,7 +550,7 @@ TEST_F(LiveSegmentTimelineTest, LastSegmentNumberSupplementalProperty) {
                    "  <SegmentTemplate media=\"$Number$.m4s\" "
                    "                   startNumber=\"1\" duration=\"100\"/>"
                    "</Representation>"));
-  FLAGS_dash_add_last_segment_number_when_needed = false;
+  absl::SetFlag(&FLAGS_dash_add_last_segment_number_when_needed, false);
```
joeyparrish, line 553: This becomes unnecessary.

joeyparrish (changes requested):

packager/mpd/base/representation_unittest.cc
cosmin: done

joeyparrish (approved):

PR #1238, opened 2023-07-17
feat: port HLS module to CMake

cosmin:
Related to issue #1047


cosmin:
rebased now that cmake port of mpd has landed, next I'll need to address some of the same feedback from the MPD branch here (starting with FlagSaver)

packager/hls/base/simple_hls_notifier_unittest.cc
```diff
@@ -674,8 +671,8 @@ class WidevineSimpleHlsNotifierTest : public SimpleHlsNotifierTest,
  protected:
   WidevineSimpleHlsNotifierTest()
       : enable_legacy_widevine_hls_signaling_(GetParam()) {
-    FLAGS_enable_legacy_widevine_hls_signaling =
-        enable_legacy_widevine_hls_signaling_;
+    absl::SetFlag(&FLAGS_enable_legacy_widevine_hls_signaling,
```
cosmin, line 674: use flag saver here

packager/hls/base/simple_hls_notifier.cc
```diff
@@ -253,10 +256,8 @@ bool HandleWidevineKeyFormats(
 
 bool WriteMediaPlaylist(const std::string& output_dir,
                         MediaPlaylist* playlist) {
-  std::string file_path =
-      FilePath::FromUTF8Unsafe(output_dir)
-          .Append(FilePath::FromUTF8Unsafe(playlist->file_name()))
-          .AsUTF8Unsafe();
+  std::filesystem::path file_path =
+      std::filesystem::path(output_dir) / playlist->file_name();
```
cosmin, line 260: std::filesystem::u8path

packager/hls/base/simple_hls_notifier.cc
```diff
@@ -41,51 +40,51 @@ const char kWidevineDashIfIopUUID[] =
...
-std::string MakePathRelative(const std::string& media_path,
-                             const FilePath& parent_path) {
-  FilePath relative_path;
-  const FilePath child_path = FilePath::FromUTF8Unsafe(media_path);
-  const bool is_child =
-      parent_path.AppendRelativePath(child_path, &relative_path);
-  if (!is_child)
-    relative_path = child_path;
-  return relative_path.NormalizePathSeparatorsTo('/').AsUTF8Unsafe();
+// TODO: this is duplicated with MPD module
```
cosmin, line 70: this seems like a good time to handle these TODO's and move this out of MPD module somewhere shared we can use in both

packager/hls/base/simple_hls_notifier_unittest.cc
cosmin: done

packager/hls/base/simple_hls_notifier.cc
cosmin: done

packager/hls/base/simple_hls_notifier.cc
cosmin: moved to file_util.h

cosmin:
My `clang-format` locally and the Github Actions clang-format seem to be doing different things for this change in `GenerateSegmentUrl`. For now I applied the patch out of the failing Github Actions check locally.

joeyparrish (approved):

joeyparrish:
Looks good, except for Windows build failures like:

`packager\hls\base\simple_hls_notifier.cc(239,16): error C2664: 'bool shaka::hls::MediaPlaylist::WriteToFile(const std::string &)': cannot convert argument 1 from 'std::filesystem::path' to 'const std::string &'`

cosmin:
the problem is that on Windows there is no implicit conversion of std::filesystem::path to std::string (instead on Windows it can convert to std::wstring) whereas on Linux/Mac there is. Going to take a few iterations to flush out all usages by waiting for the Windows checks to run.

cosmin:
Fixing docs build in #1257

PR #1239, opened 2023-07-17
test: Fix data directory computation

joeyparrish:
The `__FILE__` macro does not always get you an absolute path, so our assumptions about it in test_data_util.cc were flawed.

If `foo.c` references `__FILE__`, something like `gcc -c foo.c`, will define `__FILE__` as `"foo.c"`.  If you use `gcc -c /path/to/foo.c`, then `__FILE__` will be `"/path/to/foo.c"`.

The Ninja generator for CMake only generates absolute paths for source files in certain CMake versions.  (Exact range unknown.)  Rather than require newer CMake versions and depend on CMake's Ninja generator maintaining the latest behavior forever, set the macro TEST_DATA_DIR to point to the test data folder.  This is consistent and never depends on unspecified behavior.

This change will allow the use of the Ninja generator with older versions of CMake, as found in most of our Docker builds.

### DIFF
--- a/packager/media/test/CMakeLists.txt
+++ b/packager/media/test/CMakeLists.txt
@@ -6,8 +6,14 @@
 
 add_library(test_data_util STATIC
     test_data_util.cc)
+target_compile_definitions(test_data_util
+    PRIVATE
+    # We used to build off of __FILE__, but that is not always an absolute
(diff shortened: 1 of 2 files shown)


cosmin (approved):
Seems much cleaner than relying on __FILE__

PR #1241, opened 2023-07-17
ci: Enable parallel builds

joeyparrish:
Our workflows and Dockerfiles now use Ninja on Linux & Mac, which enables safe parallel builds.  This significantly speeds up our workflows.

GitHub Actions typical compilation times (build step only):
 - Linux 19m => 9m
 - macOS 23m => 8m
 - Windows 12m => 10m
 - Linux arm64 (self-hosted) 72m => 29m
 - Docker build 25m => 14m

Overall workflow time: 84m => 33m

Compilation time on my workstation (12 CPUs @3.3GHz): 15m => 3m

This also adds a new environment variable "PACKAGER_LOW_MEMORY_BUILD".  If defined when CMake is first run, this will configure the build to disable parallel linking to reduce memory usage.  This helps us avoid failures on our self-hosted arm64 machines, where 6 CPUs share 4GB of RAM.

NOTE: Parallel builds are **NOT** recommended with Unix Makefiles due to the use of excessive RAM during parallel linking.  Unix Makefiles, unlike Ninja, cannot be configured to restrict parallel linking during a parallel build.  Anecdotally, parallel builds with Makefiles have exhausted a system with 32GB RAM.  (My workstation.)

In a follow-up, I will update the build documentation to refer to CMake and recommend all of the flags now used in our workflows.

### DIFF
--- a/.github/workflows/build-matrix.json
+++ b/.github/workflows/build-matrix.json
@@ -5,19 +5,22 @@
       "os": "ubuntu-latest",
       "os_name": "linux",
       "target_arch": "x64",
-      "exe_ext": ""
+      "exe_ext": "",
+      "generator": "Ninja"
     },
     {
       "os": "macos-latest",
       "os_name": "osx",
       "target_arch": "x64",
-      "exe_ext": ""
+      "exe_ext": "",
+      "generator": "Ninja"
     },
     {
       "os": "windows-latest",
       "os_name": "win",
       "target_arch": "x64",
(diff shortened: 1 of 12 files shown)


cosmin (approved):
Such a great improvement in overall execution time. If only Github Actions added support for arm64 natively.

joeyparrish:
I checked yesterday, and arm64 mac VMs are still on GitHub's roadmap for this year.  So at least we'll have that, in theory.

PR #1242, opened 2023-07-17
test: Avoid crashing a test executable

joeyparrish:
LOG(FATAL) aborts the executable, which means gtest can't write any reports it might be configured to write when tests are complete.  This will interfere with reporting.

This converts the only use of LOG(FATAL), which was in ReadTestDataFile(), to LOG(ERROR).  This also updates test cases to avoid crashing when ReadTestDataFile() returns an empty buffer.

### DIFF
--- a/packager/media/formats/mp4/mp4_media_parser_unittest.cc
+++ b/packager/media/formats/mp4/mp4_media_parser_unittest.cc
@@ -103,9 +103,13 @@ class MP4MediaParserTest : public testing::Test {
 
   bool ParseMP4File(const std::string& filename, int append_bytes) {
     InitializeParser(NULL);
+
     if (!parser_->LoadMoov(GetTestDataFilePath(filename).AsUTF8Unsafe()))
       return false;
+
     std::vector<uint8_t> buffer = ReadTestDataFile(filename);
+    ASSERT_FALSE(buffer.empty());
+
     return AppendDataInPieces(buffer.data(), buffer.size(), append_bytes);
   }
 };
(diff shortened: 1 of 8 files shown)


joeyparrish:
The issue isn't an empty file.  It's that the failure mode is to return an empty buffer, where `->data() == NULL`.  Some tests were feeding this pointer directly to a parser, where the parser would `DCHECK(ptr)` and crash the executable.  While this DCHECK() behavior may make sense for a debug build of the library/main executable, tests want to fail without crashing.

You can't call `ASSERT_` macros from inside a helper or library method, because they trigger an early return without a value.  So they can only be called from void functions (won't compile elsewhere) and if you call them from a void helper, it ends the helper, not the test.

So the options seemed to be either to make every test `ASSERT_FALSE(buffer.empty())`, which some already did, or to remove `DCHECK()` from parsers and make them fail gracefully on empty buffers or NULL pointers.

packager/media/formats/mp4/mp4_media_parser_unittest.cc
```diff
@@ -103,9 +103,13 @@ class MP4MediaParserTest : public testing::Test {
 
   bool ParseMP4File(const std::string& filename, int append_bytes) {
     InitializeParser(NULL);
+
     if (!parser_->LoadMoov(GetTestDataFilePath(filename).AsUTF8Unsafe()))
       return false;
+
     std::vector<uint8_t> buffer = ReadTestDataFile(filename);
+    ASSERT_FALSE(buffer.empty());
```
cosmin, line 111: some of the introduced usages are not in void functions, such as this one. I think instead of ASSERT_FALSE we use EXPECT_FALSE, but at that point couldn't we put this inside of `ReadTestDataFile` directly?

joeyparrish: See PR #1247

PR #1243, opened 2023-07-17
ci: Fix workflow warnings

joeyparrish:
Fixes the following warnings from GitHub Actions:
 - "The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/"
 - "The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/"

### DIFF
--- a/.github/workflows/test-linux-distros.yaml
+++ b/.github/workflows/test-linux-distros.yaml
@@ -29,7 +29,7 @@ jobs:
     outputs:
       MATRIX: ${{ steps.configure.outputs.MATRIX }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           ref: ${{ inputs.ref }}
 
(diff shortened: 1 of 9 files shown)


.github/workflows/test-linux-distros.yaml
```diff
@@ -44,7 +44,9 @@ jobs:
           });
 
           // Output a JSON object consumed by the build matrix below.
-          console.log(`::set-output name=MATRIX::${ JSON.stringify(matrix) }`);
+          fs.writeFileSync(process.env.GITHUB_OUTPUT,
+              `MATRIX=${ JSON.stringify(matrix) }`,
+              {flag: 'a'});
```
kqyang, line 49: OOC, what does "flag: 'a'" mean? append?

cosmin (approved):

kqyang (approved):

.github/workflows/test-linux-distros.yaml
joeyparrish: Yes, append.  It's the nodejs fs API's equivalent of `fopen(..., "a")`

PR #1244, opened 2023-07-17
ci: Collect and report test results

joeyparrish:
All gtest-based tests should now use add_gtest.  Results from these tests will be surfaced more prominently in GitHub Actions.  Before this, it was hard to find test failures in the log.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -152,6 +152,12 @@ jobs:
       - name: Test
         run: ctest -C "${{ matrix.build_type }}" -V --test-dir build/
 
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@150e2f992e4fad1379da2056d1d1c279f520e058
+        if: ${{ always() }}
+        with:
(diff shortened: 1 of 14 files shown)


joeyparrish:
For an example of the output of this change, see:
 - https://github.com/joeyparrish/shaka-packager/actions/runs/5580552439/attempts/1#summary-15112736524
 - https://github.com/joeyparrish/shaka-packager/pull/5/files

kqyang (approved):

cosmin:
One of the windows tests failed, it looks like the test web server was not able to get started. I'm wondering if this is transient if it happens to pick a bad port. Re-running it now to see if passes on retry.

joeyparrish:
> One of the windows tests failed, it looks like the test web server was not able to get started. I'm wondering if this is transient if it happens to pick a bad port. Re-running it now to see if passes on retry.

Hrm... I wonder if we need to make the tests smarter about this.  We could have TestWebServer try a few random ports and make the base URL available through a getter, so that the full URL with port number doesn't have to be hard-coded in the tests.

PR #1245, opened 2023-07-18
fix: Fix libpng include directories

joeyparrish:


joeyparrish:
I verified this on all platforms with an extra commit that linked file_unittest against libpng and included "png.h" in file_unittest.cc.  I then removed that extra commit after all the tests passed.  Here is that pass: https://github.com/shaka-project/shaka-packager/actions/runs/5583017925/jobs/10202920120?pr=1245

cosmin (approved):

PR #1260, opened 2023-08-29
feat: port demuxer and app to cmake

cosmin:
Work in progress pull request to finish the rest of the CMake conversion. This builds upon the cmake port of media/formats pull request #1258 . Submitting early to get feedback from CI on any cross-platform build issues on the remaining modules.


packager/app/libcrypto_threading.cc
```diff
@@ -6,46 +6,19 @@
...
-      id, static_cast<unsigned long>(base::PlatformThread::CurrentId()));
-}
-
-}  // namespace
-
 LibcryptoThreading::LibcryptoThreading() {
-  global_locks.reset(new base::Lock[CRYPTO_num_locks()]);
-  CRYPTO_THREADID_set_callback(ThreadIdFunction);
-  CRYPTO_set_locking_callback(LockFunction);
+  //    mbedtls_threading_set_alt();
```
joeyparrish, line 17: Do we need this class at all now?

packager/CMakeLists.txt
```diff
@@ -64,4 +65,115 @@ add_subdirectory(third_party)
...
+  demuxer
+  media_event
+  dvb
+  mp2t
+  mp4
+  packed_audio
+  ttml
+  formats_webm
+  wvm
+#  media_public
```
joeyparrish, line 105: What's up with this?

packager/media/demuxer/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+# Copyright 2017 Google LLC. All rights reserved.
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file or at
+# https://developers.google.com/open-source/licenses/bsd
+
+add_library(demuxer STATIC
```
joeyparrish, line 7: I don't see the gyp files for this library deleted yet

packager/CMakeLists.txt
```diff
@@ -64,4 +65,115 @@ add_subdirectory(third_party)
 add_subdirectory(tools)
 add_subdirectory(utils)
 add_subdirectory(version)
-add_subdirectory(app)
+
+set(libpackager_sources
```
joeyparrish, line 69: I don't see the gyp files for this library deleted yet

joeyparrish (changes requested):

packager/app/libcrypto_threading.cc
cosmin: doesn't look like this is needed anymore, cleaning it up

packager/CMakeLists.txt
cosmin: I should have cleaned that up, there isn't any code to link against for media_public

cosmin:
I'm still trying to debug the shared library build issues on Windows, will re-submit for review once that's working.

packager/file/CMakeLists.txt
```diff
@@ -29,6 +29,10 @@ target_link_libraries(file
     status
     version)
 
+if(LIBPACKAGER_SHARED)
+    target_compile_definitions(file PUBLIC SHAKA_IMPLEMENTATION)
```
cosmin, line 33: wince we have a SHAKA_EXTERNAL definition in file I think this is needed

packager/status/status.h
```diff
@@ -153,7 +153,7 @@ class SHAKA_EXPORT Status {
   // generated copy constructor and assignment operator.
 };
 
-SHAKA_EXPORT std::ostream& operator<<(std::ostream& os, const Status& x);
+std::ostream& operator<<(std::ostream& os, const Status& x);
```
cosmin, line 156: I had to remove SHAKA_EXPORT from this line to fix the duplicate symbol errors when building shared library mode on Windows. Not sure whether this is going to break any users of shared libpackager?

packager/app/mpd_generator.cc
```diff
@@ -87,41 +89,42 @@ ExitStatus RunMpdGenerator() {
...
+    LOG(ERROR) << "Failed to write MPD to " << absl::GetFlag(FLAGS_output);
     return kFailedToWriteMpdToFileError;
   }
 
   return kSuccess;
 }
 
 int MpdMain(int argc, char** argv) {
-  base::AtExitManager exit;
+  //  base::AtExitManager exit;
```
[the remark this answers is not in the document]
joeyparrish: I think it was deleted.  The most recent reference I see to AtExit was from 2017, where it said "Needed by base::WorkedPool used in ThreadedIoFile".  So I suspect it was always used to satisfy some part of chromium base.

packager/status/status.h
joeyparrish: Hrm... interesting.  I can't say what people are or aren't using with the library.  But it seems like a useful ability to output a Status object for logging or debugging.

Sometimes duplicate symbol errors occur when a definition is inline in a header, but I see that isn't the case here.  Could status.o be included in multiple shared libraries, or both in the shared library and an executable?

cosmin: The implementation is fairly trivial though, it just calls .ToString() on the status object. 

I did try to figure out why it's breaking though, I checked to see whether packager executable and libpackager both pull in status for example (or the test suite) and didn't have any luck. It's also only breaking on Windows (shared) specifically, none of the other platforms have this issue. It seems related to the dllimport / dllexport tagging going on here but I'm not too familiar with these Microsoft extensions.

Perhaps someone that knows more about Windows development can shed some light on it. But since the implementation of << is fairly trivial I think it's ok to drop this for now until we can find a proper solution.

joeyparrish: Agreed.  I wanted to do a major-version bump with the first cmake release anyway, since I don't know that we can guarantee ABI compat after all these changes.

packager/app/packager_main.cc
```diff
@@ -296,9 +331,18 @@ bool ParseProtectionSystems(const std::string& protection_systems_str,
...
+  auto tokens = absl::StrSplit(absl::AsciiStrToLower(protection_systems_str),
+                               ',', absl::SkipEmpty());
+  std::vector<std::string> protection_systems_vec;
+  for (const absl::string_view& token : tokens) {
+    std::string trimmed = std::string(token);
+    absl::StripAsciiWhitespace(&trimmed);
+    if (!trimmed.empty()) {
+      protection_systems_vec.push_back(trimmed);
+    }
+  }
```
joeyparrish, line 343: I feel like I've seen this pattern a few times now.  Maybe this could be factored out into a utility function somewhere?

packager/app/vlog_flags.h
```diff
@@ -7,9 +7,10 @@
 #ifndef APP_VLOG_FLAGS_H_
 #define APP_VLOG_FLAGS_H_
 
-#include <gflags/gflags.h>
+#include <absl/flags/declare.h>
+#include <absl/flags/flag.h>
 
-DECLARE_int32(v);
-DECLARE_string(vmodule);
+// ABSL_DECLARE_FLAG(int32_t, v);
```
joeyparrish, line 13: Still WIP?

joeyparrish (changes requested):

packager/app/vlog_flags.h
cosmin: Going to kill this now, and circle back on logging after using abseil log.

packager/file/CMakeLists.txt
cosmin: moved it to macros.h, trying to include packager.h would be more complicated due to everything else it defines

joeyparrish (approved):

joeyparrish:
Please see if you can fix the build errors before merging.  Looks like maybe just some missing headers.

PR #1264, opened 2023-09-18
fix: apply cmake fixes after merge of #1210

cosmin:
build in CMake branch is broken at least locally on OS X after #1210 was merged which introduced some new warnings and reintroduced base::nullopt.


joeyparrish (approved):

joeyparrish:
There are a couple failures on ArchLinux building in docker.  I'll try to take a look and offer a fix.

PR #1269, opened 2023-09-27
feat: add install target for cmake

cosmin:
this adds an install target which will install binaries, libraries, pkg-config and headers.

For example to customize the install prefix

```
cmake .. -G Ninja -DINSTALL_GTEST=OFF -DCMAKE_INSTALL_PREFIX=/opt/shaka
```

then running

```
cmake --install .
```

will do the following

```
-- Installing: /opt/shaka/bin/mpd_generator
-- Installing: /opt/shaka/bin/packager
-- Installing: /opt/shaka/lib/libpackager.a
-- Installing: /opt/shaka/include/file.h
-- Installing: /opt/shaka/include/macros.h
-- Installing: /opt/shaka/include/packager.h
-- Installing: /opt/shaka/include/status.h
-- Installing: /opt/shaka/bin/pssh-box.py
-- Installing: /opt/shaka/lib/pkgconfig/packager.pc
```

I don't really use libpackager directly (yet) so need to double check whether the include headers are comprehensive set needed for the public API. The list can be controlled via `LIBPACKAGER_HEADER_FILES`.


joeyparrish:
> -- Installing: /opt/shaka/include/file.h
> -- Installing: /opt/shaka/include/macros.h
> -- Installing: /opt/shaka/include/packager.h
> -- Installing: /opt/shaka/include/status.h

In the case of /usr/local/, I feel like /usr/local/include/file.h could be a problem.  It's such a simple name.

I think we should make sure that includes go into $PREFIX/include/packager/ to keep them from conflicting with anything.

joeyparrish:
> I don't really use libpackager directly (yet) so need to double check whether the include headers are comprehensive set needed for the public API.

I don't either.  I think we might need an automated test that installs libpackager and headers to some temp folder, then links some very simple test app using that prefix.

packager/CMakeLists.txt
```diff
@@ -86,6 +86,13 @@ else()
   add_library(libpackager STATIC ${libpackager_sources})
 endif()
 
+set(LIBPACKAGER_HEADER_FILES
+  file/file.h
+  macros.h
+  packager.h
```
joeyparrish, line 92: Yeah, this is definitely not a complete set of headers.  You can just look at the list of `#include`s in `packager.h`.

I'll create a test for this.

joeyparrish:
I merged in the other changes, converted all internal headers to angle-brackets, moved our header install to a subdirectory, and added a test application to link against shared libpackager.  I'm getting linter failures now, and based on that, I'm going to split this up a bit and fix a bunch of style issues separately since the fixes touch almost every file and are mostly mechanical changes.

joeyparrish:
See PR #1271 (angle brackets for packager/...) and PR #1273 (include order cleanup).  I'll rebase this after those land.

joeyparrish:
Rebased.  We have a target now in this PR that depends on the installed library and headers working.  The headers don't work, though, so that part needs to be fixed.  But at least there is an automated test for it in the build system.

joeyparrish:
Rebased and the link test redone on top of #1275.

joeyparrish (approved):

joeyparrish:
As far as I can tell, the test install doesn't happen on Windows.  At least, there's no log of it.  Then when the linker test runs, there's no headers installed.

Perhaps the dependency between those two steps is broken on Windows?  I'm really not sure.  Until I can debug it live on a Windows machine, I'm not sure what else to do.

cosmin:
I'll try on the Windows laptop I setup for this purpose, let's see how it goes.

cosmin:
I was able to repro this locally, although the error is the same (just the path is different).

```
  CMake Error at packager/cmake_install.cmake:89 (file):
    file INSTALL cannot find
    "<...>/shaka-packager/build/packager/Release/mpd_generator.exe":
```

And sure enough if I look in the Release folder there is no mpd_generator.exe

```
10/15/2023  09:40 PM         4,932,608 libpackager.dll
10/15/2023  09:40 PM           145,599 libpackager.exp
10/15/2023  09:40 PM           243,768 libpackager.lib
10/15/2023  09:40 PM           467,968 packager.exe
10/15/2023  09:40 PM             3,761 packager.exp
10/15/2023  09:40 PM             6,644 packager.lib
```

So looks like the install target is being called before `mpd_generator` was built, and it fails.

PR #1275, opened 2023-10-13
chore: Refactor public headers and macros

joeyparrish:
This organizes all public headers for the library into
packager/public/ and makes sure then don't rely on any headers from
other folders.

To accomplish this, this change also refactors macros.h,
media/base/macros.h, and status/status_macros.h into macros/classes.h,
macros/compiler.h, macros/crypto.h, macros/logging.h, macros/status.h,
and public/export.h.  Now the export macros from macros.h live in
public/ to keep public/ from requiring anything else.

This refactor enables an install target that includes public headers
only.


joeyparrish:
Includes #1274, so I'll rebase this after than lands.

joeyparrish:
Rebased and ready for review.  There are a ton of changes to `#include`s all over the codebase, so it will probably be easier just to examine macros/* and public/* to get a good sense of the refactor.

cosmin (approved):

PR #1278, opened 2023-10-17
build: Fix rebuild of link-test targets

joeyparrish:
The link-test targets rebuilt every time you built the project because that is the behavior of commands attached to add_custom_target.  This is fixed by combining add_custom_command and add_custom_target.

Fixing that led to a situation where the link-test targets built on the first build, but didn't properly rebuild when their dependencies changed.  This is fixed by using set_source_file_properties to treat test.cc as dirty when the test-install command runs.

### DIFF
--- a/link-test/CMakeLists.txt
+++ b/link-test/CMakeLists.txt
@@ -11,16 +11,30 @@ if(LIBPACKAGER_SHARED)
   # Install the library and headers to a temporary location.
   set(TEST_INSTALL_DIR ${CMAKE_BINARY_DIR}/test-install)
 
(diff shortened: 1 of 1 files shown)


cosmin (approved):

joeyparrish:
I really needed to see that work on Windows before merging!

cosmin:
Sorry, forgot that over here merge is immediate, at work hitting land means "after tests finish, assuming they all pass". I'm testing locally on Windows now to confirm.

cosmin:
Confirmed locally that it works on Windows.

PR #1279, opened 2023-10-17
build: Fix pssh-box.py targets, installation, and docs

joeyparrish:
pssh-box.py needs Python protos, which we were not building.  It also needs to have those protos installed properly.

This fixes generation of Python proto interfaces and fixes the install targets to put those protos where they are needed.  This also updates the documentation to match.

### DIFF
--- a/packager/tools/pssh/README.md
+++ b/packager/tools/pssh/README.md
@@ -3,7 +3,7 @@ pssh-box - Utility to parse and generate PSSH boxes
 
 ## Prerequisite
 
-- Python 2.6 or newer.
+- Python 3.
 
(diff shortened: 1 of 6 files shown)


packager/tools/pssh/README.md
```diff
@@ -3,7 +3,7 @@ pssh-box - Utility to parse and generate PSSH boxes
 
 ## Prerequisite
 
-- Python 2.6 or newer.
+- Python 3.
```
cosmin, line 6: is there a specific version of Python 3.x+ we want to call out here?

cosmin (approved):

PR #1280, opened 2023-10-17
feat: Implement stream ordering in HLS playlist based on input order.

### LINKED ISSUE #560: Order of variants in HLS master playlist - can this be customized?
The Apple HLS authoring spec says:

1.32. The default video variant(s) SHOULD be the 2000 kb/s (average bit rate) variant. (Defaults are the first variant listed in the master playlist within a group of variants having compatible audio.)

I'm trying to set the order for the EXT-X-STREAM-INF's in the master.m3u8 so that my 2000 kb/s appears first but no matter what I try I cannot manipulate it so it appears in the top spot.   I've tried reordering my inputs, different naming (noting the std::sort()) in packager.cc, and various other ploys to try to trick shaka packager into ordering things the way I want - but to no avail.  Shaka packager always puts my lowest resolution and bandwidth stream first in the hls master playlist.

I'm probably missing something obvious but could someone point me in the right direction as to how I can manipulate the ordering.

Thank you for your help (and amazing product!).

wjywbs:
Fixed #560 with suggested implementation:
https://github.com/google/shaka-packager/issues/560#issuecomment-611336440

### DIFF
--- a/packager/app/packager_main.cc
+++ b/packager/app/packager_main.cc
@@ -540,6 +540,7 @@ int PackagerMain(int argc, char** argv) {
         ParseStreamDescriptor(argv[i]);
     if (!stream_descriptor)
       return kArgumentValidationFailed;
+    stream_descriptor->output_order = i;
     stream_descriptors.push_back(stream_descriptor.value());
   }
   Packager packager;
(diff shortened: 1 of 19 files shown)


joeyparrish:
Closing this in favor of https://github.com/shaka-project/shaka-packager/pull/1329. Thank you!

PR #1281, opened 2023-10-17
fix: Update to use official FairPlay UUID.

wjywbs:
Reference: https://github.com/Dash-Industry-Forum/Identifiers/issues/119


cosmin (approved):
Cross checked with the published version at https://dashif.org/identifiers/content_protection/ and this looks correct

PR #1282, opened 2023-10-18
fix: Fix static executable build on Linux

joeyparrish:


joeyparrish:
The failures here should be resolved by #1285.  I'll rebase after that lands.

joeyparrish:
Ready for review again, assuming CI passes this time

PR #1285, opened 2023-10-18
feat: Build c-ares locally

joeyparrish:
c-ares (used on Linux only) was an exception to the rule of only linking against internally-built libraries.  This fixes that, to support a truly static build of packager.


joeyparrish:
Should only impact Linux builds, but I'd like to let the CI run before merging just in case.

joeyparrish:
The macOS failure was temporary, and is fixed by #1286

cosmin (approved):

PR #1286, opened 2023-10-18
fix: Use stable GitHub mirror for libxml2 source

joeyparrish:
The upstream repo at gnome.org sometimes fails in CI


joeyparrish:
For example, see https://github.com/shaka-project/shaka-packager/actions/runs/6565813257/job/17835185145?pr=1285

cosmin (approved):

PR #1291, opened 2023-10-19
fix: Fix dependency on system zlib

joeyparrish:


joeyparrish:
A slimmed version of the previous PR.  No -static in the linker flags for now.

cosmin (approved):

PR #1293, opened 2023-10-19
build: Separate shared and static library targets

joeyparrish:
This splits the target so that we always build static libraries, but optionally also build shared libraries.  This also tweaks the installation so that the static library is never installed, because it is not usable without the other internal static library deps.  The headers and pkgconfig will only be installed if we have a shared library to use them with.

### DIFF
--- a/packager/CMakeLists.txt
+++ b/packager/CMakeLists.txt
@@ -86,16 +86,9 @@ set(libpackager_sources
   app/single_thread_job_manager.h
   packager.cc
   ../include/packager/packager.h
-  )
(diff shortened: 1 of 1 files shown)


joeyparrish:
@cosmin, I'm testing release workflows on my fork right now, and those PRs depend on this one.  If you don't like it, please let me know so I can make adjustments and finish testing those workflow PRs.

cosmin (approved):
good enough for now, we can revisit later and see if we can get a proper standalone static libpackager

Issue #1316, opened 2023-12-08
Packaging fails on MPEG-TS with SAMPLE-AES 


chenda6:
## System info
macOS Monterey 12.6
(cmake) https://github.com/shaka-project/shaka-packager/tree/cmake-porting-history

## Issue and steps to reproduce the problem

**Packager Command**:
```
packager \                                                                                
        in=input.mp4,stream=video,segment_template=\$Number\$.ts,drm_label=sd \
        --enable_raw_key_encryption \                                                     
        --keys label=sd:key_id=abba271e8bcf552bbd2e86a434a9a5d9:key=69eaa802a6763af979e8d1940fb88392:iv=22334455667788990011223344556677 \
        --protection_scheme cbcs \                                                        
        --protection_systems FairPlay \                                                   
        --hls_master_playlist_output h264_master.m3u8 \                                   
        --hls_key_uri skd://testAssetID  
```

What is the expected result?
MPEG-TS segments successfully encrypted and packaged.

What happens instead?
Packager terminates early with following stack trace:
```
E1208 10:04:23.204043       2 aes_encryptor.cc:145] Expecting output size of at least 32 bytes.
F1208 10:04:23.204063       2 encryption_handler.cc:400] Check failed: encryptor_->Crypt(source, source_size, dest, &dest_size) 
*** Check failure stack trace: ***
    @        0x1085cef60  absl::lts_20230125::log_internal::LogMessage::PrepareToDie()
    @        0x1085ce818  absl::lts_20230125::log_internal::LogMessage::SendToLog()
    @        0x1085cdb04  absl::lts_20230125::log_internal::LogMessage::Flush()
    @        0x1085cf42c  absl::lts_20230125::log_internal::LogMessageFatal::~LogMessageFatal()
    @        0x1085cf47c  absl::lts_20230125::log_internal::LogMessageFatal::~LogMessageFatal()
    @        0x107d41800  shaka::media::EncryptionHandler::EncryptBytes()
    @        0x107d40280  shaka::media::EncryptionHandler::ProcessMediaSample()
    @        0x107d3f148  shaka::media::EncryptionHandler::Process()
    @        0x108090244  shaka::media::MediaHandler::Dispatch()
    @        0x107c9fa28  shaka::media::MediaHandler::DispatchMediaSample()
    @        0x107c9ef50  shaka::media::ChunkingHandler::OnMediaSample()
    @        0x107c9e120  shaka::media::ChunkingHandler::Process()
[...]

PR #1317, opened 2023-12-13
fix: Proposed fix for #1316

synologic:
Fixes check in AesCbcEncryptor::CryptInternal which breaks CBCS encryption

### DIFF
--- a/packager/media/base/aes_encryptor.cc
+++ b/packager/media/base/aes_encryptor.cc
@@ -141,7 +141,8 @@ bool AesCbcEncryptor::CryptInternal(const uint8_t* plaintext,
   const size_t num_padding_bytes = NumPaddingBytes(plaintext_size);
   const size_t required_ciphertext_size = RequiredOutputSize(plaintext_size);
 
-  if (*ciphertext_size < required_ciphertext_size) {
+  // RequiredOutputSize returns the plaintext_size plus AES_BLOCK_SIZE as required by mbedtls, subtract that here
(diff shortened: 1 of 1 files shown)


cosmin:
The right fix is to update the definition of required ciphertext size. Currently it claims it needs to be one block larger than the real size because of implementation details mbedtls. This needs to be verified, and hopefully it's either not accurate, an alternative api could be used with mbedtls or we need to switch away from mbedtls if there is no workaround.

synologic:
Switching away from mbedtls it's probably not a 5 minute job. Not having Fairplay, currantly, is bad.
You're saying that porting AES to mbedtls was not done properly ?

cosmin:
Yes, there are some known issues currently with the mbedtls implementation in main, please use latest 2.6.1 release until this is resolved,

synologic:
For my needs, i will stick to this fix on the latest main.
There are notable differences (I am not sure how/when this changed) between 2.6.1 memory bandwidth usage and the latest main, but i'm seeing considerable drops in memory read/writes, as a matter of fact, memory traffic dropped about 3.5x just by using the latest main, which makes high density packaging much more efficient.

cosmin:
@synologic  what I believe to be the correct fix for this is in the `fix-integration-tests` if you want to give that a shot. Unless the current state of `main` that branch passes all integration tests so I have reasonable confidence that it should work for all scenarios. Due to the holidays it will take some time to get that reviewed and merged (and split into several smaller pull requests) but feel free to report any problems you find on that branch and I'll take a look.

PR #1321, opened 2023-12-25
fix: preserve case for stream descriptors

cosmin:
The accidental tolower in `SplitAndTrimSkipEmpty` was causing stream descriptors to not preserve case for certain things like accessibilities.


xavierlaffargue:
@joeyparrish can you add @cosmin as maintainer for Shaka packager? 
a single maintainer is not enough and causes a bottleneck.

cosmin:
I am a maintainer but every change should get reviewed anyway so I'm waiting for @joeyparrish to get back and approve a few pull requests that are blocking a 3.0 release.

joeyparrish (approved):

PR #1323, opened 2023-12-25
fix: remove extra block assumptions in mbedtls integration

cosmin:
The current mbedtls integration was not working for some modes. See for example #1316 and also lots of failing integration tests.

For example in pattern encryptor it works on one block at a time so it cannot assume it's going to always get a buffer with a padding for an extra block. 

From what I can tell when the padding mode is correctly set to `MBEDTLS_PADDING_NONE` there is no extra block being written to or required.

This passes all crypto unit tests and integration tests.


joeyparrish:
I want to check this with valgrind to make sure we're not overflowing any buffers, but then I will merge it.

joeyparrish:
I see no buffer overflows.  I caught some of my mistakes in mbedtls padding with the old system through valgrind, so I just wanted to double-check.

I did find some uninitialized values being used in branch decisions, though, so I'll follow-up with a PR to fix those.

joeyparrish (approved):

PR #1327, opened 2024-01-04
chore: upgrade to alpine 3.19

misiek08:
We wanted to use newest ffmpeg. Without switching to ubuntu static builds are not properly resolving DNS (glibc vs musl), so the easiest way is to upgrade alpine.


misiek08:
Not sure if `feat` or `chore`

CMakeLists.txt
```diff
@@ -9,6 +9,9 @@
 # Minimum CMake version.  This must be in the root level CMakeLists.txt.
 cmake_minimum_required(VERSION 3.16)
 
+# Musl define off_t and it's already 64bit
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Doff64_t=off_t")
```
joeyparrish, line 13: I don't think this is universally true on all platforms.  We shouldn't assume off_t is 64-bits, even if it's true in your case.

joeyparrish (changes requested):

joeyparrish:

CMakeLists.txt
joeyparrish: It appears that the use of the non-standard off64_t in absl was fixed upstream in https://github.com/abseil/abseil-cpp/commit/4500c2fada4e952037c59bd65e8be1ba0b29f21e .  Upgrading absl should resolve this.  Testing now.

joeyparrish: Resolved by #1338 and #1337

cosmin (approved):

joeyparrish (approved):

PR #1313, opened 2023-12-03
feat: Preserve order of variants in HLS master playlist

xavierlaffargue:
Feat #560

PR #1329, opened 2024-01-18
feat: order streams in manifest based on command-line order

### LINKED ISSUE #560: Order of variants in HLS master playlist - can this be customized?

SteveR-PMP:
This will force the muxer to order streams in the order given on the  command-line.

Closes #560
Closes #1280
Closes #1313


packager/app/manifest_flags.cc
```diff
@@ -37,3 +37,8 @@ ABSL_FLAG(std::string,
           "",
           "Same as above, but this applies to text tracks only, and "
           "overrides the default language for text tracks.");
+ABSL_FLAG(bool,
+          force_cl_index,
+          false,
```
joeyparrish, line 42: Let's default this to true.  We are about to release v3.0.0, where it is okay to introduce a change that is not strictly backward compatible.  I think some specific order is better than no specific order.

joeyparrish (changes requested):

joeyparrish (changes requested):
Looks like it fails to build on Windows.  PTAL at the CI logs, one of which I copied here.  Also, apologies that the Windows build logs are so hard to sift through.

packager/hls/base/master_playlist.cc
```diff
@@ -437,15 +450,37 @@ void AppendPlaylists(const std::string& default_audio_language,
...
+  std::list<std::pair<std::string, std::list<const MediaPlaylist*>>>
+      audio_groups_list(audio_playlist_groups.begin(),
+                        audio_playlist_groups.end());
+  std::list<std::pair<std::string, std::list<const MediaPlaylist*>>>
+      subtitle_groups_list(subtitle_playlist_groups.begin(),
+                           subtitle_playlist_groups.end());
+  if (force_cl_index) {
+    audio_groups_list.sort(GroupOrderFn);
+    for (const auto& group : audio_groups_list) {
+      std::list<const MediaPlaylist*> playlists = group.second;
```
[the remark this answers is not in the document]
joeyparrish: This may be the only error, or it may be that the build was aborted before other errors could be seen.

packager/app/manifest_flags.cc
joeyparrish: I edited this to set the default to true for the CLI.  Let me know if you disagree.

packager/hls/base/master_playlist.cc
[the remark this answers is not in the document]
joeyparrish: I made an edit to fix this.  I renamed the inner `playlists` variable for each group to `group_playlists`.

joeyparrish (approved):

joeyparrish:
@SteveR-PMP, if you are okay with the two edits I made, and if the builds and tests all pass, I will merge this for release in v3.0.0.

vish91:
@joeyparrish  @SteveR-PMP  one comment would be to change hls_params and other references initializing bool to true so that's default when the cli argument is not passed ?  
Also documentation change to explicitly note that this is the default behavior and set it to false to get old random ordering behavior

packager/app/manifest_flags.cc
```diff
@@ -37,3 +37,8 @@ ABSL_FLAG(std::string,
           "",
           "Same as above, but this applies to text tracks only, and "
           "overrides the default language for text tracks.");
+ABSL_FLAG(bool,
+          force_cl_index,
+          true,
+          "Force the muxer to order streams in the order given "
```
vish91, line 43: update comment

include/packager/hls_params.h
```diff
@@ -63,6 +63,9 @@ struct HlsParams {
   /// Custom EXT-X-MEDIA-SEQUENCE value to allow continuous media playback
   /// across packager restarts. See #691 for details.
   uint32_t media_sequence_number = 0;
+  /// Force the muxer to order streams in the order given
+  /// on the command-line
+  bool force_cl_index = false;
```
vish91, line 68: set to true by default

include/packager/mpd_params.h
```diff
@@ -102,6 +102,7 @@ struct MpdParams {
   /// and is greatly influnced by the player.
   /// This parameter is required by DASH-IF Low Latency standards.
   double target_latency_seconds = 1;
+  bool force_cl_index = false;
```
vish91, line 105: set to true by default

include/packager/packager.h
```diff
@@ -58,6 +58,9 @@ struct PackagingParams {
   /// Only use a single thread to generate output.  This is useful in tests to
   /// avoid non-deterministic outputs.
   bool single_threaded = false;
+  /// Force the command-line order in the output manifest(s)
+  bool force_cl_index = false;
```
vish91, line 62: set to true by default

packager/app/manifest_flags.cc
SteveR-PMP: I'm thinking a mention of the old behavior should be in the release notes and not in this flag description.

include/packager/hls_params.h
SteveR-PMP: yes, these bool values should also be set to 'true'.

joeyparrish (approved):

joeyparrish:
Looks like test outputs may need to be updated.

cosmin:
There may also be some bugs here, I'm looking at the test output and seeing some cases where adaptation set ID ends up being 4294967295 which might mean there are some code paths that are not setting this properly. I'm digging into it.

joeyparrish:
Thank you, @cosmin.  I would welcome any changes you want to push to the fork for this PR to correct these issues before we merge it.

cosmin (changes requested):
The current approach doesn't seem sound. The cli_index ends up overwriting explicitly supplied IDs which is why tests are breaking. This would not work for library users. Command line concerns should remain limited to the command line code and not leak into the implementation of adaptation sets.

cosmin:
I'm taking a look to see if there's a way to do this more cleanly.

cosmin:
@SteveR-PMP it should be generic, the underlying code may not be called from a command line and yet an index can be provided. The flag (on by default) is whether the command line interface will set the order, and nothing outside of the packager_main ought to look at the `force_cl_index` flag.

joeyparrish (commented):
LGTM % test failures

cosmin:
It builds locally, this usually means there's some missing include of probably optional somewhere, I'll track it down.

joeyparrish (approved):

PR #1330, opened 2024-01-18
feat: default text zero bias

SteveR-PMP:
A positive value, in milliseconds. It is the threshold used to determine if we should assume that the text stream actually starts at time zero. If the first sample comes before default_text_zero_bias_ms, then the start will be padded as the stream is assumed to start at zero. If the first sample comes after default_text_zero_bias_ms then the start of the stream will not be padded as we cannot assume the start time of the stream.


joeyparrish (approved):
Will merge after successful test pass

PR #1337, opened 2024-02-09
chore: Upgrade absl and protobuf

joeyparrish:
protobuf depends on absl, so both needed an update.

Updating absl fixes issues on Alpine 3.19 (see PR #1327), and also removes the need for hacks around vlog flags.


cosmin (approved):

cosmin:
Shall we revisit treating warnings as errors in 3rd party deps? It seems like a constant game of whack-a-mole every time we upgrade deps or toolchain.

joeyparrish:
Yes, I think you're right.  At the time I started down this path, I wanted to avoid any surprise warnings that might be meaningful things we should take action on... but third_party is big and we are not really reviewing them.

PR #1346, opened 2024-02-16
feat: add startwithSAP/subsegmentstartswithSAP for audio tracks

### LINKED ISSUE #364: startWithSAP & subsegmentStartsWithSAP attributes are not populated and this is not spec compliant
## System info
**Operating System**: all
**Shaka Packager Version**: v2.0.2 (and previous versions)

When ShakaPackager generates DASH presentation, it doesn't indicate startWithSAP & subsegmentStartsWithSAP attributes either on AdaptationSet or Representation elements.

This is not spec compliant. From ISO/IEC 23009-1:2014(E), section 8.3 (on-demand profile) :
Representation elements with a @subsegmentStartsWithSAP value absent, zero or
greater than 3 may be ignored.

and 8.4 (live profile):
Representation elements with a @startWithSAP value (either supplied directly or inherited
from the containing) absent, zero or greater than 3 may be ignored.

cosmin:
Replaces #1055

This pull request adds startwithSAP/subsegmentstartswithSAP for aac, ac3, ec3 and ac4 audio tracks according to LIVE or VOD profile.

Closes #364

### DIFF
--- a/packager/mpd/base/adaptation_set.cc
+++ b/packager/mpd/base/adaptation_set.cc
@@ -288,6 +288,14 @@ std::optional<xml::XmlNode> AdaptationSet::GetXml() {
       return std::nullopt;
     }
   }
+  if (subsegment_start_with_sap_ == 1) {
+    if (!adaptation_set.SetIntegerAttribute("subsegmentStartsWithSAP", 1))
+      return base::nullopt;
+  }
+  else if (start_with_sap_ == 1) {
+    if (!adaptation_set.SetIntegerAttribute("startWithSAP", 1))
+      return base::nullopt;
+  }
 
   if (video_frame_rates_.size() == 1) {
     suppress_representation_frame_rate = true;
@@ -412,6 +420,14 @@ void AdaptationSet::ForceSetSegmentAlignment(bool segment_alignment) {
 void AdaptationSet::AddAdaptationSetSwitching(
     const AdaptationSet* adaptation_set) {
   switchable_adaptation_sets_.push_back(adaptation_set);
+}
+
+void AdaptationSet::ForceSubsegmentStartswithSAP(uint32_t sap_value) {
+  subsegment_start_with_sap_ = sap_value;
+}
+
+void AdaptationSet::ForceStartwithSAP(uint32_t sap_value) {
+  start_with_sap_ = sap_value;
 }
 
 // For dynamic MPD, storing all start_time and duration will out-of-memory
--- a/packager/mpd/base/adaptation_set.cc
+++ b/packager/mpd/base/adaptation_set.cc
@@ -276,6 +276,7 @@ std::optional<xml::XmlNode> AdaptationSet::GetXml() {
       return std::nullopt;
     }
   }
+
   if (video_heights_.size() == 1) {
     suppress_representation_height = true;
     if (!adaptation_set.SetIntegerAttribute("height", *video_heights_.begin()))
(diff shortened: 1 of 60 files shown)


cosmin:
Puzzled by how that test is failing only on Windows release static.

joeyparrish:
> Puzzled by how that test is failing only on Windows release static.

I see an "actual" MPD containing: `<!--Generated with https://github.com/shaka-project/shaka-packager version 00fc734-release-->`, but that line is missing from "expected".

Perhaps the mechanism for stripping that is failing on a particular build.

packager/mpd/base/adaptation_set.cc
```diff
@@ -288,6 +288,13 @@ std::optional<xml::XmlNode> AdaptationSet::GetXml() {
       return std::nullopt;
     }
   }
+  if (subsegment_start_with_sap_ == 1) {
```
joeyparrish, line 291: Why store a uint8_t, then only use it if it's 1?  Other values are defined in the spec, such as 2.  Could this maybe set the integer attribute with any non-zero value?

joeyparrish (changes requested):
Still looking into the mystery test failure.

joeyparrish:
I'm going to test this theory about the failed test by running individual suites in separate invocations using a filter on the command-line.  If I'm right, I should be able to trigger a failure on any platform and build config using a filter.  This can also help hunt down other missing calls to SetPackagerVersionForTesting().

joeyparrish:
I was wrong.  First, I overlooked this line: `xml_compare.cc:128] Attributes of element AdaptationSet do not match.`

Second, xml_compare.cc seems not to care about comments or about the order of attributes.

The actual difference is:

```
Expected:

...
<AdaptationSet contentType="video" maxWidth="720" maxHeight="480" maxFrameRate="10/1" id="0">
...

Actual:

...
<AdaptationSet contentType="video" maxWidth="720" maxHeight="480" subsegmentStartsWithSAP="1" maxFrameRate="10/1" id="0">
...
```

That was very hard to spot, but actual has `subsegmentStartsWithSAP="1"`, and expected does not.

joeyparrish:
One of the failing tests is `OnDemandMpdBuilderTest.TwoVideosWithDifferentResolutions`, which compares output against `kFileNameExpectedMpdOutputVideo1And2` which is `video_media_info1and2_expected_mpd_output.txt`.  That file does not mention `subsegmentStartsWithSAP`... so how is the test passing?

I'll look into xml_compare more deeply.

joeyparrish:
I added some logs to the comparison and just ran the one test on Linux.  When it passes, I see that the actual MPD is missing `subsegmentStartsWithSAP`:

`<AdaptationSet contentType="video" maxWidth="720" maxHeight="480" maxFrameRate="10/1" id="0">`

This leads me to think we have an uninitialized value involved.

joeyparrish:
`packager/mpd/base/adaptation_set.h` doesn't initialize the new fields you added.  In a release build, where memory is generally not sanitized for you on allocation, they happen to have a starting value that is non-zero.

joeyparrish:
The test should be fixed now.

joeyparrish (approved):

packager/mpd/base/adaptation_set.cc
joeyparrish: I made this change and updated the PR, and then approved it.  If you agree with the change, feel free to merge.  Otherwise, you can undo this part.

PR #1348, opened 2024-02-21
fix: http_file: Close upload cache on task exit

### LINKED ISSUE #1347: Shaka hangs on http upload if curl does not read upload cache
## System info
**Operating System**: Ubuntu 20.04.6 LTS (Focal Fossa)
**Shaka Packager Version**: main, sha: 76eb2c1575-release

**Packager Command**:
```sh
packager -v=3 input=short_h264_video.mp4,stream=0,init_segment=output/init.mp4,segment_template=output/$Number%05d$.mp4 --segment_duration=10 --fragment_duration=10 --use_fake_clock_for_muxer=1 --io_block_size=65536 --time_shift_buffer_depth=50 --preserved_segments_outside_live_window=0 --mpd_output=http://localhost:43831/dash.mpd
```

(1) Setup a local web server to receive the manifest updates, and configure the webserver to respond with an error code (403 for example) on each request.
(2) Run repeatedly until shaka hangs, it appears this issue is timing sensitive.

What is the expected result?
Shaka exits unsuccessfully

What happens instead?
Shaka hangs.

Input file is probably not important here, but for reference  it was generated via:
```sh
ffmpeg -f lavfi -i testsrc=duration=10:size=1280x720:rate=30  -t 10s  sample-content/short_h264_video.mp4
```

Problem was originally discovered on shaka 2.6.1, attaching abbreviated verbose logs from that run (with some additional locally added logs). Issue is verified to reproduce on shaka main, but could not get the log verbosity to work well enough to extract logs there.

[shaka_hangs_abbrev.log](https://github.com/shaka-project/shaka-packager/files/14355083/shaka_hangs_abbrev.log)

From the logs it can be seen that, `HttpFile::Flush()` is entered but not exited:
```
stderr(372): [0220/063540:VERBOSE2:http_file.cc(265)] Flush enter
```

There are no logs from `HttpFile::CurlReadCallback()` and the main process has finished:
```
stderr(424): [0220/063540:VERBOSE1:status.cc(74)] 7 (HTTP_FAILURE): HTTP response code said error, response code: 403.
stderr(425): [0220/063540:VERBOSE2:http_file.cc(363)] ThreadMain about to finish
stderr(426): [0220/063540:VERBOSE2:http_file.cc(367)] ThreadMain finished
```

petzeb:
In some cases it can happen that the http server responds with a non-successful status code without reading the response body. In this case curl may decide not to read from the cache since there is really no point in sending data to the server. In case some other thread of shaka has already called HttpFile::Flush it may end up deadlocked there waiting for the cache to either close or become empty. Thus, we close the cache when leaving the main thread as no data will be read by curl after it has finished anyways.

Closes #1347

### DIFF
--- a/packager/file/http_file.cc
+++ b/packager/file/http_file.cc
@@ -372,6 +372,12 @@ void HttpFile::ThreadMain() {
         error_message);
   }
 
+  // In some cases it is possible that the server has already closed the
+  // connection without reading the request body. This can for example happen
+  // when the server responds with a non-successful status code. In this case we
+  // need to make sure to close the upload cache here, otherwise some other
+  // thread may block forever on Flush().
+  upload_cache_.Close();
   download_cache_.Close();
   task_exit_event_.Notify();
 }


joeyparrish:
This seems reasonable to me.  @cosmin, since you were looking at details of the HTTP upload implementation recently for another PR, can you comment on this one?

cosmin (approved):

PR #1353, opened 2024-02-26
build: Disable warnings in third-party code

joeyparrish:
We already disabled warnings-as-errors in third-party code, but this goes futher to completely disable third-party warnings.  They are mostly noise, and distracting.  If they aren't errors, we shouldn't have our build logs flooded with them.


cosmin (approved):

cosmin:
Thank you, I was about to suggest the same thing as I got tired of looking at random warnings in the build output.

PR #1370, opened 2024-03-12
fix: Fix tags in official Docker images and binaries

### LINKED ISSUE #1366: [minor] Fix the version number in the docker build
## System info
**Operating System**: macOS Ventura 13.2.1 (Apple Silicon)
**Shaka Packager Version**: v3.0.1

It might be an issue specific to the docker build.
https://hub.docker.com/r/google/shaka-packager/tags
Tag: v3.0.0 and v3.0.1

**Packager Command**:

```
docker run --rm google/shaka-packager:v3.0.1 packager --version
```

**What is the expected result?**
* A string with the version number v3.0.1 is printed.
* It's from a build in release mode.

Like back in tag v2.5.1:
```
packager version v2.5.1-9f11077-release
```

**What happens instead?**
Getting this output:
```
packager version unknown-version-debug
```

## Note
Note that this version string is also included in all the manifest files generated.

joeyparrish:
The release workflows did not run checkout with `fetch-tags: true`, so the builds were unable to compute the correct release version number. I audited all instances of `actions/checkout` to add `fetch-tags` where needed and clean up unneeded options.

I also had to fix options to `docker/build-push-action`, which by default ignores `actions/checkout` and tries to pull from git itself.  This led to the Docker build running in a context without the new tag.

Finally, to make verification easier and provide version info in the build logs, this adds debugging info to the version-generation script via stderr.

Closes #1366

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -96,6 +96,7 @@ jobs:
         with:
           ref: ${{ inputs.ref }}
           submodules: recursive
+          fetch-tags: true
 
       - name: Install Linux deps
         if: runner.os == 'Linux'
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -47,6 +47,7 @@ jobs:
         with:
           ref: ${{ inputs.tag }}
           submodules: recursive
+          fetch-tags: true
 
       - name: Log in to Docker Hub
         uses: docker/login-action@v3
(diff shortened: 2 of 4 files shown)


joeyparrish:
See the fix at work here:

```sh
$ docker run joeyparrish/shaka-packager-staging:v3.0.4 packager --version
packager version v3.0.4-f52052a-debug
```

cosmin (approved):

PR #1371, opened 2024-03-12
fix: Fix NPM binary publication

### LINKED ISSUE #1369: NPM packages broken before v3.0.3
All NPM releases before v3.0.3 (upcoming) are missing their binaries.  The prepublish script downloads the binaries, but the default .npmignore file prevents those from being picked up in the archive.  Because of this, we have never made a correct NPM release ever before.

For history, NPM packages were originally requested by @nampdn in 2018 in #290.

There are currently 0 projects referencing the `shaka-packager` packages through npmjs, but there are ~100 downloads per week.  Those could all be bots mirroring NPM, though.  Hard to say.

I will mark all existing releases in NPM as "deprecated", but I won't unpublish them.  This should be fixed in v3.0.3.

joeyparrish:
The default .npmignore would disallow bin/, which contains binaries for our NPM release.  Every release before v3.0.3 was missing .npmignore, and therefore had no binaries.  These were unusable in NPM, and have been marked as deprecated because of it.

Closes #1369

### DIFF
--- a/.github/workflows/publish-npm.yaml
+++ b/.github/workflows/publish-npm.yaml
@@ -44,7 +44,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ inputs.tag }}
-          submodules: recursive
 
       - uses: actions/setup-node@v4
         with:
--- a/npm/.npmignore
+++ b/npm/.npmignore
@@ -1,0 +1,8 @@
+# This file is required for us, even if empty.
+#
+# The default .npmignore would disallow bin/, which contains binaries for our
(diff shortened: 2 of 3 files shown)


joeyparrish:
I verified this manually through `npm pack`.  Final verification will be through the job https://github.com/joeyparrish/shaka-packager/actions/runs/8254276102, which will publish to https://www.npmjs.com/package/@joeyparrish/shaka-packager-staging

joeyparrish:
Now verified through the full publication process.

cosmin (approved):

PR #1376, opened 2024-03-20
fix: Fix NPM binary selection on ARM Macs

### LINKED ISSUE #1375: Unable to run executable via NPM module on Apple Silicon (even with Rosetta 2 enabled)
## System info
**Operating System**: macOS Sonoma 14.3.1 (23D60)
**Shaka Packager Version**: v3.0.3

(1) Find a Mac computer running Apple Silicon chips.
(2) On this Mac create a Node.js project with `index.js` file, install the npm package `shaka-packager`, then try to call installed module in that `index.js` file: `require('shaka-packager')`.

What is the expected result?
Macs on M chips have Rosetta 2 technology, which makes it possible to seamlessly run applications with x64 architecture. I would like to be able to run the executable on Apple Silicon: either allow the mac x64 build to run, or add a mac arm64 variant to the builds.

What happens instead?
An error occurs when I call a module from Node.js script.

```
/Users/user/project/node_modules/shaka-packager/index.js:28
  throw new Error(
  ^

Error: Architecture not supported: darwin/arm64
    at Object.<anonymous> (/Users/user/project/node_modules/shaka-packager/index.js:28:9)
```

vitalygashkov:
Fixes #1375

### DIFF
--- a/npm/index.js
+++ b/npm/index.js
@@ -13,6 +13,7 @@ var commandNames = {
   },
   darwin: {
     'x64': 'packager-osx-x64',
+    'arm64': 'packager-osx-arm64',
   },
   win32: {
     'x64': 'packager-win-x64.exe',


joeyparrish (approved):

joeyparrish:
Thank you for the fix!  When the tests pass, I will merge it.  Looks like I will need to rerun one job that got canceled somehow.

PR #1395, opened 2024-05-07
fix: escape media URLs in MPD

### LINKED ISSUE #1107: Packaging to DASH does not work when output filename contains an ampersand (unterminated entity reference)
## System info
**Operating System**: macOS Monterey 12.6
**Shaka Packager Version**: v2.6.1-634af65-release

When packaging to DASH and the output filename contains an ampersand (&), an `unterminated entity reference` error is printed.
Packaging completes, but the resulting MPD manifest is invalid.

**Packager Command**:
Working:
`packager-osx "input=SampleVideo.mp4,output=Sample_Video.mp4,output_format=mp4,stream_selector=video" -mpd_output manifest_ok.mpd`

Not working:
`packager-osx "input=SampleVideo.mp4,output=Sample&Video.mp4,output_format=mp4,stream_selector=video" -mpd_output manifest_nok.mpd`

```
[1020/191510:INFO:demuxer.cc(89)] Demuxer::Run() on file 'SampleVideo.mp4'.
[1020/191510:INFO:demuxer.cc(155)] Initialize Demuxer for file 'SampleVideo.mp4'.
[1020/191510:INFO:single_segment_segmenter.cc(111)] Update media header (moov) and rewrite the file to 'Sample&Video.mp4'.
error : unterminated entity reference       Video.mp4
[1020/191511:INFO:mp4_muxer.cc(186)] MP4 file 'Sample&Video.mp4' finalized.
error : unterminated entity reference       Video.mp4
Packaging completed successfully.
```

Diff of the manifests:
```diff
diff manifest_ok.mpd manifest_nok.mpd
7c7
<         <BaseURL>Sample_Video.mp4</BaseURL>
---
>         <BaseURL/>
```

(1) Package any video to DASH and set the output video filename to something containing an ampersand (`&`)
(2) Check the manifest

What is the expected result?
A manifest with filename filled out at `BaseURL`.

What happens instead?
The resulting manifest has an empty `BaseURL`

Additionally, if you run the following command:
`packager-osx "input=SampleVideo.mp4,output=Sample&amp;Video.mp4,output_format=mp4,stream_selector=video" -mpd_output manifest_nok2.mpd`

The BaseURL looks like....

```xml
<BaseURL>Sample&amp;Video.mp4</BaseURL>
```

But the resulting file has a filename `Sample&amp;Video.mp4`

cosmin:
Currently `media_info.media_file_url()` is not escaped when placed into MPD for things like BaseURL. This for example breaks when trying to us a file name that contains special characters like &. Since these are supposed to be URLs let's URL encode them.

Fixes #1107

### DIFF
--- a/packager/mpd/base/xml/xml_node.cc
+++ b/packager/mpd/base/xml/xml_node.cc
@@ -18,6 +18,7 @@
 #include <absl/strings/escaping.h>
 #include <absl/strings/numbers.h>
 #include <absl/strings/str_format.h>
+#include <curl/curl.h>
 #include <libxml/tree.h>
 
 #include <packager/macros/compiler.h>
@@ -52,6 +53,22 @@ const char kAC4Codec[] = "ac-4";
 const char kDTSCCodec[] = "dtsc";
 const char kDTSECodec[] = "dtse";
 const char kDTSXCodec[] = "dtsx";
+
+std::string urlEncode(const std::string& input) {
+  CURL* curl = curl_easy_init();  // Initialize a CURL session
+  if (curl) {
+    char* output = curl_easy_escape(curl, input.c_str(), input.length());
+    if (output) {
+      std::string encodedUrl(output);
+      curl_free(output);        // Free the output string when done
+      curl_easy_cleanup(curl);  // Clean up the CURL session
+      return encodedUrl;
+    }
+    curl_easy_cleanup(
+        curl);  // Clean up the CURL session even if encoding fails
+  }
+  return "";  // Return empty string if initialization fails
+}
 
 std::string RangeToString(const Range& range) {
   return absl::StrFormat("%u-%u", range.begin(), range.end());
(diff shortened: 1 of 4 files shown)


packager/mpd/base/xml/xml_node.cc
```diff
@@ -53,6 +54,22 @@ const char kDTSCCodec[] = "dtsc";
 const char kDTSECodec[] = "dtse";
 const char kDTSXCodec[] = "dtsx";
 
+std::string urlEncode(const std::string& input) {
+  CURL* curl = curl_easy_init();  // Initialize a CURL session
+  if (curl) {
+    char* output = curl_easy_escape(curl, input.c_str(), input.length());
```
joeyparrish, line 60: According to the `curl_easy_escape` docs:

> Since 7.82.0, the curl parameter is ignored. Prior to that there was per-handle character conversion support for some old operating systems such as TPF, but it was otherwise ignored.

So I think it's best if we don't allocate and free the `CURL*` handle, just to cut down on memory churn and risk of leaking if an exception is thrown.

joeyparrish (changes requested):

packager/mpd/base/xml/xml_node.cc
joeyparrish: Done.  PTAL

joeyparrish (approved):

unext-wendong:
Was it intentional to use URL encode instead of XML encode?

Actually for XML/HTML, only the following 4 characters need to be escaped
```
> < & %
```

URL encoding does much more than this. Characters like forward slashes are also encoded.

PR #1398, opened 2024-05-09
fix: set yuv full range flag to 1 for VP9 with sRGB

### LINKED ISSUE #990: Incorrect implementation of color_range in VP9?
I was looking through the specs of VP9 (https://storage.googleapis.com/downloads.webmproject.org/docs/vp9/vp9-bitstream-specification-v0.6-20160331-draft.pdf) and the implementation of VP9 parser in shaka-packager. I think I found a bug in the parsing of color_range.

In particular, if color_space is VPX_COLOR_SPACE_SRGB, the specs says that color_range should be 1 i.e. yuv_full_range = true. However, yuv_full_range is default initialized as false and isn't set in the branch for color_space is VPX_COLOR_SPACE_SRGB.

cosmin:
If color_space is VPX_COLOR_SPACE_SRGB, the specs says that color_range should be 1 i.e. yuv_full_range = true. However, yuv_full_range is default initialized as false and isn't set in the branch for color_space is VPX_COLOR_SPACE_SRGB.

Fixes #990

### DIFF
--- a/packager/media/codecs/vp9_parser.cc
+++ b/packager/media/codecs/vp9_parser.cc
@@ -285,7 +285,7 @@ bool ReadBitDepthAndColorSpace(BitReader* reader,
     }
   } else {
     // Assume 4:4:4 for colorspace SRGB.
-    yuv_full_range = 1;
+    yuv_full_range = true;
     chroma_subsampling = VPCodecConfigurationRecord::CHROMA_444;
     if (codec_config->profile() & 1) {
       bool reserved;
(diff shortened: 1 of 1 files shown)


joeyparrish (approved):

packager/media/codecs/vp9_parser.cc
```diff
@@ -285,6 +285,7 @@ bool ReadBitDepthAndColorSpace(BitReader* reader,
     }
   } else {
     // Assume 4:4:4 for colorspace SRGB.
+    yuv_full_range = 1;
```
joeyparrish, line 288: nit: it's a bool, so `yuv_full_range = true`

joeyparrish (changes requested):

packager/media/codecs/vp9_parser.cc
joeyparrish: Done

joeyparrish (approved):

Issue #1415, opened 2024-08-02
Immersive Audio Model and Formats (IAMF) support


felicialim:
IAMF is a codec-agnostic container specification for immersive spatial audio developed by the Alliance for Open Media (AOMedia). 

The working draft for IAMF v1.1 is at https://aomediacodec.github.io/iamf.

I will create a PR to add support for IAMF.

joeyparrish:
Fixed in #1416

PR #1416, opened 2024-08-02
feat: Add IAMF support (#1415)

felicialim:


sr1990:
Please add an end-to-end test in src/packager/app/test/packager_test.py to create golden files for verifying the correctness of the generated codec strings in DASH and HLS manifest files.

packager/media/formats/mp4/mp4_muxer.cc
```diff
@@ -576,11 +581,20 @@ bool MP4Muxer::GenerateAudioTrak(const AudioStreamInfo* audio_info,
     audio.channelcount = audio_info->num_channels();
     //ETSI TS 103 190-2, E.4.6 samplesize shall be set to 16.
     audio.samplesize = 16;
+  } else if (audio_info->codec() == kCodecIAMF) {
+    // IAMF sets channelcount to 0
+    // https://aomediacodec.github.io/iamf/#iasampleentry-section
+    audio.channelcount = 0;
```
sr1990, line 587: I am not very familiar with IAMF, but after reviewing the spec, I have a few questions:

1. Based on [this section](https://aomediacodec.github.io/iamf/#brands), it seems that the iamf compatible brand is not present in ftyp box. Should this be added after line 265 above?
2. I noticed that not all points mentioned in [this section](https://aomediacodec.github.io/iamf/#isobmff-singletrack-basicencapsulationscheme) are included in the PR. Is there a specific reason for this?
3. The [documentation on common encryption](https://aomediacodec.github.io/iamf/#commonencryption) mentions details about encrypting the IAMF stream. Are only AudioFrameOBUs encrypted?

felicialim: Thank you for reviewing! 

1. Good catch, done.
2. I'm probably just misunderstanding what needs to be implemented here - I thought that these are expected already be in the input mp4 files and would be copied over when encrypting, or is this something that needs to be added separately here? 
3. Parameter Block OBUs would also be encrypted along with AudioFrameOBUs - both of them are part of a [Temporal Unit](https://aomediacodec.github.io/iamf/#temporal-unit), which is in turn packaged as one temporal unit per [IA Sample](https://aomediacodec.github.io/iamf/#iasampleformat).

felicialim:
> Please add an end-to-end test in src/packager/app/test/packager_test.py to create golden files for verifying the correctness of the generated codec strings in DASH and HLS manifest files.

Thanks for the pointer, done.

packager/media/formats/mp4/mp4_muxer.cc
sr1990: 2.Please ignore. Sorry about that.
Do you think it would be beneficial to add checks for the constraints mentioned in the section to ensure the provided input is correct?

3.Thanks for the explanation.
I have a few questions:
a. For each sample indicated by `trun`, does it contain one TU (comprising OBU header type 5, Audio Frame OBU, OBU header type 3, and OBU parameter block)?

b. In the encryption handler, during media sample processing at [EncryptionHandler::ProcessMediaSample](https://github.com/shaka-project/shaka-packager/blob/9cb00e91ab0d081b1f2132abaad7c23ef15047b7/packager/media/crypto/encryption_handler.cc#L265), media subsamples are generated in [SubsampleGenerator::GenerateSubsamples](https://github.com/shaka-project/shaka-packager/blob/9cb00e91ab0d081b1f2132abaad7c23ef15047b7/packager/media/crypto/subsample_generator.cc#L213). 
Is the entire sample encrypted, meaning no subsamples are expected as per [here](https://github.com/shaka-project/shaka-packager/blob/9cb00e91ab0d081b1f2132abaad7c23ef15047b7/packager/media/crypto/subsample_generator.cc#L242)? 
Or are the OBU headers skipped, requiring the exact addresses and sizes of the Audio Frame OBU and Parameter Block OBU to be extracted as subsamples, similar to how AV1 tiles are extracted?

c. Suggestion: It might be helpful to include a diagram similar to [AV1](https://aomediacodec.github.io/av1-isobmff/#subsample-encryption-illustration) in the IAMF documentation

cosmin:
@sr1990 does this seem good enough to merge for now and potentially refine later or do we have anything outstanding that must be addressed first?

packager/media/formats/mp4/mp4_muxer.cc
felicialim: Sorry I missed this before.

2. Agreed checking could be useful, and if these constraints are not met, we could try to fix or at least warn the user. To implement all checks we'd want to make the OBU parsing more detailed to extract relevant info. Do you think this could be a separate PR?

3a. Right, one `trun` sample is one TU with those OBUs. In one TU, there could be more than one of each Audio Frame OBU and Parameter Block OBU (https://aomediacodec.github.io/iamf/#temporal-unit). E.g., the number of Audio Frame OBUs depend on the number of Audio Substreams for each Audio Element, and the total number of Audio Elements.

3b,c. IAMF uses full sample or whole-block full sample encryption (https://aomediacodec.github.io/iamf/#commonencryption), including the OBU headers, so I believe the code path you indicated is right, we don't expect subsamples. Given that, we weren't sure that a diagram similar to AV1 would be as useful, but it sounds like we could look at some clarification in [6.3. Common Encryption](https://aomediacodec.github.io/iamf/#commonencryption).

sr1990: @felicialim 
2. Sounds good.
3. Thanks for the explanation.

sr1990:
> @sr1990 does this seem good enough to merge for now and potentially refine later or do we have anything outstanding that must be addressed first?

@cosmin  lgtm. Thanks @felicialim.

cosmin (approved):

joeyparrish (approved):

PR #1423, opened 2024-08-24
ci: Build Docker for `arm64` arch

mathix420:
- Trivial upgrades:
  - docker/build-push-action@v5 => v6
- Added actions:
  - docker/setup-buildx-action@v3
- Added `workflow_dispatch` in `build-docker.yaml` for better DX

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -14,6 +14,12 @@ on:
       ref:
         required: true
         type: string
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: The ref to build from.
+        required: true
+        type: string
 
 # By default, run all commands in a bash shell.  On Windows, the default would
 # otherwise be powershell.
@@ -32,6 +38,11 @@ jobs:
           ref: ${{ inputs.ref }}
           submodules: recursive
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          platforms: linux/amd64,linux/arm64
+
       - name: Build
         shell: bash
-        run: docker buildx build .
+        run: docker buildx build --platform linux/amd64,linux/arm64 .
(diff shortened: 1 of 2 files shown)


cosmin (approved):

cosmin:
@joeyparrish do we need a CLA for a contribution that only touches the github workflows and not the actual code?

.github/workflows/build-docker.yaml
```diff
@@ -32,6 +38,11 @@ jobs:
           ref: ${{ inputs.ref }}
           submodules: recursive
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          platforms: linux/amd64,linux/arm64
```
joeyparrish, line 44: Does this use qemu for emulation?  How slow is it?

mathix420: Yes it does, and it's quite slow, took 1h on my gh action. Best would be to setup a matrix step on an external runner like you did on other actions but I can't setup this myself.

mathix420:
@joeyparrish I believe this might be accessible by a google project https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/

.github/workflows/build-docker.yaml
joeyparrish: I don't think I want to add 1h of emulated Docker builds to every PR.  Let me try to set it up through our self-hosted runners, at least until GitHub starts offering Arm Linux runners publicly.

joeyparrish (changes requested):

joeyparrish:
I'm closing this in favor of native arm64 builds, which I'm currently testing on my fork using GitHub's arm64 Linux VMs.

joeyparrish:
Just got #1490 working on my fork!

Issue #36, opened 2015-08-06
Taking Live Audio From LINE Input


jeprojects:
I have gone through the documentation. Is it possible for the library to take the audio from LINE in on the sound card and process it to several different bitrate audio and dash?

kqyang:
Not directly. But you can use vlc to capture and transcode it into different bitrate audio. edash-packager can then process the output from vlc and generates dash streams.

PR #1427, opened 2024-08-28
feat: EXT-X-SESSION-KEY support (#36)

SteveR-PMP:
* adding EXT-X-SESSION-KEY to master_playlist.cc

* adding 'create_session_keys' command-line option

* adding unit test

### DIFF
--- a/packager/hls/base/media_playlist.cc
+++ b/packager/hls/base/media_playlist.cc
@@ -181,7 +181,7 @@ class SegmentInfoEntry : public HlsEntry {
                    uint64_t segment_file_size,
                    uint64_t previous_segment_end_offset);
 
-  std::string ToString() override;
+  std::string ToString(std::string) override;
   int64_t start_time() const { return start_time_; }
   double duration_seconds() const { return duration_seconds_; }
   void set_duration_seconds(double duration_seconds) {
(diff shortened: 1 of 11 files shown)


packager/hls/base/media_playlist.cc
```diff
@@ -181,7 +181,7 @@ class SegmentInfoEntry : public HlsEntry {
                    uint64_t segment_file_size,
                    uint64_t previous_segment_end_offset);
 
-  std::string ToString() override;
+  std::string ToString(std::string) override;
```
cosmin, line 184: Having to modify the signature of `ToString()` all over the place just to be able to do `entry->ToString("#EXT-X-SESSION-KEY")` doesn't feel right.

Can probably either add a method on `EncryptionInfoEntry` or a subclass, etc.

cosmin (changes requested):
Other than ToString() everything else looks good to me

packager/hls/base/media_playlist.cc
SteveR-PMP: @cosmin
I changed this to add a ToString(std::string) to `EncryptionInfoEntry` only and then cast entry to type EncryptionInfoEntry at L580-L581.

cosmin (approved):

PR #1442, opened 2024-10-24
feat(http): Add DELETE method support

joeyparrish:
When we try to remove an old segment from a live stream we uploaded via HTTP, we need to send DELETE requests.

### DIFF
--- a/docs/source/tutorials/http_upload.rst
+++ b/docs/source/tutorials/http_upload.rst
@@ -101,14 +101,6 @@ Backlog
 Please note the HTTP upload feature still lacks some features
 probably important for production. Contributions are welcome!
 
-HTTP DELETE
-===========
-Nothing has be done to support this yet:
-
-    Packager supports removing old segments automatically.
-    See ``preserved_segments_outside_live_window`` option in
-    DASH_ options or HLS_ options for details.
-
 Software tests
 ==============
 We should do some minimal QA, check whether the test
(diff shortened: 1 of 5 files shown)


joeyparrish:
I'm still in the process of verifying this change.  Converting to draft.

cosmin (approved):
lgtm. should we add an option to enable/disable this behavior? If DELETE was never implemented then someone might be relying on the existing behavior, and it might be good to have an option to revert to this behavior at least temporarily.

joeyparrish:
I am not sure there's a storage API out there that doesn't have both PUT and DELETE.  And without DELETE, your cloud storage fills up and you have to clean it out periodically with some other process.  My guess is that HTTP output support was half-implemented and never seriously used.  I think this PR finishes the feature for the first time.

joeyparrish:
I suppose `preserved_segments_outside_live_window` could be configured in a way that would never attempt to delete files.  If anyone is using it that way, then this isn't going to impact them anyway.

joeyparrish:
Verified on my live stream!

I'm going to leave this open for comment for another day or so, in case anyone thinks I'm being unreasonable about not putting this change behind a flag.

cosmin:
I agree that `preserved_segments_outside_live_window` seems like a reasonable workaround for anyone that doesn't want to delete.

joeyparrish:
It's been nearly 24 hours without comment, so I'm going to merge this.  If this ruins somebody's day, we can add a flag later.  I suspect it will be fine.

PR #1459, opened 2024-12-11
ci: Ignore cert error for apt.kitware.com

joeyparrish:
This should fix the retrieval of the GPG key and allow an updated cmake to be installed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -105,16 +105,16 @@ jobs:
         # that may be greater than what is available in Ubuntu, so we set up
         # the official CMake PPA first.
         run: |
-          kitware_key_url="https://apt.kitware.com/keys/kitware-archive-latest.asc"
+          kitware_key_url="http://apt.kitware.com/keys/kitware-archive-latest.asc"
           kitware_key_path="/usr/share/keyrings/kitware-archive-keyring.gpg"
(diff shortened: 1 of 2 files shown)


cosmin (approved):

cosmin:
Need to also fix https://github.com/shaka-project/shaka-packager/blob/a26e6623d694fa4ca592e0b4594a01e4c23aa291/packager/testing/dockers/Ubuntu_Dockerfile#L18 I believe

cosmin:
This is probably good to merge since it fixed the cert error, but the build is still failing, now with a CMake version being out of date.

> CMake Error at CMakeLists.txt:10 (cmake_minimum_required): CMake 3.24 or higher is required.  You are running version 3.22.1
  
  can probably fix that in a separate PR?

joeyparrish:
> > CMake Error at CMakeLists.txt:10 (cmake_minimum_required): CMake 3.24 or higher is required.  You are running version 3.22.1

This is because of a failure of apt to read from the third-party repo with up-to-date packages.  Essentially, I didn't fix the Dockerfile.

Moving to plain HTTP works, though.  I think the only reason the workflow worked with only the curl change was through some sort of caching in GitHub Actions, so apt never had to access the server with the bad cert in that context.

We'll see if I'm right, if this next test pass is all green.

PR #1462, opened 2024-12-16
ci: Update workflow permissions

joeyparrish:
Now that default permissions are read-only, we must enable specific permissions for certain workflow jobs.


joeyparrish:
Permission changes verified on my fork.  This was the only workflow failing after our change to default token permissions on December 3.

JulianDomingo (approved):

Issue #1483, opened 2025-02-25
MV-HEVC support


dannyshong:
Stereo video coded in MV-HEVC is becoming more widely available and the packager should support this.

dannyshong:
I'll be adding this support

PR #1488, opened 2025-03-10
MV-HEVC support for stereo video

dannyshong:
Adding MV-HEVC support limited to stereo video to resolve  Issue #1483.  Stereo video coded in MV-HEVC is becoming more widely available: Apple Vision Pro supports stereo video playback in MV-HEVC and both the headset and iPhone supports capturing stereo video using the format.  FFmpeg has also added support for MV-HEVC.


cosmin:
Do you happen to have a sample file so we can add an integration test as well?

dannyshong:
Yes, I share a sample MV-HEVC .mp4 for stereo video.  Where should I place the binary?

cosmin:
In the `packager/app/test/testdata` we have folders with one per integration test. The tests are defined in `packager/app/test/packager_test.py` and should be fairly straightforward to copy one of the existing tests with the new asset and then check in the golden file outputs as well.

cosmin:
Two tests are failing on Windows, likely due to MSVC having some differences.

```
5: [ RUN      ] H265VideoSliceHeaderParserTest.BasicStereoVideoSupport
5: D:\a\shaka-packager\shaka-packager\packager\media\codecs\video_slice_header_parser_unittest.cc(278): error: Expected equality of these values:
5:   22
5:   parser.GetHeaderSize(nalu)
5:     Which is: 8
5: [  FAILED  ] H265VideoSliceHeaderParserTest.BasicStereoVideoSupport (0 ms)
```

and

```
5: [ RUN      ] H265ParserTest.ParseStereoVideoSliceHeaderInter
5: D:\a\shaka-packager\shaka-packager\packager\media\codecs\h265_parser_unittest.cc(477): error: Expected equality of these values:
5:   176u
5:     Which is: 176
5:   header1.header_bit_size
5:     Which is: 64
5: [  FAILED  ] H265ParserTest.ParseStereoVideoSliceHeaderInter (0 ms)5: [ RUN      ] H265ParserTest.ParseStereoVideoSliceHeaderInter
```

dannyshong:
Thanks!  Made the fix.

cosmin:
the newly added integration test looks good but seems to be missing the source and golden files in the `'mv-hevc-mp4` and `'mv-hevc-mp4-with-encryption` folders.

dannyshong:
> the newly added integration test looks good but seems to be missing the source and golden files in the `mv-hevc-mp4` and `mv-hevc-mp4-with-encryption` folders.

I thought they were added.  Sorry.  Added them.

sr1990:
Hey, I have a few questions: 
1. For HLS:
a. The [Apple HLS Authoring Specification](https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Stereo-Video) mentions the REQ-VIDEO-LAYOUT tag. Specifically, section 16.5 states:
"If stereo video content includes sections of monoscopic video content, then the REQ-VIDEO-LAYOUT attribute MUST include 'CH-MONO' as one entry."

In the test available, this tag is not present.
Should the value for this tag be parsed from the MP4 boxes, or is it expected that users provide it explicitly? Also, what are the possible values? 

 b. According to [RFC 8216bis-17, section 8](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-17#section-8), a playlist MUST indicate an EXT-X-VERSION of 12 or higher if it contains any attribute whose name starts with "REQ-".
Currently, EXT-X-VERSION is set to 6.
Should the version be updated in the case of MV-HEVC video to ensure compliance?

 c. [Apple HLS Authoring Specification, section 16.4](https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Stereo-Video) states that switches between stereo and monoscopic video must be marked with an EXT-X-DISCONTINUITY tag, or the Media Initialization Section must contain sample descriptions for both types, with each sample’s index field pointing to the appropriate sample description.
Is this requirement currently handled?
If the content is restricted to only stereoscopic video, should this be enforced while parsing the command line arguments?

 d. Additional considerations related to trickplay and audio streams are mentioned in the [Apple Trick-play Documentation](https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Trick-play).
If trickplay, audio, or other streams (in addition to MV-HEVC video) are restricted, should this be checked during command-line argument parsing?

2. For DASH:
Is DASH supported? and if so, what is the equivalent of the REQ-VIDEO-LAYOUT attribute?
If DASH does not support this feature, should the parameters be checked to confirm that the content is HLS-only?

dannyshong:
Thank you for raising the issues related to HLS and DASH.  I totally overlooked them as my focus was on encrypting MV-HEVC mp4 files.  Is it a requirement to have HLS and DASH support in order to submit the changes?  If not, we can make the integration test only focus on encryption and decryption of MV-HEVC mp4.

cosmin:
Lint is failing now on the test

packager/app/test/packager_test.py:1599:0: C0301: Line too long (81/80) (line-too-long)

dannyshong:
Anything else needed to get the approval?

cosmin (approved):

dannyshong:
Thank you!  I don't think I have push access so I guess I'll wait for someone to merge?

joeyparrish:
When you say:

> Proper HLS and DASH support will still need to be added.

What is involved in that?

Also, when the PR says:

> Fixes https://github.com/shaka-project/shaka-packager/issues/1483

Is that true if we don't have HLS & DASH support?  Should we reference the issue, but leave it open for that follow-up HLS & DASH work?

dannyshong:
The changes needed for HLS and DASH are mostly written down by [sr1990's comment above](https://github.com/shaka-project/shaka-packager/pull/1488#issuecomment-2716568081).  This is mostly making sure that correct tags and corresponding data are in place for MV-HEVC - e.g., REQ-VIDEO-LAYOUT with CH-STEREO for HLS.

Got rid of Fixes #1483

cosmin:
@dannyshong are you planning to work on the DASH and/or HLS support as well?

dannyshong:
I'm not an expert in HLS or DASH, so I would prefer not to if I don't have to.  But, if I need to add the support, then I can come back to it in about 2 months if that's ok?

joeyparrish:
Released in v3.5.0.  Thanks!

PR #1490, opened 2025-03-15
feat(ci): Official multiarch docker images for x64 + arm64

joeyparrish:
This adds support for Docker-based testing on arm64 images and runners.

This also adds a multiarch build process so that future releases cover both x64 and arm64.

This is tagged as "feat" so that this gets called out in the changelog for the next release.


cosmin (approved):

joeyparrish:
Released in v3.5.0.  Thanks!

PR #1505, opened 2025-08-08
ci: Use a mirror of musl.cc

joeyparrish:
musl.cc is overwhelmed by GitHub traffic, and therefore has blocked all of Microsoft.  This switches to a mirror run on GitHub so that we can continue to access the toolchains.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -190,7 +190,7 @@ jobs:
             else
               MUSL_ARCH="x86_64"
             fi
-            curl -LO https://musl.cc/"$MUSL_ARCH"-linux-musl-native.tgz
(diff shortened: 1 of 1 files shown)


cosmin:
This change looks good although we seem to have some other issues with old CMake versions < 3.5 on some builds

joeyparrish:
I'll try to work on that next.

cosmin:
I took a pass at trying to upgrade all dependencies a while ago to get things into a better state, and get things to build on my Mac with latest toolchain, but ran into some problems around libpng + zlib, there were some changes in latest version, and then I ran out of time.

joeyparrish:
Yeah, I have a mac laptop now, and I just ran into the missing fp.h that libpng wants unconditionally on all macs.

PR #1507, opened 2025-08-09
fix: Upgrade libpng to fix build on new macs

joeyparrish:
This newer libpng doesn't assume the existence of fp.h on mac (which isn't present on newer ones), but the CMake options for libpng changed and had to be adjusted somewhat.


joeyparrish:
Ah, the zlib path in the new format needs adjusting for Windows.  I was sending this on my way out the door for a short vacation... I'll follow up next week when I'm back.

cosmin (approved):

joeyparrish:
Released in v3.5.0.  Thanks!

PR #1508, opened 2025-08-11
fix: Upgrade deps that are meant for older versions of CMake

cosmin:
Current deps require an old CMake version, support for which has been removed from latest CMake in Ubuntu and Arch.

```
CMake Error at packager/third_party/c-ares/source/CMakeLists.txt:3 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 3.5 has been removed from CMake.
```

This upgrades:
 - c-ares, v1.20.1 => v1.34.5
 - json, v3.10.1-115-g954b10ad => v3.12.0
 - libwebm, 1.0.0.28 => 1.0.0.32

With these changes, the build has been verified with CMake 3.31.11 (latest v3) and CMake 4.2.3 (latest v4).


cosmin:
json library needs to be upgraded as well for the same reason.

cosmin:
I'll rebase this after we land the libpng changes

joeyparrish (approved):

niklaskorz:
FWIW, @cosmin, we're devendoring all dependencies in nixpkgs (for both security and build system reasons), and that is working fine. If that's of interest for Arch too, you can have a look at our patch here: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/sh/shaka-packager/0002-Unvendor-dependencies.patch

Edit: looks like I have misinterpreted and this did not occur in the context of packaging for these distros, but just general development on them :)

niklaskorz:
Also, contrary to the PR description, this PR does not update abseil-cpp and shaka-packager is currently still incompatible to newer abseil-cpp versions.

Edit: the only hard incompatibility is a missing include of `<absl/log/vlog_is_on.h>` in `packager/file/http_file.cc`, everything else are deprecation warnings

joeyparrish:
Apologies!  I updated abseil-cpp, then had to revert it due to incompatibilities, then forgot to update the PR description.  I edited it for accuracy just now.

abseil-cpp can definitely be upgraded, and PRs are welcome that bring Packager up-to-date with the latest abseil APIs.

PR #1519, opened 2025-10-08
feat: --local_targetduration

### LINKED ISSUE #1510: EXT-X-TARGETDURATION causing problems with track swith buffering
Question
In Shaka Packager we have this issue from a while back which is sort of related to problem we are having.
https://github.com/shaka-project/shaka-packager/issues/947

We ran into something similar recently and opened an issue with Apple Team.

Basically when we package a CMAF HLS manifest which is fragmented mp4 file with byte range requests with shaka packager it has EXT-X-TARGETDURATION : <max of segment size across all variants> across all variant playlists video/audio/subtitles

This leads to unexpected buffer when someone switches across the tracks of a longform content.

Reached out to Apple and they helped clarify that the EXT-X-TARGETDURATION is to be the local max segment size within a playlist and not across playlist. e.g all your video / audio are say 6 second segments but your subtitles are single VTT file = duration of longform content

https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.3.1

So is that just a bug in shaka packager ? Happy to work on it and help fix it just need some confirmation if we think that the APPLE HLS spec is clear and that its a bug .

NOTES from that open issue :
What is the expected result?
Chunk duration should be honored, and EXT-X-TARGETDURATION should refer to the durations in the playlist (not in packager's runtime history).

What happens instead?
Chunk duration is not honored, and EXT-X-TARGETDURATION always keeps it's maximum historical value (in the context of the current runtime), disregarding the actual playlist content.

SteveR-PMP:
change EXT-X-TARGETDURATION to have the value set to the highest duration for each variant instead of the highest duration for all the variants

Fixes #1510


joeyparrish (changes requested):

SteveR-PMP:
Hi Joey, I'll make these changes and update the PR.

vish91:
Hey @joeyparrish  @cosmin  if you can review this one hoping to get it merged for next release.

joeyparrish:
Reviewing now.  Sorry for dropping the ball!

joeyparrish (approved):

joeyparrish:
Apologies for the delay! This will be in the next release, which will be released as soon as I finish going through pending PRs.

joeyparrish:
Released in v3.5.0.  Thanks!

PR #1525, opened 2025-11-24
feat: Switch Live to VOD stream when event/live streams end

ronak2121:
Implement the MPEG DASH/HLS specifications to switch event/live streams to VOD when the streams end.
    
This change implements the DASH standard by:
    
1. Change MPD@type to "static"
2. Set MPD@mediaPresentationDuration to the media duration
3. Removing "dynamic" specific attributes in element, e.g. availabilityStartTime, minimumUpdatePeriod, timeShiftBufferDepth etc.
4. Set SegmentTemplate@presentationTimeOffset in Representations
    
This change implements the HLS standard by:
    
1. Switch the PLAYLIST-TYPE to VOD at the conclusion of a LIVE stream
2. Append an EXT-X-ENDLIST tag at the end of the stream

### DIFF
--- a/packager/hls/base/media_playlist.cc
+++ b/packager/hls/base/media_playlist.cc
@@ -510,20 +510,25 @@ void MediaPlaylist::AddPlacementOpportunity() {
   entries_.emplace_back(new PlacementOpportunityEntry());
 }
 
-bool MediaPlaylist::WriteToFile(const std::filesystem::path& file_path) {
+bool MediaPlaylist::WriteToFile(const std::filesystem::path& file_path, const bool endStream) {
   if (!target_duration_set_) {
     SetTargetDuration(ceil(GetLongestSegmentDuration()));
   }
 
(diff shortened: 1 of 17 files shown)


packager/hls/base/media_playlist.cc
```diff
@@ -510,20 +510,25 @@ void MediaPlaylist::AddPlacementOpportunity() {
...
-      media_info_, target_duration_, hls_params_.playlist_type, stream_type_,
+      media_info_, target_duration_, playlist_type, stream_type_,
       media_sequence_number_, discontinuity_sequence_number_,
       hls_params_.start_time_offset);
 
   for (const auto& entry : entries_)
     absl::StrAppendFormat(&content, "%s\n", entry->ToString().c_str());
 
-  if (hls_params_.playlist_type == HlsPlaylistType::kVod) {
+  if (endStream) {
```
xavierlaffargue: You can change to playlist_type here

PR #1526, opened 2025-12-01
fix: handle EEXIST error in directory creation

rlaisqls:
Fix incorrect error handling in `LocalFile::Open()` when `create_directories()` encounters an already-existing directory.

  ### Problem

  The current code treats `create_directories()` returning `false` as a failure:

  ```cpp
  if (!std::filesystem::create_directories(parent_path, ec)) {
    return false;  // BUG: ec is not checked
  }
  ```

  However, according to the C++ standard, [std::filesystem::create_directories()](https://en.cppreference.com/w/cpp/filesystem/create_directory.html):
  - Returns true if a new directory was created
  - Returns false with ec.clear() if the directory already exists (not an error)
  - Returns false with ec set if an actual error occurred

  This causes failures in several scenarios:
  1. Concurrent packaging: Multiple threads creating the same output directory
  2. External directory creation: Another process creates the directory between is_directory() check and create_directories()
  call
  3. Retry after partial failure: Directory was created in a previous failed attempt

  ### Example Failure (strace)

  ```bash
  newfstatat("/tmp/out/dir") = -1 ENOENT      ← Directory doesn't exist
  mkdirat("/tmp/out/dir", 0777) = -1 EEXIST   ← Created by another process/thread
  exit(0)                                      ← Packager exits without writing output
  ```

  The directory exists and is ready to use, but the code incorrectly treats this as a failure.

###  Solution

  Check the error_code after create_directories() returns false. Only return failure if ec indicates an actual error.

  ```cpp
  if (!std::filesystem::create_directories(parent_path, ec)) {
    if (ec) {  // Only fail if there's an actual error
      return false;
    }
    // ec is empty means directory already exists - continue normally
  }
  ```

### DIFF
--- a/packager/file/file_unittest.cc
+++ b/packager/file/file_unittest.cc
@@ -313,7 +313,8 @@ TEST_F(LocalFileTest, WriteInDirectory) {
   EXPECT_EQ(data_, read_data);
 
   // Wipe the folder.
-  std::filesystem::remove_all(std::filesystem::u8path(temp_folder));
+  std::error_code ec;
+  std::filesystem::remove_all(std::filesystem::u8path(temp_folder), ec);
 }
 
 class ParamLocalFileTest : public LocalFileTest,
--- a/packager/file/file_unittest.cc
+++ b/packager/file/file_unittest.cc
@@ -83,10 +83,7 @@ class LocalFileTest : public testing::Test {
       data_[i] = i % 256;
 
     local_file_name_no_prefix_ = generate_unique_temp_path();
-
-    // Local file name with prefix for File API.
-    local_file_name_ = kLocalFilePrefix;
-    local_file_name_ += local_file_name_no_prefix_;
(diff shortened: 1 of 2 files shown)


rlaisqls:
Would it be possible to get a brief reason why this PR isn't being accepted for review yet?
I'm open to making adjustments if there are any concerns with the code.

cosmin (approved):

joeyparrish (approved):

joeyparrish:
I added a test case for creating directories, and it's failing on Windows.  I'll debug.

joeyparrish:
Should be fixed now.

joeyparrish:
Apologies for the delay! This will be in the next release, which will be released as soon as I finish going through pending PRs.

joeyparrish:
Released in v3.5.0.  Thanks!

PR #1539, opened 2026-01-15
feat: Add --cencv1 support

TheModMaker:
This allows generating legacy content for testing purposes.

PR #1540, opened 2026-01-15
fix: Move cipher block align to start of block

TheModMaker:
Before, we would add clear ranges to the start of the subsample, therefore moving the encrypted start further into the video frame; now we add clear ranges to the end by adding a new subsample (if needed).

This seems more correct. Both are allowed by the spec, but it would be better to encrypt the beginning of the frame to give attackers less info about the start of the frame. Having the clear data to the end of the frame doesn't give attackers much info since most data depends on the encrypted state.


joeyparrish:
Merged in main, fixed lint, and running CI now.

> Before, we would add clear ranges to the start of the subsample, therefore moving the encrypted start further into the video frame; now we add clear ranges to the end by adding a new subsample (if needed).

Why do we do this now?  Is it better or more correct?

> Depends on https://github.com/shaka-project/shaka-packager/pull/1539

Is this change specific to cencv1, or does it only depend on it because this change has to touch things touched by that?

TheModMaker:
> Merged in main, fixed lint, and running CI now.
> 
> > Before, we would add clear ranges to the start of the subsample, therefore moving the encrypted start further into the video frame; now we add clear ranges to the end by adding a new subsample (if needed).
> 
> Why do we do this now? Is it better or more correct?

It seems more correct.  Both are allowed by the spec, but it would be better to encrypt the beginning of the frame to give attackers less info about the start of the frame.  Having the clear data to the end of the frame doesn't give attackers much info since most data depends on the encrypted state.

> 
> > Depends on #1539
> 
> Is this change specific to cencv1, or does it only depend on it because this change has to touch things touched by that?

Just would cause conflicts.  It would affect all encrypted content.

joeyparrish:
I put your explanation into the PR description.  I'll merge with main again after #1539 lands so the diff is cleaner, re-review, and merge.  Thanks!

joeyparrish:
Looks like some golden output from packager_test.py needs updating.  I'll push an update shortly.

joeyparrish (approved):

joeyparrish:
Released in v3.5.0.  Thanks!

PR #1547, opened 2026-03-06
fix: AdaptationSet ID inconsistency across periods

### LINKED ISSUE #1546: DCHECK_EQ failure due to Period ID mismatch during early Flush
When multiple bitrates/streams are present, they are processed incrementally. If a cuepoint is added, it creates a new Period. If a Flush is called (which happens in OnMediaEnd for each stream) before all bitrates have been added to the new period, Period::GetXml would reassign IDs starting from 0 for whatever bitrates were already there. This lead to a mismatch with the IDs in the previous period, triggering the DCHECK_EQ when the remaining bitrates were processed.

Seems to be a regression, since 3.5.0

xavierlaffargue:
Fixes https://github.com/shaka-project/shaka-packager/issues/1546


joeyparrish (approved):

joeyparrish:
I'm seeing linter failures in CI.  Try `git clang-format --style Chromium main` to reformat all files touched since `main`.

xavierlaffargue:
@joeyparrish Done

joeyparrish:
It looks like some of the golden output files need to be updated with this fix.  You can run `./build/packager/packager_test.py --test_update_golden_files` to update them, then review changes, git add, git commit, git push.  Thanks!

xavierlaffargue:
@joeyparrish Indeed! It's done now

PR #1551, opened 2026-03-12
fix: HLS - remove CLOSED-CAPTIONS from EXT-X-I-FRAME-STREAM-INF

xavierlaffargue:
This PR removes the CLOSED-CAPTIONS attribute (and a few others) from the EXT-X-I-FRAME-STREAM-INF tag in the HLS master playlist.

According to the HLS RFC (Section 4.4.6.3), the CLOSED-CAPTIONS attribute is not defined for EXT-X-I-FRAME-STREAM-INF tags. Including it causes non-compliance and may result in playback issues on strictly compliant players.

References
[RFC 8216 Section 4.4.6.3](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-16#section-4.4.6.3)

Affected version : >= 2.5


joeyparrish (approved):

joeyparrish:
Amazing, thanks!  If you can cite in the PR description what range of Shaka Packager versions are producing non-compliant output, that would be helpful for anyone exploring the changelog later.

joeyparrish:
The one failing job on Docker build looks like some kind of flake.  I'm going to rerun it when the other jobs complete.

PR #1552, opened 2026-03-13
feat(cmake): allow external declaration of version string

niklaskorz:
In nixpkgs, we derive the version from the package definition's version attribute.
As we only fetch the tarball instead of the full git repository, trying to derive
the version from git fails.

We have been using this patch in nixpkgs since 2024, so upstreaming it would be appreciated!


joeyparrish (approved):

joeyparrish:
Thanks so much for contributing!

Issue #13, opened 2015-03-23
HEVC(H.265) Support


nibaboojit:
are edash-packager is HEVC(H.265) codec support plan? 
If you have a support plan to show up?

kqyang:
Yes, that will be likely supported next quarter. I have just created a [public mailing list](https://groups.google.com/forum/#!forum/edash-users) for announcements.You may subscribe to it to receive the updates.

nibaboojit:
thanks for public mailing list :)
but are you saying next quarter means 2nd quarter?

kqyang:
Yes.

nibaboojit:
thank you :)

PR #1553, opened 2026-03-13
fix(deps): update abseil-cpp to 20260107.1, protobuf to 33.5

niklaskorz:
The previous update PR #1508 mentioned updating abseil-cpp while not actually doing so.
The update is included in this PR.
All deprecations have been addresses in this PR and Protobuf has been updated to 33.5 as older versions lead to build failures on Linux with this abseil-cpp version.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@
 
 3Q GmbH <*@3qsdn.com>
 Alen Vrecko <alen.vrecko@gmail.com>
+Alugha GmbH <*@alugha.com>
 Amazon Music <*@amazon.com>
 Anders Hasselqvist <anders.hasselqvist@gmail.com>
 Audible <*@audible.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -47,6 +47,7 @@ Leandro Moreira <leandro.ribeiro.moreira@gmail.com>
 Leo Law <leoltlaw.gh@gmail.com>
 Marcus Spangenberg <marcus.spangenberg@eyevinn.se>
 Michal Wierzbicki <mwierzbicki1@cyfrowypolsat.pl>
+Niklas Korz <nk@alugha.com>
 Ole Andre Birkedal <o.birkedal@sportradar.com>
 Piotr Srebrny <srebrny.piotr@gmail.com>
 Prakash Duggaraju <duggaraju@gmail.com>
--- a/packager/CMakeLists.txt
+++ b/packager/CMakeLists.txt
@@ -19,8 +19,6 @@ endif()
 if(MSVC)
   # Warning level 4 and all warnings as errors.
   add_compile_options(/W4 /WX)
-  # Silence a warning from an absl header about alignment in boolean flags.
-  add_compile_options(/wd4324)
   # Silence a warning about STL types in exported classes.
   add_compile_options(/wd4251)
   # Silence a warning about constant conditional expressions.
--- a/packager/app/job_manager.cc
+++ b/packager/app/job_manager.cc
@@ -86,7 +86,7 @@ Status JobManager::RunJobs() {
   // Wait for all jobs to complete or any job to error.
   Status status;
   {
-    absl::MutexLock lock(&mutex_);
+    absl::MutexLock lock(mutex_);
     while (status.ok() && active_jobs.size()) {
       // any_job_complete_ is protected by mutex_.
       any_job_complete_.Wait(&mutex_);
@@ -119,7 +119,7 @@ Status JobManager::RunJobs() {
 }
 
 void JobManager::OnJobComplete(Job* job) {
-  absl::MutexLock lock(&mutex_);
+  absl::MutexLock lock(mutex_);
   // These are both protected by mutex_.
   complete_[job] = true;
   any_job_complete_.Signal();
--- a/packager/file/http_file.cc
+++ b/packager/file/http_file.cc
@@ -10,6 +10,7 @@
 #include <absl/flags/flag.h>
 #include <absl/log/check.h>
 #include <absl/log/log.h>
+#include <absl/log/vlog_is_on.h>
 #include <absl/strings/escaping.h>
 #include <absl/strings/str_format.h>
 #include <curl/curl.h>
--- a/packager/file/io_cache.cc
+++ b/packager/file/io_cache.cc
@@ -33,7 +33,7 @@ IoCache::~IoCache() {
 uint64_t IoCache::Read(void* buffer, uint64_t size) {
   DCHECK(buffer);
 
-  absl::MutexLock lock(&mutex_);
+  absl::MutexLock lock(mutex_);
   while (!closed_ && (BytesCachedInternal() == 0)) {
     write_event_.Wait(&mutex_);
   }
@@ -63,7 +63,7 @@ uint64_t IoCache::Write(const void* buffer, uint64_t size) {
   const uint8_t* r_ptr(static_cast<const uint8_t*>(buffer));
   uint64_t bytes_left(size);
   while (bytes_left) {
-    absl::MutexLock lock(&mutex_);
+    absl::MutexLock lock(mutex_);
     while (!closed_ && (BytesFreeInternal() == 0)) {
       VLOG(1) << "Circular buffer is full, which can happen if data arrives "
                  "faster than being consumed by packager. Ignore if it is not "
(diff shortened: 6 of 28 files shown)


joeyparrish:
Looks like more changes are going to be needed.  From the CI log:

`#13 11.01 /shaka-packager/packager/media/test/test_web_server.cc:82:33: error: 'absl::lts_20260107::MutexLock::MutexLock(absl::lts_20260107::Mutex*)' is deprecated: Use the constructor that takes a reference instead [-Werror=deprecated-declarations]`

joeyparrish:
And thank you for taking this on!  I really appreciate the contribution.

niklaskorz:
@joeyparrish I've addressed all deprecations, but the `hex_parser.cc` might need another look. Before, it left the output `std::string* bytes` completely alone on invalid input, now it just relies on abseil's own validation, which starts writing into the output but aborts once it encounters invalid input from what I understand. If desired, I can just re-add the previous validation as an additional safeguard. I also removed the other `HexStringToBytes` function from it as it was entirely unused.

The update to a newer Protobuf was also required to resolve include errors on Linux. I first tried updating to Protobuf 34, which led to other issues, so I settled on 33.5 for now which appears to not require any additional changes to work.

joeyparrish:
I looked at your changes in hex_parser and I approve.  As long as the caller is checking the boolean return value, it's fine that the output string is in an arbitrary state.

niklaskorz:
Running clang-format seems to reformat a lot more than intended (in the modified files), looking into that...

Edit: If I read https://github.com/shaka-project/shaka-packager/actions/runs/23066590020/job/67008017329?pr=1553 correctly, they were wrongly formatted before the PR. I can add an additional formatting commit for these files so it stays obvious what is just plain reformatting and what is part of the update.

joeyparrish:
Yeah, the infra was designed to avoid needing one massive cleanup by only looking at changed files... before my time, but it's causing friction now.  Let's land this PR, then I'll put one up that does a massive all-repo linter cleanup.

niklaskorz:
Alright, I reverted the clang-format, so it's at the pre-format diff again

joeyparrish:
> Let's land this PR, then I'll put one up that does a massive all-repo linter cleanup.

Wait, no, that's backwards.  I'm uploading a PR to cleanup the repo's style now.  Then it will be easier to read the diff in this PR and see the meaningful changes that go with the dependency updates.  Are you okay with rebasing this after my PR lands?

niklaskorz:
> Are you okay with rebasing this after my PR lands?

Yup works for me! (mergiraf will sort it out for me)

joeyparrish:
Okay, clang-format ate some arrays of test data in ways that I disagree with.  I'm having Claude take some time to identify and protect those with clang-off/clang-on, then I'll do a massive clang-format-only PR that should require less scrutiny.  Thanks for your patience!

joeyparrish:
Okay, #1561 landed.  You can rebase off of main and that should greatly simplify the diff here.  Thank you again for working on this!

niklaskorz:
Diff is looking clean now :)

niklaskorz:
Also added the entries for AUTHORS and CONTRIBUTORS now, looks like I forgot that in #1552

joeyparrish (approved):

joeyparrish:
Thank you so much!  Looks great.  I'll merge it if the CI passes.

joeyparrish:
Trimmed from the windows build log:

```
packager\third_party\protobuf\source\src\google\protobuf\micro_string.h(87,32):
error C2220: the following warning is treated as an error 
[build\packager\mpd\base\mpd_media_info_proto.vcxproj]
  (compiling source file 'media_info.pb.cc')
  
packager\third_party\protobuf\source\src\google\protobuf\micro_string.h(87,32):
warning C4458: declaration of 'size' hides class member
[build\packager\mpd\base\mpd_media_info_proto.vcxproj]
  (compiling source file 'media_info.pb.cc')
      packager\third_party\protobuf\source\src\google\protobuf\micro_string.h(72,14):
      see declaration of 'google::protobuf::internal::MicroString::LargeRep::size'
```

joeyparrish:
My first thought was that we had a proto field called `size` that's conflicting with some generated internals, but we don't.  Not sure what's up with that.  Maybe it's fixed in a newer protobuf?

niklaskorz:
I'm trying to find a way to get the same warnings on a GCC build to make it easier to debug, but `-Werror=shadow` only shows other unrelated occurrences of member shadowing that so far did not cause a CI failure.

niklaskorz:
> Maybe it's fixed in a newer protobuf?

No, it's unchanged in latest protobuf:

- https://github.com/protocolbuffers/protobuf/blob/686782208778e6fd9bbfe8df73edca46b698c389/src/google/protobuf/micro_string.h#L87
- https://github.com/protocolbuffers/protobuf/blob/686782208778e6fd9bbfe8df73edca46b698c389/src/google/protobuf/micro_string.h#L109

I think there is no other solution than disabling this specific warning for now and reporting it to protobuf.

joeyparrish:
> I think there is no other solution than disabling this specific warning for now and reporting it to protobuf.

That makes sense.  Based on the failing target (build sources generated from our protos), you may have to fix it with another `add_compile_options` inside `if(MSVC)` in `packager/CMakeLists.txt`.  It seems like you wouldn't be able to limit the scope of that flag to `packager/third_party/protobuf/CMakeLists.txt`.

I updated the repo settings so that CI will run automatically for people like you who have contributed before.  Feel free to iterate on this until it's working, and I'll check back in later today.

joeyparrish:
Oh, and if you can please avoid a force-push from here on, it will make it easier for me to see the delta from what I already approved when I review the updates later.  Thanks!

niklaskorz:
I'll try to get hold of a Windows machine tomorrow to see what's needed for a passing build without the whole CI back and forth.

Edit: oh it's only Windows debug builds that fail now, release passes, interesting.

joeyparrish:
> I'll try to get hold of a Windows machine tomorrow to see what's needed for a passing build without the whole CI back and forth.

You can trigger the CI on your fork with the "debug" input checked, and then you'll be able to SSH into GitHub's VM with the same SSH key registered to your GitHub account.  If that sounds better than getting a Windows machine locally, you could try it.  I usually debug that way for Packager.

joeyparrish:
Maybe it's only debug because of a DCHECK on some part of a proto?

```
see reference to function template instantiation
'const char *absl::lts_20260107::log_internal::Check_GEImpl<unsigned int,T>(
const T1 &,const T2 &,const char *)' being compiled
```

joeyparrish:
~Ex:~

~packager/hls/base/master_playlist.cc:  DCHECK_GE(audio_variants.size(), 1u);~

~Maybe .size() is signed now?  (Maybe only on Windows?? 🤷)~

Never mind.  That's not a proto instance.

niklaskorz:
It's again caused by protobuf's own `MicroString` implementation (that one is really attracting MSVC warnings it seems 😄), where they are comparing unsigned integers to enum values. The C++ standard actually doesn't dictate whether enums are signed or unsigned, so it's up to the compiler to decide. The debug check that makes the build fail is this one: https://github.com/protocolbuffers/protobuf/blob/686782208778e6fd9bbfe8df73edca46b698c389/src/google/protobuf/micro_string.h#L78

Curious because they actually re-enabled this particular warning type in https://github.com/protocolbuffers/protobuf/pull/9263

niklaskorz:
Explanation to the new changes:

- Setting `SYSTEM` on `add_subdirectory` makes all targets defined inside of that directory be treated like external dependencies for code analysis, as if they were imported through `find_package`. This way we can even avoid disabling MSVC warning C4324 (boolean alignment) project-wide, which was only disabled because of abseil-cpp in the past.
- I've only marked protobuf and abseil-cpp as SYSTEM for now as these were the only ones causing problems here. It might make sense to add it to the whole `add_subdirectory(third_party)` in the future, and I think the `EXCLUDE_FROM_ALL` can also be moved there instead of declaring it for every dependency individually.
- Fedora 35 failed to copy some protobuf files with this flag enabled, which seems to have been a bug in that particular cmake version. As Fedora 35 has been EOL for over three years by now, I've chosen to update it to the oldest still supported Fedora version instead, which does not have this issue.

joeyparrish:
> * I've only marked protobuf and abseil-cpp as SYSTEM for now as these were the only ones causing problems here. It might make sense to add it to the whole `add_subdirectory(third_party)` in the future, and I think the `EXCLUDE_FROM_ALL` can also be moved there instead of declaring it for every dependency individually.

I really like this idea.  Would be happy to review a PR for those changes as well!

PR #1556, opened 2026-03-13
ci: Make linting happen entirely in parallel

joeyparrish:
It has been very, very annoying lately to have PRs get stuck in several loops of linter failures without finding out if the thing builds or passes tests.  From now on, lint in parallel with everything else.


joeyparrish:
@tobbee, @niklaskorz, JFYI

avelad (approved):

PR #1562, opened 2026-03-16
feat(cmake): allow using system dependencies

niklaskorz:
Another of our currently vendored patches for nixpkgs / NixOS that has been in use for about two years, hidden behind an option to allow both variants to coexist.
This might need some more back and forth discussion to get into an acceptable state for upstream.
The idea is that opting into `USE_SYSTEM_DEPENDENCIES` works without any submodules, so when this option is enabled, we also expect the system to already have Python with the `google.protobuf` module ready and available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,10 @@ option(BUILD_SHARED_LIBS "Build libpackager as a shared library" OFF)
 # against mimalloc to replace the standard allocator in musl, which is slow.
 option(FULLY_STATIC "Attempt fully static linking of all CLI apps" OFF)
 
+# Whether to look for third-party dependencies on the system through
+# `find_package` instead of building them from git submodules.
+option(USE_SYSTEM_DEPENDENCIES "Search for dependencies on the system" OFF)
+
 # Enable CMake's test infrastructure.
 enable_testing()
 
--- a/packager/CMakeLists.txt
+++ b/packager/CMakeLists.txt
@@ -73,6 +73,47 @@ if(NOT Python3_EXECUTABLE)
   find_package(Python3 COMPONENTS Interpreter REQUIRED)
 endif()
 
+if(USE_SYSTEM_DEPENDENCIES)
+  # find_package results cannot be propagated to parent directories, so we have to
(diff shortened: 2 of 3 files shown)


joeyparrish (changes requested):
All of this makes sense to me from a packaging perspective.

joeyparrish:
My only concern now is that packager/third_party/CMakeLists.txt will get out of sync with packager/CMakeLists.txt, since both list the deps.  I don't have any plans to change/add dependencies right now, but it may happen in the future.  And with no CI or Dockerfile to do a system-deps build, I might have a hard time maintaining that.

I'm okay to merge this as-is, but follow-ups like these would be very welcome:
 - Bring the lists of deps (system and git-submodule) into one location
 - Add a Dockerfile to validate a system-deps build on the distro of your choice (maybe with the others in ./packager/testing/dockers/, which get picked up automatically in CI)

joeyparrish (approved):

niklaskorz:
Reasonable request, thank you! Longterm, it might even be preferable to replace the submodules with [FetchContent_Declare](https://cmake.org/cmake/help/latest/module/FetchContent.html), which can be made to automatically search for system dependencies first and then fallback to fetching and building the dependency. That at least helps avoid the issue of having to declare them twice. I had a quick shot at that too but it's not a trivial change at least, so if we can move that to a follow-up that'd be much appreciated.

joeyparrish:
> Longterm, it might even be preferable to replace the submodules with [FetchContent_Declare](https://cmake.org/cmake/help/latest/module/FetchContent.html), which can be made to automatically search for system dependencies first and then fallback to fetching and building the dependency.

Our preference is hermetic builds with deps included, so I don't think I would like a system that automatically searches for system deps and prefers them.  That would require the build system to be cleaned of all deps before a hermetic build could be possible, and that's hard.

niklaskorz:
`FetchContent_Declare` defaults to not looking for system dependencies, it only does so when `FIND_PACKAGE_ARGS` is passed as well. But there also aren't that many external dependencies in shaka-packager, so I guess the benefit is negligible.

PR #1563, opened 2026-03-18
fix(deps): update to curl v8.19.0, mbedtls v3.6.5, libxml2 v2.15.2

niklaskorz:
These are security sensitive dependencies that haven't been updated since 2024.
Thankfully, there do not appear to be any major incompatibilities that prevent an update.

libxml2 now enforces that buffer size must only refer to the actual content excluding the null terminator, which required replacing `std::size` with `strlen` in two unit tests.

I originally tried to bump mbedtls to 4.0.0, but that seems to be blocked by an upstream issue trying to write to the cmake source directory, which cmake errors on.
3.6.5 has all CVE fixes included that come with 4.0.0 though, so from a security perspective it is sufficient to update to this version with no incompatibilities first.


niklaskorz:
The static build failure is caused by the explicit `-I/usr/include` in the musl-gcc invocation I think (which leads to the inclusion of glibc headers). I have yet to find out what's sneaking that in...

joeyparrish:
The -I/usr/include seems to come from the zstd library, which we aren't grabbing explicitly.  I found it by `grep -r -I/usr/include build/`, which turned up:

```
build/CMakeCache.txt:_zstd_CFLAGS:INTERNAL=-I/usr/include
build/CMakeCache.txt:_zstd_STATIC_CFLAGS:INTERNAL=-I/usr/include
```

This library is being pulled in by the updated curl.  I'm pushing a fix that syncs all our curl options against the diff between the old & new curl's CMakeLists.txt.

joeyparrish:
I'll audit the changes to the other deps' options soon.

joeyparrish:
I pushed changes to sync our flags in libxml2, abseil-cpp, and libwebm, too.  The only one that was really problematic was curl with that new system dep.  The others were just "disable this new build target we don't need" and "remove this flag that doesn't exist any more".  No changes were needed in other recent updates (mbedtls, json, protobuf, c-ares).

joeyparrish (approved):

joeyparrish:
It's already approved by me, but please mark "ready for review" if you sign off on my contributions to this PR, and then I'll merge it.

cosmin (approved):

PR #1566, opened 2026-03-31
fix: Ignore already-deleted files when removing old segments

### LINKED ISSUE #1367: Packager fails to remove old media segments if both HLS and DASH is being used for LIVE segmentation
## System info
**Operating System**: Ubuntu 22.04.2 LTS
**Shaka Packager Version**: 5ee2b7f-release

**Packager Command**:
packager-linux-x64 in=udp://230.2.2.222:1111?interface=172.22.238.253&reuse=1,stream=audio,drm_label=AUDIO,init_segment=audio_init.mp4,segment_template=audio_$Number$.m4s,playlist_name=/mnt/life/audio.m3u8,hls_group_id=audio in=udp://230.2.2.222:1111?interface=172.22.238.253&reuse=1,stream=video,init_segment=p00_init.mp4,segment_template=p00_$Number$.m4s,playlist_name=p00.m3u8 in=udp://230.2.2.222:1112?interface=172.22.238.253&reuse=1,stream=video,init_segment=p01_init.mp4,segment_template=p01_$Number$.m4s,playlist_name=p01.m3u8 in=udp://230.2.2.222:1113?interface=172.22.238.253&reuse=1,stream=video,init_segment=p02_init.mp4,segment_template=p02_$Number$.m4s,playlist_name=p02.m3u8 --hls_master_playlist_output /mnt/life/life.m3u8 --hls_playlist_type LIVE --segment_duration 6.4 --preserved_segments_outside_live_window 2 --time_shift_buffer_depth 32 --mpd_output /mnt/life/life.mpd --suggested_presentation_delay 25

What is the expected result?

Old (outside the live window) media segments should be removed without an error.

What happens instead?

In case of using both HLS and DASH manifest generation, both formats' "RemoveOldSegment" function is being triggered. The media segments are removed by the first format's invocation, but the other format shows a warning message and keeps the segment filenames in a list (to try to remove them later). This "segments_to_be_removed_ list" will grow forever, which can lead to OOM / memory leak.

Media segments should be removed only once in a common code section, or if the file removal is being failed with a "not found" error, then the filename should be popd from the removal queue.

joeyparrish:
When generating both DASH and HLS outputs, they could both be tracking the same files and racing to delete them as they fall out of the live window. If one output deletes a segment before the other tries to do so, File::Delete() returns false. We only want to retry if the file actually exists but failed to delete. We can use GetFileSize() to check if the file still exists (>= 0).

Also updated MemoryFile::Delete to return a boolean indicating whether the file existed and was deleted successfully so that we can accurately mock file deletion behaviors in unit tests.

Fixes #1367

### DIFF
--- a/packager/file/file.cc
+++ b/packager/file/file.cc
@@ -142,8 +142,7 @@ File* CreateMemoryFile(const char* file_name, const char* mode) {
 }
 
 bool DeleteMemoryFile(const char* file_name) {
-  MemoryFile::Delete(file_name);
-  return true;
+  return MemoryFile::Delete(file_name);
 }
 
 static const FileTypeInfo kFileTypeInfo[] = {
--- a/packager/file/memory_file.cc
+++ b/packager/file/memory_file.cc
@@ -31,17 +31,17 @@ class FileSystem {
     return &instance;
   }
 
-  void Delete(const std::string& file_name) {
+  bool Delete(const std::string& file_name) {
     absl::MutexLock auto_lock(mutex_);
 
     if (open_files_.find(file_name) != open_files_.end()) {
       LOG(ERROR) << "File '" << file_name
                  << "' is still open. Deleting an open MemoryFile is not "
                     "allowed. Exit without deleting the file.";
-      return;
+      return false;
     }
 
-    files_.erase(file_name);
+    return files_.erase(file_name) > 0;
   }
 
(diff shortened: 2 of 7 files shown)


joeyparrish:
@cosmin, can you take a look?

cosmin (approved):

PR #1572, opened 2026-04-06
feat: add HLS AES-128 clear-key encryption for TS + MP4

### LINKED ISSUE #776: How to implement plain AES-128 encryption while packaging for HLS?
## System Info
**Operating System**: Ubuntu 20.04
**packager Version**: v2.3.0-5bf8ad5-release

**How to implement plain AES-128 encryption in HLS? I didn't find it in the documentation and other issues.**

What is the expected result?
Plain AES-128 encryption to be implemented so that players like shaka and videoJS will play them without any configuration. AES-128 encryption of HLS like that done by Bento4 is expected. 

I just want to encrypt my HLS stream using plain AES-128 encryption and provide a URL to access the key but I can't actually find how to implement that using shaka-packager. Please point me in the right direction @kqyang . Thank you in advance.

vish91:
fixes: #776 

**HLS AES-128 Clear-Key Encryption**
Implements plain AES-128 segment encryption for HLS

**What it does**
Adds --protection_scheme aes128 support for HLS+TS outputs. Encrypts entire TS segments with AES-128-CBC and produces #EXT-X-KEY:METHOD=AES-128 tags — no DRM infrastructure required, works natively in all HLS players (hls.js, VideoJS, Shaka, Safari).

**Usage**
```
packager \
  'in=input.mp4,stream=video,output=seg_%05d.ts,playlist_name=video.m3u8' \
  'in=input.mp4,stream=audio,output=seg_%05d.ts,playlist_name=audio.m3u8' \
  --protection_scheme aes128 \
  --enable_raw_key_encryption \
  --keys label=:key_id=<16-byte-hex>:key=<16-byte-hex> \
  --hls_key_uri https://keys.example.com/stream.key \
  --hls_master_playlist_output master.m3u8
```

**Changes**
- New FOURCC_aes8 / kProtectionSchemeAes128 constant
- EncryptionConfig::key field for carrying raw key to segment encryptor
- TsMuxer::EncryptSegmentIfNeeded() — full-segment AES-128-CBC encryption
- SimpleHlsNotifier — generates EXT-X-KEY:METHOD=AES-128 without KEYFORMAT (identity default per RFC 8216)
- Unit tests + integration golden test
- Also syncs to upstream master

Co-author: 🤖 Generated with [Claude Code](https://claude.com/claude-code)


joeyparrish:
Does this only support AES-128 full-segment encryption for TS, or will it also work for MP4 outputs?

joeyparrish:
I've skimmed the PR now and I see that it only uses the key material to encrypt in the TS writer.  Could you expand this to cover MP4 outputs as well?

joeyparrish:
And to fail on unsupported output types?  It would be a real problem to configure for full-segment encryption + WebM, for example, and end up generating unencrypted WebM outputs.  (Even though WebM is not supported in HLS, it could still be a way for someone to accidentally leak plain content.)

vish91:
Yeah just the TS writer for now.. didn't think about mp4 for this protection type to be honest.. 
Do you think we should scope it or just throw error ? 
Also for mp4 would that mean like just encrypted the whole m4s / fmp4  segment ?

joeyparrish:
I believe that HLS AES-128 encryption is whole-segment encryption no matter the underlying format, so I think MP4 should be easy enough to add.  Are you willing to try adding it?

vish91:
> I believe that HLS AES-128 encryption is whole-segment encryption no matter the underlying format, so I think MP4 should be easy enough to add. Are you willing to try adding it?

yeah makes sense.. i am trying on side for couple days and got it working.. had claude write tests .. check it out see if it makes sense

joeyparrish (approved):

PR #1577, opened 2026-04-28
fix: hls default yes

SteveR-PMP:
This ensures that "first rendition encountered per (group, language)" refers to the first rendition in the final output order, not the arbitrary input arrival order. This is particularly important when --force_cl_index is enabled by default, which can result in different output order than input order.


xavierlaffargue:
@SteveR-PMP  Your branch seems to be not up to date. Can you try to merge main into your branch?

Issue #1567, opened 2026-03-31
IWYU (Include What You Use) violation audit


joeyparrish:
A frequent cause of pain and wasted time in PRs is violations of the IWYU (Include What You Use) rule.  If headers are only implicitly included for a dependency (e.g. `std::vector` used without `#include <vector>`), that may not always be true across operating systems, toolchains, or even compiler versions.  Any variation in the standard library or OS headers could cause a shift in what is implicitly included.  Therefore, we should never depend on implicit inclusion.  This will help avoid PRs that work fine at home and fail on another OS in CI, in ways that are difficult to spot and understand.

joeyparrish:
We should audit the codebase for these violations and fix them, to make future PRs easier for all involved.

joeyparrish:
The source has been audited and violations fixed, but there was enough manual effort required that I'm hesitant to finish the project by integrating enforcement.  Some of the things that made this hard:

 - Linux-specific headers added without OS guards
 - Kernel headers added instead of the normal entrypoints (bits/... instead of sys/...)
 - Requires several passes: run clang-include-cleaner on headers first, then on CC files, then commit (since clang-format won't run against pending changes), then run clang-format (since clang-include-cleaner won't format), then amend commit, plus the iteration of human review & CI builds across platforms

PR #1589, opened 2026-07-06
chore: Clean up all IWYU violations

joeyparrish:
This is a massive audit and cleanup of all IWYU (Include-What-You-Use) violations.  This was done with clang-include-cleaner-20, clang-format-20, some new clang-format configs, and a bunch of scripting, and a bunch of manual labor.  The tooling for this will be introduced into workflows soon, to maintain enforcement.

This is important because many cross-platform build errors in PRs can be traced to IWYU violations.  If the source isn't strict about IWYU, we can end up with different implicit includes on different platforms. A PR that builds find on your Linux machine may not work on Mac or Windows.  Or a PR that works with gcc won't work with clang.

By strictly including the right headers everywhere, we can avoid those surprise build failures on other platforms that are so hard to debug.

One change here that doesn't fit with the others:
packager/media/base/text_track.h was not referenced anywhere.  Rather than make it compile, I deleted it.

Issue #1567


avelad (approved):

avelad:
LGTM, but you need fix the build errors

joeyparrish:
> LGTM, but you need fix the build errors

On it.  Some of the tooling inserted linux-specific headers like bits/* that you would normally not include directly.

PR #1594, opened 2026-07-12
fix: correct AES-128 whole-segment encryption for MP4/CMAF HLS

### LINKED ISSUE #1587: bug: --protection_scheme aes128 with single_segment=true emits non-16-byte-aligned HLS byte-ranges, breaking AES-CBC decryption
### **Packager Version**

`v3.8.0` (Regression introduced in recent `#EXT-X-KEY:METHOD=AES-128` support for fMP4)

### **Summary**

When using Shaka Packager to emit a single-file fragmented MP4 (fMP4/CMAF) asset via `single_segment=true` combined with the raw-key `--protection_scheme aes128` full-segment encryption engine, the calculated subsegment byte-ranges emitted in the HLS `index.m3u8` manifest are not padded to multiples of 16 bytes.

Because `METHOD=AES-128` treats each independent HLS byte-range as a distinct ciphertext stream envelope decrypted via AES-CBC, native player engines (such as Android Media3/ExoPlayer’s `Aes128DataSource`) reject non-16-byte-aligned ranges, crashing with immediate block-size padding / invalid length errors.

---

### **Steps to Reproduce**

Execute a single-file automated raw-key packaging pass over an asset (e.g., audio source):

```bash
packager \
  input=source_audio.wav,stream=audio,output=audio.mp4,playlist_name=index.m3u8 \
  --enable_raw_key_encryption \
  --keys label=AUDIO:key_id=00112233445566778899aabbccddeeff:key=00112233445566778899aabbccddeeff \
  --protection_scheme aes128 \
  --clear_lead 0 \
  --single_segment true \
  --segment_duration 4

```

---

#### 1. Generated Playlist Slices (`index.m3u8`):

The generated index lists initialization map boxes cleanly, but the subsegment references break block boundaries:

```manifest
#EXTM3U
#EXT-X-VERSION:6
#EXT-X-TARGETDURATION:5
#EXT-X-MAP:URI="audio.mp4",BYTERANGE="833@0"
#EXT-X-KEY:METHOD=AES-128,URI="enc.key",IV=0x00112233445566778899aabbccddeeff
#EXT-X-BYTERANGE:66328@901
audio.mp4
#EXT-X-BYTERANGE:131000@67229
audio.mp4

```

#### 2. Local Decryption Verification Failing on Single-File Ranges:

Attempting to decrypt an extracted byte-range slice directly via openssl or piping it to processing tools highlights the alignment failure:

```bash
# Slicing the exact byte-range from the single asset:
dd if=audio.mp4 of=segment_1.m4s bs=1 skip=901 count=66328

# Attempting decryption:
openssl enc -d -aes-128-cbc -in segment_1.m4s -out clear.m4s -K 00112233445566778899aabbccddeeff -iv 00112233445566778899aabbccddeeff

```

**Output Error:**

```text
bad decrypt: wrong final block length

```

*(Note: 66328 mod 16 = 8. The block structure lacks padding).*

#### 3. Client Engine Failures:

On Android applications utilizing `Media3/ExoPlayer`, the player halts with an immediate `CryptoException` or `ParserException` originating from the native `Aes128DataSource` because the final chunk fails PKCS#7 / padding compliance.

---

### **Expected Behavior**

When `single_segment=true` is coupled with `--protection_scheme aes128`, Shaka Packager should automatically inject padding or alter interior ISOBMFF subsegment allocations (such as padding out the `mdat` container sizes or fragment boundaries) so that every emitted `#EXT-X-BYTERANGE` block length is exactly divisible by 16 bytes.

* **Contrast with Multi-Segment Mode:** When running the exact same command *without* `single_segment=true`, Shaka successfully creates standalone chunk files (e.g., `s1.m4s`, `s2.m4s`) that are perfectly 16-byte aligned (e.g., lengths of 66416, 131440, 164320 where length mod 16 = 0) and decrypt flawlessly. The single-file mode should match this mathematical alignment constraint.

---

### **Cryptographic Root Cause**

The implementation maps the `#EXT-X-BYTERANGE` tags strictly to the natural interior sizing of the fMP4 box fragments (`moof` + `mdat`). While this is completely fine for standard unencrypted streaming or true CENC sample-level models (`cbcs`/`cenc`), it is mathematically incompatible with full-segment `METHOD=AES-128` encryption.

Because `AES-128` mode forces the player to perform an overall block-cipher decryption over the complete extracted HTTP network range before parsing the box structures, any chunk length that does not hit `length % 16 == 0` causes a total cryptographic breakdown.

---

### **Identified Temporary Workaround**

We verified that we can manually fix this by packaging in multi-segment mode (which yields natively aligned files), concatenating the encrypted segments sequentially into a single container via shell scripts, and manual-building a byte-range manifest matching the stitched offsets. Shaka should natively apply this alignment constraint when writing single-file outputs directly.

### LINKED ISSUE #1588: bug: --protection_scheme aes128 in multi-segment mode appends chunks cumulatively instead of flushing at boundaries
### **Packager Version**

`v3.8.0` (Regression introduced in the new HLS `METHOD=AES-128` full-segment envelope encryption implementation)

### **Environment / Operating System**

Reproduced identically on both `Linux x64 (Debian/Ubuntu container environments)` and `Windows x64 desktop binaries`.

---

### **Description of the Bug**

When executing a multi-segment HLS packaging pass over a media container (e.g., extracting or fragmenting a stream) using the new full-segment envelope encryption engine (`--protection_scheme aes128`), the underlying output stream filter layer fails to clear its internal cipher context, flush its memory buffers, or reset its stream write-pointer when crossing segment duration fragment boundaries.

Consequently, each sequential fragment file output (e.g., `seg_2.m4s`, `seg_3.m4s`, etc.) does not store an isolated segment of media. Instead, it cumulatively appends all previous media chunks from the beginning of the stream and re-encrypts the entire accumulated block.

For longer media timelines, this causes a catastrophic triangular-sum payload footprint inflation, easily ballooning a standard **17-minute audio file into a corrupted 4.23 GB payload.**

---

### **Minimal Working Replication Command**

Run a standard multi-segment fMP4 packaging execution loop over any valid input file with full-segment `aes128` tracking active and clear lead disabled:

```bash
packager \
  input=encode.mp4,stream=audio,output=seg_$Number$.m4s,init_segment=init.mp4,playlist_name=index.m3u8 \
  --enable_raw_key_encryption \
  --keys label=AUDIO:key_id=00112233445566778899aabbccddeeff:key=00112233445566778899aabbccddeeff \
  --protection_scheme aes128 \
  --clear_lead 0 \
  --segment_duration 4

```

---

#### **Actual Behavior (Broken `aes128` Output Tracking)**

Inspecting the resulting directory shows that the physical byte size of each subsequent segment file expands continuously by roughly the footprint of an entire base segment:

* `seg_1.m4s` $\rightarrow$ **120 KB** (Contains exclusively seconds 0–4)
* `seg_2.m4s` $\rightarrow$ **235 KB** (Contains seconds 0–4 + seconds 4–8 concatenated)
* `seg_3.m4s` $\rightarrow$ **350 KB** (Contains seconds 0–4 + 4–8 + 8–12 concatenated)
* `seg_266.m4s` $\rightarrow$ **31.7 MB** (Contains the entire prior 17.7 minutes of content re-concatenated)

The total storage footprint scales as a triangular series rather than a linear progression. For a uniform base segment size $S$ and total segments $N$:

$$\text{Total Footprint} = S \times \frac{N(N+1)}{2}$$

#### **Expected Behavior (Correct Standard Output Tracking)**

The full-segment `aes128` wrapper filter should preserve the exact fragment-flushing lifecycle properties seen in the sample-based paths. Running the exact same command with sample-based parameters (`--protection_scheme cenc` or `--protection_scheme cbcs`) yields a clean, uniform, non-cumulative asset distribution as expected:

* `seg_1.m4s` $\rightarrow$ **120 KB**
* `seg_2.m4s` $\rightarrow$ **119 KB**
* `seg_3.m4s` $\rightarrow$ **120 KB**
* `seg_266.m4s` $\rightarrow$ **121 KB**

---

### **Cryptographic Root Cause Analysis**

Because `cenc` and `cbcs` utilize sample-based common encryption schemas, they mutate specific interior tracking block arrays (`moof`/`mdat` configurations) directly without interfering with the overarching multiplexer chunk-writing hooks.

Conversely, full-segment `METHOD=AES-128` functions as a macro binary output filter stream wrapper around the final serialized block bytes. The current multiplexer loop structure misses or fails to trigger a critical context-reset or boundary-flush operation on the active `aes128` cryptographic cipher stream state filter when hitting segment-transition markers, causing all previous ciphertext bytes to cascade indefinitely into subsequent output file descriptors.

shunyagatha:
## Summary

Fixes two related bugs in the HLS `--protection_scheme aes128` whole-segment encryption path for fMP4/CMAF, both introduced by the same feature.

**Fixes #1588** — multi-segment mode (`--segment_template` set) grows every segment file cumulatively instead of independently.

`MultiSegmentSegmenter::WriteSegment`'s AES-128 branch copies `fragment_buffer()` into a local `BufferWriter` via `AppendBuffer()`, but `AppendBuffer()` only copies — it never drains the source. The non-AES-128 branch happens to drain it as a side effect of `WriteToFile()` (which clears its buffer after writing, see `BufferWriter::WriteToFile`), so only the AES-128 path leaked. Each segment then re-encrypts every prior segment's plaintext bytes plus its own — a triangular-sum size blowup (the reporter's 17-minute audio track ballooned to 4.2 GB) and wildly overstated `#EXT-X-STREAM-INF` `BANDWIDTH` (measured **3.4x high** on the fixed test asset: `3756250` → `1108115`).

Fix: explicitly `Clear()` `fragment_buffer()` right after copying it out (one line + a comment explaining why the non-AES-128 branch didn't need it).

**Fixes #1587** — single-segment mode (VOD, no `--segment_template`) never implemented AES-128 whole-segment encryption at all.

`MP4Muxer::Initialize` unconditionally calls `SetAes128EncryptionConfig()` on whichever segmenter was constructed, but `SingleSegmentSegmenter::DoFinalizeSegment` silently ignored the config and wrote **plaintext** to disk, while the HLS manifest still advertised `#EXT-X-KEY:METHOD=AES-128` with byte ranges sized from the natural (unencrypted) fragment lengths. Since those lengths are essentially never multiples of 16, any AES-CBC decrypt of an extracted byte range failed with `bad decrypt: wrong final block length` — exactly the symptom in the report.

Fix: encrypt each subsegment independently (PKCS7-padded CBC, same construction `MultiSegmentSegmenter` already uses) before writing it to the temp file, and correct the recorded `referenced_size` to the actual ciphertext length so `GetSegmentRanges()` reports byte ranges that match what's actually on disk. Key-frame byte-offset (I-frame playlist) reporting is skipped for AES-128 subsegments, mirroring `MultiSegmentSegmenter`'s existing behavior, since a keyframe partway through a whole-block-CBC-encrypted subsegment isn't independently seekable in the ciphertext.

## Testing

- Added `testSingleSegmentMp4HlsWithAes128Encryption`, a new golden end-to-end test covering the previously-untested single-segment + AES-128 combination.
- Regenerated the existing `testFmp4HlsWithAes128Encryption` golden files. The **previously-checked-in golden already encoded bug #1588** — segment sizes grew `17120 → 33408 → 42960` instead of shrinking on the shorter final segment. I confirmed this against the parallel, unaffected TS-muxer AES-128 golden (`avc-ts-with-aes128-encryption`, a completely separate code path), which already showed flat, non-cumulative sizes (`23328, 24256, 17312`) for the identical source clip.
- Manually verified end-to-end with the **exact reproduction steps from both bug reports**: extracted each HLS byte range with `dd` and decrypted it with `openssl enc -d -aes-128-cbc`, confirming every range is a 16-byte-aligned, independently-decryptable ciphertext that parses back to a valid ISOBMFF `moof`/`styp` box (previously: `bad decrypt: wrong final block length`).
- Full local Debug build (Windows/MSVC 14.29, cmake+ninja) + full local test run:
  - `mp4_unittest`: 351/351 passing
  - `hls_unittest`: 120/120 passing
  - `media_base_unittest`: 432/432 passing
  - `packager_test.py` (all golden end-to-end tests, not just the two touched by this change): 163 ran, all passing (4 skipped, pre-existing — missing external AES/RSA credentials, unrelated to this change)
- `git clang-format main --diff` clean (clang-format 22, within the project's documented 18–22 compatible range).

## Notes for reviewers

- No behavior change for `cenc`/`cbcs` (sample-level encryption) or unencrypted packaging — both fixes are scoped to the `protection_scheme == kAes128ProtectionScheme` branch.
- `SingleSegmentSegmenter::DoFinalizeSegment` was restructured (if/else around the encrypt-vs-plain path) but the non-AES-128 branch's logic is unchanged, just moved; the full golden suite passing (including many single-segment, non-AES-128 tests: WVM, DASH VOD, etc.) confirms no behavior change there.
- Happy to split this into two PRs if preferred — I combined them since both live in the same feature (whole-segment AES-128) and the same file family, and fixing #1588 first is what let me generate a trustworthy golden for the #1587 test.

Fixes #1588
Fixes #1587


joeyparrish (approved):

joeyparrish:
Thank you for your contribution!

PR #1611, opened 2026-07-21
fix: Fix parsing of av1 metadata OBUs

mlrsmith:
If an AV1 input contains metadata OBUs in the MP4 codec configuration box (av1c)'s configOBUs field, shaka-packager fails to parse them and errors out.

The common use-case for this is HDR metadata, where MDCV and CLL metadata OBU should be present in configOBUs (per https://aomediacodec.github.io/av1-isobmff/ ):

```
For sample entries corresponding to HDR content, the following applies:

- The ContentLightLevelBox [clli](https://www.iso.org/standard/68960.html#) SHOULD be present. Its values SHALL equal the values contained in the [Metadata OBU](https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=46) of type METADATA_TYPE_HDR_CLL, if present (in the configOBUs or in the samples).
- The MasteringDisplayColourVolumeBox [mdcv](https://www.iso.org/standard/68960.html#) SHOULD be present. Its decoded values SHALL equal (except for precision issues, see the second note below) the decoded values contained in the [Metadata OBU](https://aomediacodec.github.io/av1-spec/av1-spec.pdf#page=46) of type METADATA_TYPE_HDR_MDCV, if present (in the configOBUs or in the samples).
```

Shaka-packager doesn't really need to parse these, and so attempts to skip them. However, it skips the entirety of the OBU (including trailing bits), then validates that the trailing bits can be read. Since we've already consumed them, this fails.

Here, we just skip the check for trailing bits if we're skipping over the OBU entirely - to do otherwise requires full parsing of these structures, which is unnecessarily complicated.


joeyparrish:
This is great, thank you!  Can you add a test case?

mlrsmith:
I'll try to add a test case and get the CLA thing sorted out - that may take a while, sorry!

joeyparrish:
I'm not strict on the CLA.  (I can't turn that off until the project leaves Google.)  A test case will suffice.

mlrsmith:
Thanks. Here's a (passing, and failed without the change) unit test.

joeyparrish (approved):

PR #1617, opened 2026-08-03
feat(mp4): multiplex several streams into one MP4 output file

### LINKED ISSUE #1480: Encrypt both audio and video streams to the same file
I manage to encrypt the audio and video streams to two different output files with something like:

```
./packager input=test-video.mp4,stream=video,output=test-video-encrypted.mp4,drm_label=video \
input=test-video-reencoded.mp4,stream=audio,output=test-audio-encrypted.mp4,drm_label=audio \
--enable_raw_key_encryption \
--keys label=video:key_id=20ae55a11663d1a2947be3ede4e2b123:key=5f07cf01b5c31d0f2251f9caad63a5a3:iv=b5096ef9bc43de36e8b57c9344c85a84,label=audio:key_id=20ae55a11663d1a2947be3ede4e2b123:key=5f07cf01b5c31d0f2251f9caad63a5a3:iv=b5096ef9bc43de36e8b57c9344c85a84
```

But is there a way to encrypt them both to the same output file?

I have tried doing it manually with:

```
ffmpeg -i test-video-encrypted.mp4 -i test-audio-encrypted.mp4 -c:v copy -c:a copy final_encrypted.mp4
```

But then both streams are marked as "not encrypted" (though they definitely are) when I run:

```
./packager input=final_encrypted.mp4 --dump_stream_info
```

Which suggests that maybe it was not the right way?

shunyagatha:
Packaging an input into a single file holding both audio and video was not possible: every stream descriptor got its own `Muxer`, and two descriptors naming the same output were rejected outright as "duplicated outputs". Users who worked around it by remuxing Packager's per-stream outputs with an external tool ended up with files whose encryption signalling had been stripped, so the tracks read back as clear even though the sample data was encrypted — which is what the issue reports.

Several stream descriptors may now name the same output file, which turns them into the tracks of one multiplexed MP4:

```
packager \
  input=in.mp4,stream=video,output=out.mp4,drm_label=video \
  input=in.mp4,stream=audio,output=out.mp4,drm_label=audio \
  --enable_raw_key_encryption --keys label=video:key_id=...:key=...,label=audio:key_id=...:key=...
```

Encryption is applied per stream before muxing, so each track keeps its own `tenc` key id and `drm_label` still selects a different key per track. On the resulting file:

```
$ packager input=out.mp4 --dump_stream_info
Found 2 stream(s).
Stream [0] type: Audio ... is_encrypted: true
Stream [1] type: Video ... is_encrypted: true
```

## Implementation

`mp4::Segmenter` already fragmented multiple tracks into a single `moof`; what was missing was the plumbing and the parts of the pipeline that assumed one stream per output.

* **`Muxer`** indexes `streams()` by input stream index instead of arrival order, so it stays in sync with the stream index carried by samples and segment info, and finalizes the output only once every input stream has been flushed (new `OnStreamEnded()` hook).
* **`MP4Muxer`** defers initialization until every track has produced its first sample. A track's edit list is derived from its own first sample, and the previous code applied the first track's offset to every `trak`, which would have desynced audio against video. Samples seen in the meantime are buffered and replayed in arrival order. It also keeps the `pssh` boxes of every encrypted track instead of only the last one, deduplicating identical boxes.
* **`mp4::Segmenter`** queues samples and segment boundaries for a track that reached the segment boundary before the others, instead of failing with `FRAGMENT_FINALIZED`, and replays them once the fragment covering all tracks has been written. At end of stream it writes a trailing fragment covering only the tracks that still had data, since tracks rarely end at exactly the same timestamp.

## Restrictions

Multiplexed output is deliberately restricted to plain MP4 files. Each of these reports what is not supported rather than failing deep in the pipeline:

* MP4 output, without `segment_template`
* audio and video only (not text, not trick play)
* all streams from the same `input` file — they share a demuxer, and therefore a thread, so no locking is needed around the shared muxer
* no manifest generation (`--mpd_output`, `--hls_master_playlist_output`, `--output_media_info`), since neither a DASH Representation nor an HLS Variant Stream can describe a file that holds more than one track

Reusing one output for the *same* stream is still rejected as a duplicated output.

## Testing

* `packager_test.cc`: the accepted combination plus each rejected one.
* `packager_test.py`: end-to-end tests packaging audio and video into one file, clear and encrypted, asserting the result reads back as two streams and, in the encrypted case, that both report `is_encrypted: true`. Two new golden files.
* Existing goldens are unchanged, confirming single-stream output is byte-for-byte identical to before.
* Manually verified that decrypting a multiplexed encrypted file back into a multiplexed file reproduces the clear multiplexed file byte-for-byte, that both `tenc` boxes carry the distinct per-`drm_label` key ids, and that multi-fragment output (`--segment_duration=0.5`) parses back with both tracks at full duration.

`TtmlMuxerTest.HandlesStyles` fails locally, but it fails the same way on an unmodified checkout of the base commit, so it is unrelated to this change.

Fixes #1480


joeyparrish:
> This PR and https://github.com/shaka-project/shaka-packager/pull/1601 touch the same part of packager.cc — whichever lands second will need a rebase. https://github.com/shaka-project/shaka-packager/pull/1618 is independent of both.

Apologies, but I just merged #1601.  I will see if I am able to resolve the conflict myself real quick.

joeyparrish:
My merge conflict resolution was not enough.  Making a couple additional changes, such as updating a test from the merge that referenced the old API.

joeyparrish:
Thanks for this, and sorry for the slow review. I merged `main` into the branch and fixed up the fallout, so the conflict is resolved and CI should be able to run now:

- The conflict in `mp4_muxer.cc` was between your per-track `edit_list_offsets_` and the negative-pts fix that landed on `main` (#1610). I kept `main`'s clamp applied to the per-track slot: `edit_list_offsets_[stream_id] = std::max(pts_dts_offset, -dts);`. The `pts < 0` `MUXER_FAILURE` block your branch carried is gone, since that fix deliberately replaced it. The remaining `MUXER_FAILURE` ("Not expecting pts < dts") is unrelated and stays.
- `mp4_muxer_unittest.cc` came from `main` and auto-merged without a conflict, but it did not compile afterwards: it still called `UpdateEditListOffsetFromSample(*sample)` and read `edit_list_offset_`. Fixed to pass a stream id and read `edit_list_offsets_`.

The per-track edit list work itself looks right to me, and it holds up in testing.

I did find two things that I think need to be addressed before this can land.

## 1. Crash when a segment boundary arrives before a track's first sample

`MP4Muxer::FinalizeSegment()` calls `segmenter_->FinalizeSegment()` behind `DCHECK(segmenter_)`, but `segmenter_` now stays null until *every* track has produced its first sample. If one track starts late, another track's chunker emits a segment boundary while the muxer is still uninitialized.

Repro with an mp4 whose audio track starts at ~10s (`ffmpeg -i in.mp4 -itsoffset 10 -i in.mp4 -map 0:v -map 1:a -c copy delayed_audio.mp4`):

```
$ packager in=delayed_audio.mp4,stream=video,output=mux.mp4 \
           in=delayed_audio.mp4,stream=audio,output=mux.mp4 \
           --segment_duration 1

F mp4_muxer.cc:267] Check failed: segmenter_
    @ shaka::media::mp4::MP4Muxer::FinalizeSegment()
    @ shaka::media::Muxer::Process()
    @ shaka::media::SegmentCoordinator::Process()
    @ shaka::media::ChunkingHandler::EndSegmentIfStarted()
```

Debug aborts on the DCHECK; in Release the DCHECK compiles out and we dereference null. A smaller `--segment_duration` widens the window. It looks like `FinalizeSegment` needs the same deferral treatment `AddMediaSample` got, so that a boundary arriving before initialization is queued rather than passed straight through.

## 2. Unbounded buffering when tracks have different lengths

`FinalizeSegmentInternal()` only writes a fragment once *every* fragmenter has finalized one:

```cpp
  // Check if all tracks are ready for fragmentation.
  for (const std::unique_ptr<Fragmenter>& fragmenter : fragmenters_) {
    if (!fragmenter->fragment_finalized())
      return Status::OK;
  }
```

Once a track ends it never finalizes again, so the remaining tracks can never satisfy that condition and their data accumulates until `Segmenter::Finalize()` flushes it all as one very large trailing fragment.

Peak RSS for the same clip, with audio truncated to 3s:

| Input | Equal-length tracks | 3s audio | Delta |
|---|---|---|---|
| 35s video | 112.5 MB | 116.7 MB | 4.2 MB |
| 3min video | 115.5 MB | 131.9 MB | 16.4 MB |

Roughly linear in the length of the tail. This is only tens of MB at 640x360, but at a realistic bitrate over a feature-length asset it becomes gigabytes, plus a single huge fragment in the output. To be clear, the output is otherwise correct here: packet counts match the non-multiplexed baseline exactly. It is a resource problem, not corruption.

## Things I checked that are fine

So you do not spend time on them: per-track edit list offsets and the merged negative-pts clamp; PSSH dedup with distinct per-track keys (correctly emits a single v1 common PSSH listing both KIDs); no sample loss with uneven track lengths; and `Segmenter::Finalize()`'s new drain/flush loop terminates.

One red herring worth mentioning: `ffprobe` reports the short audio track as having the full movie duration. That is the `mehd` fallback because `mdhd`/`tkhd` durations are deliberately left at 0, and it predates this PR.

## Smaller follow-ups

Not blockers, but worth a look while you are in here:

1. The trailing partial fragment falls back to `stream_ids.front()` for the `sidx` reference when the video reference track is absent, so that sidx entry ends up describing a different track's timeline.
2. `MP4Muxer::Finalize()` calls `InitializeAndFlushPendingSamples()` without consulting `ReadyToInitialize()`, so `DelayInitializeMuxer()`'s `DCHECK(AllStreamInfoReceived())` is unguarded on that path, and a track with no offset silently gets 0.
3. `Muxer::Process()`'s cue-event path does `Finalize()` + `ReinitializeMuxer()` per input stream when `output_file_template_` is set. With one shared muxer that would finalize the file once per track. I did not check whether `ValidateMultiplexedOutput()` already makes this unreachable.
4. `ValidateMultiplexedOutput()` compares descriptors by `stream_selector` equality, so two descriptors naming the same track differently (`stream=0` vs `stream=video`) would not be caught as duplicates.
5. `last_segment_info_` is default-constructed if no segment boundary ever occurred, and is then used for the trailing fragment's `DoFinalizeSegment()`.
6. There is no test coverage for late-starting or early-ending tracks, which is where both of the issues above come from. Worth adding alongside the fixes.

PR #1618, opened 2026-08-03
fix(mp4): keep tfhd.sample_description_index within the stsd table

### LINKED ISSUE #1616: AES-128 encryption emits tfhd.sample_description_index=2 for a single-entry stsd (video fMP4) — breaks ExoPlayer/Media3
**System info**

- Packager: `v3.9.1-68d23ea-release` (linux-x64 release binary)
- OS: Ubuntu 24.04 (container, `node:20` base)
- Also reproduced on `v3.8.0-75b2faf-release`

**What happened**

When packaging a **video** stream to fMP4/CMAF with raw-key `--protection_scheme aes128`,
the generated media segments contain `tfhd.sample_description_index = 2`, while the
init segment's `stsd` has `entry_count = 1`.

`sample_description_index` is a 1-based index into the sample description table
(ISO/IEC 14496-12, §8.8.7 `tfhd`). With `entry_count = 1`, the only valid value is `1`.

Packaging the **same input with the same flags but no encryption** correctly emits
`sample_description_index = 1`. Encryption is the only variable.

**Impact**

Players that resolve the sample description for the fragment fail on the out-of-range
index. AndroidX Media3 / ExoPlayer (1.8.0) throws while parsing the first `traf`:

```
androidx.media3.exoplayer.ExoPlaybackException: Source error
Caused by: androidx.media3.exoplayer.upstream.Loader$UnexpectedLoaderException:
  Unexpected ArrayIndexOutOfBoundsException: length=1; index=1
	at androidx.media3.extractor.mp4.Track.getSampleDescriptionEncryptionBox(Track.java:131)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor.parseTraf(FragmentedMp4Extractor.java:1054)
	at androidx.media3.extractor.mp4.FragmentedMp4Extractor.parseMoof(FragmentedMp4Extractor.java:1017)
	at androidx.media3.exoplayer.hls.HlsMediaChunk.feedDataToExtractor(HlsMediaChunk.java:535)
```

(`length=1` is the 1-entry table; `index=1` is the 0-based lookup of description 2.)

The stream is otherwise well-formed: the playlist, key and segments all fetch and
decrypt correctly, and the decrypted fragment parses cleanly up to this field.

An **audio-only** stream packaged with the same encryption flags plays fine, so this
appears specific to the video path.

**What you expected to happen**

`tfhd.sample_description_index` should be `1` (or omitted, defaulting to 1) when the
track has a single sample description — the same value the unencrypted run emits.

**How do we reproduce the error?**

```sh
# 1. any single-video-track mp4
ffmpeg -f lavfi -i "testsrc=size=1280x720:rate=25:duration=12" \
  -c:v libx264 -preset ultrafast -pix_fmt yuv420p -an -f mp4 video.mp4

# 2. CLEAR
packager \
  "in=video.mp4,stream=video,init_segment=init.mp4,segment_template=seg_\$Number\$.m4s,playlist_name=media.m3u8" \
  --segment_duration 4 --hls_master_playlist_output master.m3u8 --hls_playlist_type VOD

# 3. AES-128 (same flags + raw-key encryption)
packager \
  "in=video.mp4,stream=video,init_segment=init.mp4,segment_template=seg_\$Number\$.m4s,playlist_name=media.m3u8" \
  --enable_raw_key_encryption --protection_scheme aes128 \
  --keys "label=:key_id=00000000000000000000000000000000:key=11111111111111111111111111111111" \
  --hls_key_uri "enc-key/test" --clear_lead 0 \
  --segment_duration 4 --hls_master_playlist_output master.m3u8 --hls_playlist_type VOD
```

Read `stsd.entry_count` from `init.mp4`, then read `tfhd.sample_description_index` from
`seg_1.m4s`. For the AES-128 run the segment is whole-segment encrypted, so decrypt it
first with the key above and the `IV=0x…` from `media.m3u8`
(`openssl enc -d -aes-128-cbc -nopad -K 1111...11 -iv <IV>`).

Result:

| run | `stsd` entry_count | `tfhd` sample_description_index |
| --- | --- | --- |
| clear | 1 | 1 |
| aes128 | 1 | **2** |

I'm happy to attach the two segments or test a patch if that helps.

shunyagatha:
Reproduced on the base commit exactly as reported:

| run | `stsd` entry_count | `tfhd` sample_description_index |
| --- | --- | --- |
| clear | 1 | 1 |
| aes128 | 1 | **2** |

## Root cause

`Fragmenter::FinalizeFragmentForEncryption()` runs for any track whose `StreamInfo` is encrypted. When a fragment has no sample encryption entries it points at index 2, the clear entry of the encrypted/clear sample description pair that signals a clear lead:

```cpp
if (sample_encryption.sample_encryption_entries.empty()) {
  // This fragment is not encrypted.
  const uint32_t kClearSampleDescriptionIndex = 2;
  traf_->header.sample_description_index = kClearSampleDescriptionIndex;
```

AES-128 encrypts whole segments rather than individual samples: `EncryptionHandler` passes samples through clear for `kAes128ProtectionScheme`, and `MP4Muxer::GenerateVideoTrak`/`GenerateAudioTrak` deliberately skip the `encv`/`enca` conversion and the paired clear entry for that scheme. So the track has exactly one sample description, every fragment takes the branch above, and every fragment references an entry that is not there.

Audio was affected too. The reporter's audio-only stream plays because Media3 only resolves the sample description when it is looking for an encryption box.

## Fix

`Segmenter` owns the `moov`, so it reads the size of the sample description table the muxer actually built and passes it to each `Fragmenter`, which only switches to the clear entry when the table really has one. Deriving the bound from the built table, rather than re-testing the conditions that produce the second entry, keeps the two from drifting apart if those conditions ever change.

The alternative, skipping `FinalizeFragmentForEncryption()` entirely for `kAes128ProtectionScheme`, was proposed independently in #1621 by @horamarques and produces byte-identical output. That framing is a good description of the underlying mistake: the function exists to describe per-sample encryption, and AES-128 has none, so its clear-lead branch never applied. It was not taken because it asks the same question `MP4Muxer` asks when deciding not to emit the pair, leaving two places to keep in sync by hand.

## Testing

* New `FragmenterTest` unit tests: a fragment with no sample encryption entries points at the clear sample description when the track carries the encrypted/clear pair, and stays at index 1 when it has a single entry. The second fails on the base commit with the reported index of 2.
* Goldens updated for `fmp4-hls-with-aes128-encryption` and `single-segment-mp4-hls-with-aes128-encryption`. The change is precisely scoped: decrypting the old and new segments and diffing the plaintext shows **exactly one differing byte per segment**, the index itself, `2` to `1`. Segment lengths are unchanged, and the init segments and playlists are byte-identical, so no byte ranges move.
* All other unit and integration tests pass unchanged.

Fixes #1616


packager/media/formats/mp4/segmenter.cc
```diff
@@ -87,8 +87,18 @@ Status Segmenter::Initialize(
...
-    fragmenters_[i].reset(
-        new Fragmenter(streams[i], &moof_->tracks[i], edit_list_offset));
+    // Take the sample description count from the table the muxer actually
+    // built, so a fragment can never reference an entry that is not there.
+    const SampleDescription& sample_description =
+        moov_->tracks[i].media.information.sample_table.description;
+    const size_t num_sample_descriptions =
+        sample_description.video_entries.size() +
+        sample_description.audio_entries.size() +
+        sample_description.text_entries.size();
```
joeyparrish, line 97: `SampleDescription::ReadWrite` already derives the written `entry_count` by switching on `type` and picking exactly one of the three vectors (`box_definitions.cc:642-655`). Summing all three here agrees with that only because exactly one is ever populated, so the same invariant is now encoded in two places in two different ways.

Consider a small accessor on `SampleDescription` used by both sites. It would sit naturally next to the existing `TODO(kqyang)` about collapsing the three vectors into one.

packager/media/formats/mp4/fragmenter.cc
```diff
@@ -250,11 +253,20 @@ void Fragmenter::GenerateSegmentReference(SegmentReference* reference) const {
...
+    if (num_sample_descriptions_ >= kClearSampleDescriptionIndex) {
+      traf_->header.sample_description_index = kClearSampleDescriptionIndex;
+      return Status::OK;
+    }
+    // Otherwise there is no second entry to point at and the index must stay
+    // at 1. This is the case for whole-segment AES-128 encryption, which
+    // leaves the samples themselves clear and so generates no encrypted
+    // sample description to pair a clear one with. Pointing past the end of
+    // the table breaks players that resolve the sample description, see
+    // https://github.com/shaka-project/shaka-packager/issues/1616.
```
joeyparrish, line 269: Readability nit: this branch now has two `return Status::OK;` exits for what is really one decision, and the explanation is split across two comment blocks. It collapses to a single exit:

```c++
    // This fragment is not encrypted. When a clear lead is signalled the track
    // carries two sample description entries, an encrypted one and a clear one,
    // and the 1-based index of the clear entry is always 2. Without that pair
    // there is no second entry to point at and the index must stay at 1. That
    // is the case for whole-segment AES-128 encryption, which leaves the
    // samples themselves clear and so generates no encrypted sample
    // description to pair a clear one with. Pointing past the end of the table
    // breaks players that resolve the sample description, see #1616.
    const uint32_t kClearSampleDescriptionIndex = 2;
    if (num_sample_descriptions_ >= kClearSampleDescriptionIndex)
      traf_->header.sample_description_index = kClearSampleDescriptionIndex;
    return Status::OK;
```

packager/media/formats/mp4/fragmenter_unittest.cc
```diff
@@ -0,0 +1,108 @@
...
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+#include <packager/media/formats/mp4/fragmenter.h>
+
+#include <cstdint>
+#include <memory>
+#include <vector>
```
joeyparrish, line 11: `<vector>` does not appear to be used in this file.

packager/media/formats/mp4/fragmenter_unittest.cc
```diff
@@ -0,0 +1,108 @@
...
+// description, so a fragment that is not encrypted points at the clear one.
+TEST(FragmenterTest, ClearFragmentUsesClearSampleDescription) {
+  EXPECT_EQ(kClearSampleDescriptionIndex, GetSampleDescriptionIndex(2));
+}
+
+// Without that pair there is only one entry to point at. AES-128 leaves the
+// samples clear and generates no encrypted sample description, so every one of
+// its fragments took the branch above and referenced an entry past the end of
+// the table, which players resolving it choke on.
+// See https://github.com/shaka-project/shaka-packager/issues/1616.
```
joeyparrish, line 101: "took the branch above" points at the branch in `Fragmenter::FinalizeFragmentForEncryption`, which is not visible from this file, so it does not resolve for someone reading the test. Naming it explicitly would help.

packager/app/test/packager_test.py
```diff
@@ -1610,6 +1610,10 @@ def testFmp4HlsWithAes128Encryption(self):
                          segmented=True,
                          hls=True),
         flags)
+    # The golden segments cover tfhd.sample_description_index staying within
+    # the track's single sample description entry. See FragmenterTest in
+    # packager/media/formats/mp4/fragmenter_unittest.cc and
+    # https://github.com/shaka-project/shaka-packager/issues/1616.
```
joeyparrish, line 1616: `testSingleSegmentMp4HlsWithAes128Encryption` just below has its goldens updated for exactly the same reason but carries no equivalent comment. Worth either adding it there too or dropping this one, so the two do not read as different cases.

joeyparrish (commented):
Looks great overall.  A few nits inline.

PR #1619, opened 2026-08-18
fix: preserve path separators when URL-escaping MPD BaseURL content

### LINKED ISSUE #1593: MPD BaseURL percent-encodes path separators when output files are in a subdirectory (subtitles%2Ffile.vtt)
**System info**

- Shaka Packager Version: v3.8.0 (regression introduced in v3.2.0)
- Operating System: Linux (docker, ubuntu-noble base)

**Issue**

When a stream's `output=` path is in a subdirectory relative to the `--mpd_output` location, the generated `<BaseURL>` percent-encodes the path separator, producing a URL most web servers cannot serve.

**Repro**

```bash
packager \
  "in=input.mp4,stream=video,output=/out/video.mp4" \
  "in=subs_en.vtt,stream=text,output=/out/subtitles/subs_en.vtt,language=en" \
  --mpd_output /out/manifest.mpd
```

**Actual output** (manifest.mpd):

```xml
<Representation id="1" ...>
  <BaseURL>subtitles%2Fsubs_en.vtt</BaseURL>
</Representation>
```

**Expected output:**

```xml
<BaseURL>subtitles/subs_en.vtt</BaseURL>
```

**Impact**

Per RFC 3986, `%2F` in a path is *not* a path segment separator — standard web servers (nginx, Apache, S3/CloudFront, http-server) either 404 or reject the request, so players cannot fetch the subtitle files. Anyone emitting text (or any) streams into a subdirectory relative to the MPD gets an unplayable manifest.

**Cause**

PR #1395 ("fix: escape media URLs in MPD", first released in v3.2.0) passes the entire relative URL through `curl_easy_escape` in `packager/mpd/base/xml/xml_node.cc`. `curl_easy_escape` encodes every byte outside `[A-Za-z0-9-._~]`, including `/`.

The intent of #1395 (escaping characters like `&` in filenames) is right, but the escaping should be applied **per path segment**, preserving `/` separators — e.g. split on `/`, escape each segment, re-join.

**Workaround**

Post-process the MPD, replacing the encoded separator (`%2F` → `/`) — which is what we do today, but it's a workaround for what looks like an unintended side effect of #1395.

horamarques:
## Problem
When outputs live in subdirectories, the MPD BaseURL was passed whole through `curl_easy_escape`, which percent-encodes `/` as `%2F`, producing unservable relative URLs like `subtitles%2Fsubs_en.vtt`.

## Fix
Escape each path segment individually; path separators are preserved. Absolute URLs keep their `scheme://authority` prefix untouched.

## Testing
New `xml_node_unittest` cases covering relative subdirectory paths and absolute URLs; full `mpd_unittest` passes.

Fixes #1593


joeyparrish (changes requested):
Thanks, the diagnosis is right and per-segment escaping is the right shape for the paths we generate. Two things before this lands.

`?` and `#` are not separators either, so query strings are still mangled:

```
--base_urls=https://cdn.example.com/v1/?token=a&b=2
  ->  <BaseURL>https://cdn.example.com/v1/%3Ftoken%3Da%26b%3D2</BaseURL>
```

Rather than exempting characters one at a time, I would split by caller. `--base_urls` (`mpd_builder.cc:184`) is a user-supplied prefix that is already a URL, so it wants XML escaping and no URL encoding at all; the media paths we build (`xml_node.cc:427`) are the suffix, and those want per-segment URL encoding. That also removes the need for the `://` special case. One caveat: `SetContent` is not sufficient for the prefix, because `xmlNodeSetContent` parses its argument as an attribute value and silently ignores malformed entity references, so a raw `&` would be dropped. `AddContent` takes raw text.

Please also add test coverage for `--base_urls`. There is none anywhere today, which is how this case stayed broken since #1395.

Thank you for contributing!

PR #1620, opened 2026-08-18
fix: read FLAC sample rate from STREAMINFO for rates above 65535 Hz

### LINKED ISSUE #1404: audioSamplingRate in representation is "0" for high sample rates
Representations greater than 96kHz have incorrect audioSamplingRate values in the Representation

```
      <Representation id="16" bandwidth="470315" codecs="flac" mimeType="audio/mp4" audioSamplingRate="0">
        <AudioChannelConfiguration schemeIdUri="urn:mpeg:dash:23003:3:audio_channel_configuration:2011" value="2"/>
        <BaseURL>representations/flac_60s_2ch_16bit_196000Hz.mp4</BaseURL>
        <SegmentBase indexRange="772-875" timescale="196000">
          <Initialization range="0-771"/>
        </SegmentBase>
      </Representation>
```

horamarques:
## Problem
The stsd `AudioSampleEntry` samplerate field is 16.16 fixed point and cannot represent rates above 65535 Hz, so muxers write a wrong or truncated value for high-rate FLAC. This surfaced as `audioSamplingRate="0"` in the MPD.

## Fix
Parse the authoritative 20-bit sample rate from the STREAMINFO block in the `dfLa` box, and prefer it over the stsd value whenever it parses. STREAMINFO is authoritative for FLAC, and the stsd value is unreliable rather than merely absent: the 192 kHz test asset carries 48000 there, not 0.

## Testing
New `MP4MediaParserTest.FlacHighSampleRateFromStreaminfo` with a 192 kHz FLAC asset (`bear-flac-192kHz.mp4`, generated with ffmpeg from bear.flac); full `mp4_unittest` passes.

Fixes #1404


packager/media/formats/mp4/mp4_media_parser_unittest.cc
```diff
@@ -257,6 +258,16 @@ TEST_F(MP4MediaParserTest, NoMoovAfterFlush) {
                                  buffer.size() - kFirstMoofOffset, 512));
 }
 
+// The stsd samplerate field cannot represent rates above 65535 Hz; the
+// sample rate must come from the FLAC STREAMINFO in the dfLa box.
+TEST_F(MP4MediaParserTest, FlacHighSampleRateFromStreaminfo) {
+  ASSERT_TRUE(ParseMP4File("bear-flac-192kHz.mp4", 512));
+  EXPECT_EQ(1u, num_streams_);
+  const AudioStreamInfo* audio_info =
+      static_cast<AudioStreamInfo*>(stream_map_.begin()->second.get());
```
joeyparrish, line 267: `EXPECT_EQ` is non-fatal, so if the parse yields no streams this dereferences `stream_map_.end()` and the test crashes instead of failing cleanly. `ASSERT_EQ` would cover it, or index by track id the way the tests above do (`stream_map_[kVideoTrackId]`).

joeyparrish (changes requested):
Thanks, this is the right diagnosis and the right place for the fix. Please look into these things before it lands:
 - Please add `bear-flac-192kHz.mp4` to `packager/media/test/data/README`. Every asset there records how it was created (see `bear-flac.mp4` at line 55), and the ffmpeg command from your PR description is exactly what belongs there.
 - The write path has the same 16.16 limitation, and I would like it covered here too. `AudioSampleEntry::samplerate` is `uint32_t`, and `box_definitions.cc:2089` does `samplerate <<= 16` before writing, so muxing a 192 kHz FLAC wraps to `0xEE000000` and the output stsd reads back as 60928 Hz. That is neither the true rate nor 0, which is the value readers treat as "consult the codec config". I think this is worth fixing, even though this PR already makes Packager immune to its own output by preferring `dfLa`.

A small note: I have adjusted the PR description, which framed STREAMINFO as a fallback for when the stsd rate is unusable. Now it accurately notes that we prefer `dfLa` over `stsd`.

Thank you for contributing!

Issue #1592, opened 2026-07-09
feat: MPD support for multiple sign language tracks with different languages


jkoppen-bluebillywig:
When adding *two* (ore more) sign language tracks, via 
"dash_roles=sign,language=ase,hls_characteristics=..,playlist_name=sign_xxx", 
"dash_roles=sign,language=dse,hls_characteristics=..,playlist_name=sign_yyy" (mind the difference; ase vs. dse), the output MPD manifest mixes up the Representations, putting them together in one AdaptationSet per codec. We worked around that by untangling them in a post-processing step, which is obviously not ideal.

Problem analysis:
MediaInfo.VideoInfo has no language field (packager/mpd/base/media_info.proto:23-58), so GetLanguage() (packager/mpd/base/mpd_utils.cc:91-99) returns empty for video. Language is structurally part of the AdaptationSet key (mpd_utils.cc:207-208) and lang emission already exists (adaptation_set.cc:334-337) — video language is simply never populated. Result: all same-codec sign representations merge into one AdaptationSet regardless of language, with no lang attribute.

PR #1624, opened 2026-08-19
feat: split DASH video AdaptationSets by language and emit lang attribute

jkoppen-bluebillywig:
Corresponding issue: #1592


ghultink (approved):
Reviewed at Blue Billywig (this fixes our #1592 use case). The mechanism is sound: language on VideoInfo only when explicitly user-overridden (language_overridden gate keeps input-metadata language inert, so untagged content is guaranteed byte-identical output — verified against the golden files and the LanguageFromInputIsIgnored/UndefinedLanguageOverride tests), the AdaptationSet split rides the existing GetAdaptationSetKey path, lang lands on AdaptationSet per DASH spec, and the two period.cc kRoleMain guards are correct (the roles bucket is role-homogeneous via the ':roles_' key suffix, so the comment's invariant holds). Three-layer test coverage incl. an e2e with golden MPD. One doc-level note: GetAdaptationSetKeyForTrickPlay keeps language, so a lang-tagged rendition with trick-play tracks needs the same lang= on the trick-play descriptor — symmetric with existing dash_roles behavior, but worth a line in the docs.

